PR #82, opened 2014-05-12
added binutils submodule

inolen:
Was working to try and get things compiling on my mac last night, here's the first few minor issues I hit. I wasn't sure if binutils wasn't added just due to its size (it's a large clone), but alas:

 * added binutils submodule
 * default to 2013 when VSVERSION doesn't exist (using os.environ['missingkey'] throws exception)


benvanik:
third_party/binutils/build.sh takes care of this now

Issue #96, opened 2014-07-04
Proper ISO/STFS directory enumeration


benvanik:
Only host filesystem currently does this; need to support it on ISO/STFS to make testing easier.

* Kameo
* Street Fighter
* Puzzle Bobble
* Catan

benvanik:
Currently relying on FindFirstFile and other Win32 APIs - need to use custom enumerator with consistent ordering and wildcard support. That should then be used on the host fs to ensure consistency.

benvanik:
HostPathEntry using Win32 APIs: https://github.com/benvanik/xenia/blob/master/src/xenia/kernel/fs/devices/host_path_entry.cc#L75

Instead, it should be listing all children on creation so that a generic Entry enumeration function can be used and shared between all implementations. Both DiscImage and StfsContainer already somewhat do this, but are missing the filter support.

DiscImageEntry (https://github.com/benvanik/xenia/blob/master/src/xenia/kernel/fs/devices/disc_image_entry.cc#L59) and StfsContainerEntry (https://github.com/benvanik/xenia/blob/master/src/xenia/kernel/fs/devices/stfs_container_entry.cc#L51) are both broken and only support '*'.

Moving the child enumeration code into Entry so that all 3 implementations can share it would be really nice. Only the code filling the XDirectoryInfo in the device-specific way should be unique across them.

benvanik:
Space Giraffe does a *.png enumeration.

benvanik:
From Dust does '*'

benvanik:
Hexic 2 does *.jpg

x1nixmzeng:
I still need to look at moving the routine into common code, and hooking this up with the host paths.

benvanik:
Done.

PR #166, opened 2015-02-13
Support for wildcard queries

x1nixmzeng:
Provides support for persistent wildcard file requests, as described in #96

Also moved CanonicalizePath into common code (poly::fs)

Note: there is still plenty of room for optimization


benvanik:
oo, will give this a shot!

PR #168, opened 2015-02-16
Fixed writing data to the host device

x1nixmzeng:
New paths can now be created if they do not exist - this allows content packages such as savedata to be written to disk


benvanik:
Ah, I added that check because without it you end up with a bunch of 0-length files as the game tries to check for the existence of files. I think we need a bool create_if_missing flag on ResolvePath to control this behavior (so if !read_only && create_if_missing).

x1nixmzeng:
I can't reproduce that behaviour.

The returned HostPathEntry would still need to be opened with `Mode::READ_WRITE` in order for it to actually create a placeholder file of 0 bytes.

I'll take a look at some other games.

benvanik:
Very true -- I'm already forgetting how I had this all implemented ;P

PR #169, opened 2015-02-17
Adjustments to wildcard engine

x1nixmzeng:
Fixed incorrect logic when checking at the end of the string - some of the names had an added null character
Removed case sensitivity


benvanik:
:+1:

Issue #171, opened 2015-02-18
Simple game API usage scanner.


benvanik:
Would be nice to point a tool at a path and have it load ISO/STFS, crack open the xex, and dump a list of APIs used. The resulting data would be useful when trying to find other games using APIs or verifying compatibility.

x1nixmzeng:
Do you see this working as a static analysis tool? Running a pass on the translated machine code, resolving API calls may be something I could have a go at writing.

I'm not sure how to approach discovering arguments at the moment.

benvanik:
Oh, that'd be cool, but something even simpler would be useful: load the xex, scan the import table, and dump them to a file/table/etc. That way if we change XamFoo or some other method we have a good way of knowing which games to re-test. Going further and identifying API usage patterns would be awesome too, but not required to be useful.

I imagined a command line app linking in libxenia that used the filesystem API to load the iso/stfs/etc then xex2.h/cc to load the xex and get the header. That'd be the laziest/fastest way at least :P

x1nixmzeng:
Well there's plenty to go on - [xuser_module.cc](../blob/c846ec30c3b7dd5eab85a153aa9e2a679d1e9f78/src/xenia/kernel/objects/xuser_module.cc#L194) already dumps what we need!

I'll look into making it a bit more standalone.

PR #172, opened 2015-02-21
Added API scanner tool

x1nixmzeng:
New tool for issue #171 which links to libxenia to dump the API usage from packaged content.

Summary

* Registering XKernelModule tables has been partially moved into a static 'RegisterExportTable'
* Due to XObject usage without a kernel state, added some checks for nullptr to support basic usage
* Also commented out the STFS container dumping by default

### DIFF
--- a/src/xenia/tools/api-scanner/api_scanner_loader.h
+++ b/src/xenia/tools/api-scanner/api_scanner_loader.h
@@ -1,0 +1,69 @@
+/**
+ ******************************************************************************
+ * api-scanner - Scan for API imports from a packaged 360 game                *
+ ******************************************************************************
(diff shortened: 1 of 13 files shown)


src/xenia/tools/api-scanner/api_scanner_loader.h
```diff
@@ -0,0 +1,69 @@
...
+    const std::vector<title>& GetAllTitles() const { return loaded_titles; }
+
+  private:
+    std::vector<title> loaded_titles;
+
+    bool ReadTarget();
+    bool ExtractImports(const void* addr, const size_t length, title& info);
+  };
+
+} // tools
```
benvanik: just a nit, but }  // namespace tools
I need to get clang-format auto fixing this stuff on git precommit.

PR #185, opened 2015-05-02
Remove cxx_compat.h header

sephiroth99:
Revert commits 6d159dc2 and 3f480d24. Visual Studio 2013 is not
supported anymore, so they are not needed.


benvanik:
Woo, yeah. Will delete the file in a follow on CL.

PR #186, opened 2015-05-02
Replace some instructions for CPUs without AVX2/CNTLZ

DrChat:
Added code to dynamically select which instructions to use at runtime based on the host's CPU

I've ran the tests in xe-cpu-ppc-test, they've all passed except for the ones I still haven't replaced.

### DIFF
--- a/src/xenia/cpu/backend/x64/x64_emitter.cc
+++ b/src/xenia/cpu/backend/x64/x64_emitter.cc
@@ -67,7 +67,8 @@ X64Emitter::X64Emitter(X64Backend* backend, XbyakAllocator* allocator)
       current_instr_(0),
       source_map_count_(0),
       stack_size_(0),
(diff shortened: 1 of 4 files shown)


DrChat:
Forgot to mention: There's an issue with xbyak_util.h including "xbyak/xbyak.h". I changed it to just "xbyak.h" for now (and submitted an issue on their repo)

benvanik:
If you add this to xenia.gyp line 291 (the 'include_dirs'):
```
+        # TODO(benvanik): remove when xbyak fixed:
+        # https://github.com/herumi/xbyak/issues/20
+        'third_party/xbyak/',
```

Then you don't need to have a local xbyak change. It's ugly, but until they fix it at least builds.

Otherwise, looks great - still runs for me and it's good progress :)

PR #187, opened 2015-05-04
XexGetProcedureAddress by proc name

DrChat:
Reads the XEX's export table to find exports and stuff

### DIFF
--- a/src/xenia/kernel/util/xex2.cc
+++ b/src/xenia/kernel/util/xex2.cc
@@ -49,8 +49,6 @@ typedef struct xe_xex2 {
     size_t count;
     xe_xex2_import_info_t *infos;
   } library_imports[16];
-
-  xe_xex2_export_table *export_table;
 } xe_xex2_t;
 
 int xe_xex2_read_header(const uint8_t *addr, const size_t length,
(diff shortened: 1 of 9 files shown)


src/xenia/kernel/xboxkrnl_modules.cc
```diff
@@ -265,9 +270,24 @@ SHIM_CALL XexGetProcedureAddress_shim(PPCContext* ppc_state,
 
   if (XSUCCEEDED(result)) {
     // TODO(benvanik): implement. May need to create stub functions on the fly.
```
benvanik: kill the TODO

src/xenia/kernel/util/xex2.cc
```diff
@@ -869,6 +876,22 @@ int xe_xex2_load_pe(xe_xex2_ref xex) {
     xex->sections->push_back(section);
   }
 
+  /*
+  if (header->export_table_offset) {
```
benvanik: is this right? to prevent future confusion, kill it if not needed :)

DrChat: it is, but i commented it out because it's an absolute address so it's not reliable if we decide to relocate the module.

src/xenia/kernel/xboxkrnl_modules.cc
```diff
@@ -246,10 +246,15 @@ SHIM_CALL XexGetProcedureAddress_shim(PPCContext* ppc_state,
                                       KernelState* state) {
   uint32_t module_handle = SHIM_GET_ARG_32(0);
   uint32_t ordinal = SHIM_GET_ARG_32(1);
+  const char *name = (const char *)SHIM_MEM_ADDR(ordinal);
   uint32_t out_function_ptr = SHIM_GET_ARG_32(2);
 
-  XELOGD("XexGetProcedureAddress(%.8X, %.8X, %.8X)", module_handle, ordinal,
-         out_function_ptr);
+  if (ordinal < 0x10000)
```
benvanik: style nit: {} around all ifs/elses/etc - I think 'xb format' will fix it for you

PR #190, opened 2015-05-05
XexLoadImage support for user modules

DrChat:


src/xenia/kernel/kernel_state.cc
```diff
@@ -136,6 +145,28 @@ void KernelState::SetExecutableModule(XUserModule* module) {
   }
 }
 
+XUserModule* KernelState::LoadUserModule(const char *name) {
+  // See if we've already loaded it
```
benvanik: add std::lock_guard<std::mutex> lock(object_mutex_); --- this isn't thread safe

src/xenia/kernel/kernel_state.cc
```diff
@@ -107,8 +111,13 @@ XModule* KernelState::GetModule(const char* name) {
     // Some games request this, for some reason. wtf.
     return nullptr;
   } else {
-    // TODO(benvanik): support user modules/loading/etc.
-    assert_always();
+    for (XUserModule *module : user_modules_) {
```
benvanik: add a std::lock_guard<std::mutex> lock(object_mutex_); -- (just in this else is fine)

src/xenia/kernel/kernel_state.cc
```diff
@@ -136,6 +145,28 @@ void KernelState::SetExecutableModule(XUserModule* module) {
...
+      module->Retain();
+      return module;
+    }
+  }
+
+  // Module wasn't loaded, so load it.
+  XUserModule *module = new XUserModule(this, name);
+  X_STATUS status = module->LoadFromFile(name);
+  if (XFAILED(status)) {
+    delete module;
```
benvanik: module->Release(); (otherwise ~XObject will assert when the module is deleted)

benvanik:
:+1:

PR #192, opened 2015-05-05
PREMUTE_I32 Non-AVX2 Instructions

DrChat:


src/xenia/cpu/backend/x64/x64_sequences.cc
```diff
@@ -5490,8 +5487,8 @@ EMITTER(SPLAT_I16, MATCH(I<OPCODE_SPLAT, V128<>, I16<>>)) {
         e.vpbroadcastw(i.dest, e.xmm0);
       }
     } else {
-      // TODO(justin)
-      e.DebugBreak();
+      // TODO
+      //e.DebugBreak();
```
benvanik: worth having this break/assert/die/etc - otherwise you'll miss it

PR #199, opened 2015-05-09
Setup UI menu code

DrChat:
Menus menus menus!


src/xenia/ui/main_window.h
```diff
@@ -16,6 +16,7 @@
 // TODO(benvanik): only on windows.
 #include "xenia/ui/win32/win32_loop.h"
 #include "xenia/ui/win32/win32_window.h"
+#include "xenia/ui/win32/win32_menu_item.h"
```
benvanik: being pedantic ;) for future cls in sorted order plz (I need to add a 'xb format' script that does this automatically)

DrChat: It's sorted by uhh... length of line!

PR #209, opened 2015-05-16
Rewrite FMA mul sub/add to avoid register shuffling

DrChat:


benvanik:
@hlide yep! oh tricky intel. He fixed that here, but it would be nice to setup the tool they mention in there for detecting the penalties - there may be other places this stuff is hiding.

Thanks for fixing @DrChat !

PR #213, opened 2015-05-18
User module imports

DrChat:


src/xenia/kernel/util/xex2.cc
```diff
@@ -610,6 +610,7 @@ int xe_xex2_read_image_basic_compressed(const xe_xex2_header_t *header,
     return 1;
   }
   uint8_t *buffer = memory->TranslateVirtual(header->exe_address);
+  std::memset(buffer, 0, total_size); // Quickly zero the contents.
   uint8_t *d = buffer;
   std::memset(buffer, 0, uncompressed_size);
```
benvanik: one of these isn't needed now, right?

DrChat: Portal crashes without this - I'm guessing the BSS section isn't compressed at all.

benvanik: Oh, I mean that there's two memsets of the buffer - only memset(max(total_size, uncompressed_size)) is needed (I don't know which is bigger).

DrChat: Oh wow, you're right - I didn't see that. 2nd one is unnecessary, I'll remove it

PR #215, opened 2015-05-19
Host XThread class

DrChat:


src/xenia/kernel/objects/xthread.h
```diff
@@ -103,6 +103,17 @@ class XThread : public XObject {
   XEvent* event_;
 };
 
+class XHostThread : public XThread {
+  public:
+    XHostThread(KernelState* kernel_state, uint32_t stack_size,
+                uint32_t creation_flags, std::function<int()> host_fn);
+
+    virtual void Execute();
```
benvanik: sexy c++11 feature: you can write this as 'void Execute() override;' and then it'll error out at you if XTherad::Execute() ever changes (where as the virtual could get disconnected)

PR #217, opened 2015-05-23
Audio decoding

DrChat:


src/Xenia.Debugger/MainWindow.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
benvanik: Not sure where these all came from - can you remove them?

src/xenia/apu/audio_decoder.cpp
```diff
@@ -0,0 +1,239 @@
+/**
```
benvanik: .cc for C++ files plzz

src/xenia/apu/audio_decoder.cpp
```diff
@@ -0,0 +1,239 @@
...
+  #include "libavcodec/avcodec.h"
+}
+
+// Credits for most of this code goes to:
+// https://github.com/koolkdev/libertyv/blob/master/libav_wrapper/xma2dec.c
+
+namespace xe {
+namespace apu {
+
+AudioDecoder::AudioDecoder() : offset_(0) {
```
benvanik: should also initialize the pointers to nullptr so that the ~dtor won't crash if Initialize() fails

src/xenia/apu/audio_decoder.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+  if (bits <= 0 || bits > 32 || (bits % 8) != 0) {
+    assert_always();
+  }
+
+  // Output bits per sample
+  bits_ = bits;
+
+  // Allocate important stuff
+  codec_ = avcodec_find_decoder(AV_CODEC_ID_WMAPRO);
```
benvanik: it's probably worth logging on these failures (XELOGE) - unless they're super unlikely

src/xenia/apu/audio_decoder.cpp
```diff
@@ -0,0 +1,239 @@
...
+  packet_->size = XMAContextData::kBytesPerBlock;
+
+  // Re-initialize the context with new sample rate and channels
+  if (context_->sample_rate != sample_rate || context_->channels != channels) {
+    context_->sample_rate = sample_rate;
+    context_->channels = channels;
+
+    // We have to reopen the codec so it'll realloc whatever data it needs.
+    // TODO: Find a better way.
+    avcodec_close(context_);
```
benvanik: If this ends up being super expensive we can pool the AudioDecoders - will have to watch it

src/xenia/apu/audio_decoder.cpp
```diff
@@ -0,0 +1,239 @@
...
+    if (avcodec_open2(context_, codec_, NULL) < 0) {
+      // Shit.
+      return 1;
```
benvanik: If likely to fail, throw a log in here

src/xenia/apu/audio_system.cc
```diff
@@ -154,6 +171,156 @@ void AudioSystem::WorkerThreadMain() {
...
+            // TODO: Set error state
+
+            break;
+          }
+
+          if (read == 0) {
+            // Select sample rate.
+            int sample_rate = 0;
+            if (data.sample_rate == 0) {
+              // TODO: Test this
```
benvanik: is it worth throwing asserts here so we can verify it works? or do you think they're pretty likely to be correct?

src/xenia/cpu/frontend/test/instr_test.s
```diff
@@ -0,0 +1,3 @@
+test:
```
benvanik: ?

third_party/ninja
```diff
@@ -0,0 +1 @@
+Subproject commit c406d1c8adfedc1982e2c08ab95d581f65eb65de
```
benvanik: ninjas is dead?

src/xenia/apu/audio_decoder.cpp
DrChat: Yeah, don't think they're likely to fail.

src/xenia/apu/audio_system.cc
DrChat: data.sample_rate can't be out of range, it's only got 2 bits

benvanik: Right, but anywhere you have a TODO it's a good idea to assert, as otherwise games may start using these paths and we won't know where to look.

DrChat: Yeah, you're right. It should fail to decode the packet though and log it, so that'd be fine as-is.

benvanik: The 4 TODO's here won't log anything, as PreparePacket doesn't log anything on failure ;)

src/xenia/apu/audio_decoder.cc
```diff
@@ -0,0 +1,239 @@
...
+  if (context_->sample_rate != sample_rate || context_->channels != channels) {
+    context_->sample_rate = sample_rate;
+    context_->channels = channels;
+
+    // We have to reopen the codec so it'll realloc whatever data it needs.
+    // TODO: Find a better way.
+    avcodec_close(context_);
+    if (avcodec_open2(context_, codec_, NULL) < 0) {
+      // Shit.
+      return 1;
```
benvanik: This should log on failure.

src/xenia/apu/audio_system.cc
DrChat: It'll loop back around to DecodePacket and fail on that

PR #220, opened 2015-05-23
A few fixes for portal

DrChat:
Gets portal up to the point where it freezes initializing the audio system (cause of thread pcr+0x10C not being correct CPU id)


src/xenia/kernel/xboxkrnl_memory.cc
```diff
@@ -306,8 +306,8 @@ SHIM_CALL MmAllocatePhysicalMemoryEx_shim(PPCContext* ppc_state,
   // and the memory must be allocated there. I haven't seen a game do this,
   // and instead they all do min=0 / max=-1 to indicate the system should pick.
   // If we have to suport arbitrary placement things will get nasty.
-  assert_true(min_addr_range == 0);
-  assert_true(max_addr_range == 0xFFFFFFFF);
+  //assert_true(min_addr_range == 0);
```
benvanik: does this seem work as expected? (I think I implemented AllocRange right, but haven't tested it through this path)
if so, let's remove the asserts.

DrChat: Yeah, it looks like it's working OK. I'll do that.

PR #226, opened 2015-05-26
Fixup vs sln

shuffle2:


benvanik:
First change is good, but 4a39111 is bad - we don't want Any CPU, just x64.

rlabrecque:
VS keeps making the change in 4a39111 for me, rather annoyingly. Haven't looked into stopping it from doing that yet. Maybe you want to look at it?

shuffle2:
WinFormsUI is also attempted to be built upon each "build solution", even though it's already up to date. Something must be funky there..

shuffle2:
In response to my last comment: It's just always run because of this rule:
```
1>_CopyOutOfDateSourceItemsToOutputDirectoryAlways:
1>  Copying file from "t:\src\xenia\third_party\dockpanelsuite\WinFormsUI\license.txt" to "bin\Debug\license.txt".
```

shuffle2:
I think fixing the "AnyCPU" issue would require modifying WinFormsUI.csproj.  
Anyways, with the commit I made here, Xenia.Debug.UI still compiles and runs...

shuffle2:
BTW, with either AnyCPU or x64, I cannot use the VS2015 UI Designer. It complains that various base classes could not be loaded, such as:
```
The designer could not be shown for this file because none of the classes within it can be designed.
The designer inspected the following classes in the file: StatisticsDocument ---
The base class 'Xenia.Debug.UI.Controls.BaseDocument' could not be loaded.
Ensure the assembly has been referenced and that all projects have been built. 
```
Sorry, I'm not very familiar with this.

benvanik:
Yeah, I'm thinking about rebuilding the WinFormsUI csproj without all that crap. I'm not sure why the designer doesn't work - from what I found on the net people said you had to have the base type within your project (which is why I made BaseDocument), but that didn't help. When I rebuild the csproj I'll try to make sure it all works. For now, I just switch the base type to Form or something when editing :(

benvanik:
Have forked dockpanelsuite and fixed up the designer and warnings in 9ac61ed869cd

PR #234, opened 2015-06-02
.cpp -> .cc and add ring_buffer.cc to the project

DrChat:


libxenia.vcxproj
```diff
@@ -33,6 +33,7 @@
     <ClCompile Include="src\xenia\base\mapped_memory_win.cc" />
     <ClCompile Include="src\xenia\base\math.cc" />
     <ClCompile Include="src\xenia\base\memory_generic.cc" />
+    <ClCompile Include="src\xenia\base\ring_buffer.cpp" />
```
benvanik: hit 'save all' in vs - this still has the cpp :P

PR #236, opened 2015-06-03
Xboxkrnl SHA Crypt Stubs

DrChat:


src/xenia/kernel/xboxkrnl_crypt.cc
```diff
@@ -0,0 +1,46 @@
...
+namespace kernel {
+
+typedef struct {
+  xe::be<DWORD>             count;
+  xe::be<DWORD>             state[5];
+  xe::be<BYTE>              buffer[64];
+} XECRYPT_SHA_STATE;
+
+void XeCryptShaInit(pointer_t<XECRYPT_SHA_STATE> sha_state) {
+  std::memset((void *)sha_state.host_address(), 0, sizeof(XECRYPT_SHA_STATE));
```
benvanik: `sha_state.Zero();` will do this for you

PR #239, opened 2015-06-04
Emulate PACK/UNPACK for non-F16C CPUs

DrChat:
This code might be correct. Maybe.


src/xenia/cpu/backend/x64/x64_sequences.cc
```diff
@@ -33,6 +33,9 @@
 #include "xenia/cpu/hir/hir_builder.h"
 #include "xenia/cpu/processor.h"
 
+// For OPCODE_PACK/OPCODE_UNPACK
+#include "half.hpp"
```
benvanik: where possible, let's use root-relative paths (like third_party/half/include/half.hpp) instead of adding global include directories

PR #243, opened 2015-06-05
Fixes some issues on AMD drivers

sephiroth99:
These three commits are to fix some issues seen when trying to run on AMD video cards. For reference, this is Catalyst 15.200.1018 on Win8.1.

It's not perfect as there is still OpenGL errors being thrown, but at least the profiler display works now.

For the structs as in/out, there are still some present in command_processor.cc and gl4_shader.cc, but no errors are being thrown?? I guess we'll see if those parts will need to change too.


benvanik:
Awesome. Also, damn AMD!

PR #244, opened 2015-06-06
Sector info stub

DrChat:


src/xenia/kernel/xboxkrnl_io.cc
```diff
@@ -469,30 +469,34 @@ SHIM_CALL NtSetInformationFile_shim(PPCContext* ppc_context,
...
+  union {
+    xe::be<uint32_t> status;
+    xe::be<uint32_t> pointer;
+  };
+  xe::be<uint32_t> information;
+};
 
+dword_result_t NtQueryInformationFile(dword_t file_handle,
+                                      pointer_t<X_IO_STATUS_BLOCK>
+                                      io_status_block_ptr,
```
benvanik: did clang-format do this? I haven't seen it split args like that before

DrChat: Nope, I can't get it to run. I did it cause putting it all together is >80 cols

src/xenia/kernel/xboxkrnl_io.cc
```diff
@@ -543,10 +547,18 @@ SHIM_CALL NtQueryInformationFile_shim(PPCContext* ppc_context,
         */
         result = X_STATUS_UNSUCCESSFUL;
         break;
+      case XFileSectorInformation:
+        // TODO: Return sector this file's on.
```
benvanik: file->device()->QuerySizeInfo() gives you bytes_per_sector

src/xenia/kernel/xboxkrnl_io.cc
benvanik: hmm we should figure that out. If you install the VS plugin then the xb format script should work and you can do a hotkey in VS to format as well (makes writing code much easier). See https://github.com/benvanik/xenia/blob/master/docs/style_guide.md

Sleepy93:
I often met NtQueryInformationFile(XFileSectorInformation) unimplemented in the log files.

Refs: 
https://github.com/xenia-project/game-compatibility/issues/800
https://github.com/xenia-project/game-compatibility/issues/780
https://github.com/xenia-project/game-compatibility/issues/445

etc... etc.. 

Are there any process to figure out this issue?

PR #258, opened 2015-06-14
Fullscreen support with F6

DrChat:
* Also adds the option to the window submenu


sgtfrankieboy:
Wouldn't it be better to use F11 for fullscreen? That's the default in Windows.

DrChat:
Probably, I just chose the next available hotkey. Let me change it.

benvanik:
Neat. We'll probably need to black-bar the swaps as most monitors aren't 16:9.

PR #261, opened 2015-06-17
Add missing includes for cstring

SirCmpwn:
`std::memcpy`, `std::memset`, etc are available in the `cstring` header. On Windows it works anyway, but on other platforms you have to explicitly include these.


benvanik:
Please order the includes alphabetically; we're following the style guide here: https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Names_and_Order_of_Includes

(in most cases clang-format with the Google setting is enough, however it does not reorder includes)

SirCmpwn:
Can do.

SirCmpwn:
Force pushed a fixed revision.

PR #264, opened 2015-06-18
gl4: command_processor: fix locations in geometry shader header

sephiroth99:
While trying to debug an OpenGL error on AMD, I found out that the following was logged in the pipeline infolog after a validation of the pipeline:

    in_vtx.o[0]'s data type or qualifiers doesn't match between vertex shader and geometry shader
    draw_id's data type or qualifiers doesn't match between geometry shader and fragment shader

Basically, as described in the commit message, the Geometry shader has the wrong layout value for its VertexData input/output, compared to the Vertex and Pixel shaders.

Unfortunately, this does not fix the error I was debugging...

Also, this should probably be tested before merging by someone who can display something (i.e. someone with a nVidia card), to make sure it still works. I was not able to, for obvious reasons...

### DIFF
--- a/src/xenia/gpu/gl4/command_processor.cc
+++ b/src/xenia/gpu/gl4/command_processor.cc
@@ -276,8 +276,8 @@ bool CommandProcessor::SetupGL() {
       "  vec4 o[16];\n"
       "};\n"
       "\n"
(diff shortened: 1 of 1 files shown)


benvanik:
Seems to work!

sephiroth99:
Awesome, thanks!

PR #265, opened 2015-06-19
gl4: more color formats

raven02:


raven02:
Wonder if glReadpixels used frequently here and worthwhile implement PBO ?

raven02:
Regarding two sides stencil , do we need the following ?

     glStencilMaskSeparate ()

PR #269, opened 2015-06-20
gl4: clear color/depth/stencil buffer bit

raven02:


raven02:
I didn't come across the glClearNamedFramebufferfv() , probably doing the same thing ?

raven02:
glClearNamedFramebuffer() should be handled the glClear call.

PR #273, opened 2015-06-21
GL4 Translator: ADDs/MULs/SUBs XZ instead of XY. Loops on FLOW_CONTROL.

DarioSamo:
Getting some good results with games that involve rigging with FLOW_CONTROL on.
Sonic Unleashed Demo: http://i.imgur.com/mndwqWd.jpg

### DIFF
--- a/src/xenia/gpu/gl4/gl4_shader_translator.cc
+++ b/src/xenia/gpu/gl4/gl4_shader_translator.cc
@@ -105,6 +105,17 @@ std::string GL4ShaderTranslator::TranslateVertexShader(
   for (uint32_t n = 0; n <= temp_regs; n++) {
     Append("  vec4 r%d = state.float_consts[%d];\n", n, n);
   }
+
+#if FLOW_CONTROL
+  // Add temporary integer registers for loops that we may use.
+  // Each loop uses an address, counter, and constant
+  // TODO: Implement only for the used loops in the shader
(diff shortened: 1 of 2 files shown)


MoochMcGee:
Why is this behind a macro? If it's a non-hacky improvement, why isn't it just there in the code?

DarioSamo:
The macro was already there behind flow control for stuff like conditional jumps, loops being implemented outside of the macro would have no effect on anything else.

The only one I have doubts about is the one that marks whether to add the relative offsets to the constants or not, so I left it behind the macro for now.

src/xenia/gpu/gl4/gl4_shader_translator.cc
```diff
@@ -783,7 +800,7 @@ bool GL4ShaderTranslator::TranslateALU_ADDs(const instr_alu_t& alu) {
   AppendScalarOpSrcReg(alu, 3);
   Append(".x + ");
   AppendScalarOpSrcReg(alu, 3);
-  Append(".y");
+  Append(".z");
```
benvanik: you sure z is right here?

DarioSamo:
Most of the swizzles I've seen from the translator convert those types of instructions like r0.yw to r0.yyww, so x and z would be correct in this case. I do have my doubts about that one since I don't have many games to test with tho.

PR #274, opened 2015-06-22
gl4 : clear depth and stencil seperately using the DSA version

raven02:
Tested with RR 6 (w/ forcing depth_color_enabled)

### DIFF
--- a/src/xenia/gpu/gl4/command_processor.cc
+++ b/src/xenia/gpu/gl4/command_processor.cc
@@ -2871,21 +2871,16 @@ bool CommandProcessor::IssueCopy() {
     // TODO(benvanik): verify format.
     GLfloat depth = {(copy_depth_clear & 0xFFFFFF00) / float(0xFFFFFF00)};
     GLint stencil = copy_depth_clear & 0xFF;
-    GLint old_draw_framebuffer;
     GLboolean old_depth_mask;
     GLint old_stencil_mask;
-    glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &old_draw_framebuffer);
     glGetBooleanv(GL_DEPTH_WRITEMASK, &old_depth_mask);
     glGetIntegerv(GL_STENCIL_WRITEMASK, &old_stencil_mask);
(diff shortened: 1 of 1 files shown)


src/xenia/gpu/gl4/command_processor.cc
```diff
@@ -2871,21 +2871,16 @@ bool CommandProcessor::IssueCopy() {
...
-    glBindFramebuffer(GL_DRAW_FRAMEBUFFER, source_framebuffer->framebuffer);
     glDepthMask(GL_TRUE);
     glStencilMask(0xFF);
-    // HACK: this should work, but throws INVALID_ENUM on nvidia drivers.
-    /* glClearNamedFramebufferfi(source_framebuffer->framebuffer,
-                             GL_DEPTH_STENCIL,
-                             depth, stencil);*/
-    glClearBufferfi(GL_DEPTH_STENCIL, 0, depth, stencil);
-    glBindFramebuffer(GL_DRAW_FRAMEBUFFER, old_draw_framebuffer);
+    glClearNamedFramebufferfv(source_framebuffer->framebuffer, GL_DEPTH,
```
benvanik: Did you try the combined glClearNamedFramebufferfi GL_DEPTH_STENCIL? It's what's supposed to work for this, but the nvidia drivers were busted. It'd be nice to know if they fixed it, and if they haven't we should leave the HACK comment in as doing separate clears on depth and stencil is not good.

raven02: Yes , i tried the glClearNamedFramebufferfi GL_DEPTH_STENCIL but looks like still not working yet .

I also tried to old way to glClear GL_STENCIL_BUFFER_BIT and GL_DEPTH_BUFFER_BIT, it also works.

benvanik: Lame! Add back the comment or a TODO to try the combined one again in the future, then :+1:

raven02:
Newer Nvidia driver just out today and i will test it out soon .It fixes some rendering issues from the D3D12 backend that developed for RPCS3.

PR #277, opened 2015-06-23
Kernel regression fix

raven02:
Fixes Dragon Ball Z Raging Blast 
Fixes Dragon Ball Z Raging Blast 2


gibbed:
- Misleading to call this a regression, MmQueryStatistics is not implemented correctly to begin with.
- Don't mention a specific game in the comments like that, if Raging Blast/2 has particular issues with the way MmQueryStatistics is implemented, open an issue for it so we can track it properly.
- How is Raging Blast/2 dying with available_pages set to 0x00013300?

raven02:
Let me open an issue instead of pull regression.

PR #279, opened 2015-06-24
gl4: Implement texture max anisotropy

raven02:


raven02:
looks like missing texture fetch

PR #281, opened 2015-06-24
gl4: implement texture max anisotropy

raven02:


archshift:
For what it's worth, instead of opening new PRs every time, you can just force push `git push -f [upstream] [branch]` back to the same branch as you opened a PR for to update the PR.

raven02:
Should be fine now .Previously got some problem rebasing and even push to master

PR #290, opened 2015-06-26
gl4: more texture formats

raven02:


benvanik:
Looks like some errors: http://build.xenia.jp/builders/pull-request-builds/builds/34/steps/build%20debug/logs/stdio

Did you mean R...32I?

Also, do you know what the interlaced formats are? I'm not sure exactly how they are treated by GL.

raven02:
Yep , it is R32I

I didn't come across interlace format in rpcs3 and ppsspp either .

hlide:
RxxI and RxxUI mean "signed integer" and "unsigned integer". Instead of storing and fetching components in a texture as a vector of float, you can do as a vector of sign/zero-extended integer.

So in a shader you must use "isamplerXXX/usamplerXXX" instead of "samplerXXX" and "ivec/uvec" to fecth them through "textureXXX".

PR #298, opened 2015-06-27
gl4: fix glsl version detection when there is no vendor specific info

sephiroth99:
When there is no vendor specific info, there is no space. Fixes a crash when that happens.


benvanik:
Oo, thanks!

PR #300, opened 2015-06-28
Swap to using HMODULE instead of handles for xex modules.

DrChat:
* Fixes anything that dereferences HMODULE (GetProcAddress is a big one)


robhshu:
Your changes to byte_order.h look incorrect - you should be modifying the `value` member?

DrChat:
I can't touch be's value directly because it's big endian.
Also, I don't think a non-const operator would work, it'd be modifying a copy of the value rather than the value itself.

PR #319, opened 2015-07-02
Use SSE intrinsics for uint16_t byte swapping

raven02:
Part of https://github.com/benvanik/xenia/issues/308

### DIFF
--- a/src/xenia/base/memory_generic.cc
+++ b/src/xenia/base/memory_generic.cc
@@ -43,7 +43,14 @@ void copy_and_swap_16_aligned(uint16_t* dest, const uint16_t* src,
 
 void copy_and_swap_16_unaligned(uint16_t* dest, const uint16_t* src,
                                 size_t count) {
-  for (size_t i = 0; i < count; ++i) {
+  size_t i;
+  for (i = 0; i + 8 <= count; i += 8) {
+    __m128i s = _mm_loadu_si128((__m128i*)&src[i]);
+    __m128i d = _mm_or_si128(_mm_slli_epi16(s, 8), _mm_srli_epi16(s, 8));
+    _mm_storeu_si128((__m128i*)&dest[i], d);
+  }
+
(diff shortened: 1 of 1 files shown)


src/xenia/base/memory_generic.cc
```diff
@@ -43,7 +43,14 @@ void copy_and_swap_16_aligned(uint16_t* dest, const uint16_t* src,
 
 void copy_and_swap_16_unaligned(uint16_t* dest, const uint16_t* src,
                                 size_t count) {
-  for (size_t i = 0; i < count; ++i) {
+  size_t i;
+  for (i = 0; i + 8 <= count; i += 8) {
+    __m128i s = _mm_loadu_si128((__m128i*)&src[i]);
```
benvanik: Can you make sure these are emitting v prefixed AVX variants? (debug and look at disasm)
I'm not actually sure how VC++ handles that (I believe we are compiling with AVX mode, so it *should* be v prefixed).

raven02: Yes , here is disasm

    50:     input = _mm_loadu_si128((__m128i*)&src[i]);
00007FF63EBD7980  vmovdqu     xmm0,xmmword ptr [rcx+rax]  
    51:     output = _mm_or_si128(_mm_slli_epi16(input, 8), _mm_srli_epi16(input, 8));
00007FF63EBD7985  vpsllw      xmm1,xmm0,8  
00007FF63EBD798A  vpsrlw      xmm0,xmm0,8  
00007FF63EBD798F  vpor        xmm2,xmm1,xmm0  
00007FF63EBD7993  add         r9,8  
00007FF63EBD7997  add         r10,8  
    52:     _mm_storeu_si128((__m128i*)&dest[i], output);
00007FF63EBD799B  vmovdqu     xmmword ptr [rax],xmm2

benvanik:
:+1: thanks both of you!

PR #321, opened 2015-07-03
gl4: Swap high and low 16 bits within a 32 bit word

raven02:


src/xenia/gpu/gl4/texture_cache.cc
```diff
@@ -687,14 +687,23 @@ void TextureSwap(Endian endianness, void* dest, const void* src,
...
-    case Endian::k16in32:
-      // TODO(benvanik): make more efficient.
-      /*for (uint32_t i = 0; i < length; i += 4, src += 4, dest += 4) {
-        uint32_t value = *(uint32_t*)src;
-        *(uint32_t*)dest = ((value >> 16) & 0xFFFF) | (value << 16);
-      }*/
-      assert_always("16in32 not supported");
-      break;
+    case Endian::k16in32:  // Swap high and low 16 bits within a 32 bit word
+    {
```
benvanik: It'd be nice to move this out with the other memory functions as copy_and_swap_16_in_32_aligned or something - that way we aren't putting a bunch of xmm stuff in here, and it can be reused easier.

raven02: Sure . will do .

PR #322, opened 2015-07-04
XEX2 Code Refactoring

DrChat:
* Removes most dependencies on xex2.c/.h
* Stores the xex2 header in host(XexModule) and guest(XUserModule) memory for direct access
 * Added a bunch of structs for direct access to xex2 headers
* If an import function is unresolved, rewrite the thunk to trigger a PPC trap and return to caller.

* Bad things:
 * Due to the way we're now resolving imports, we can't dump them to debug out anymore (except when loading the module). Should be fine though, the loading code will warn if importing an unimplemented variable/function.

### DIFF
--- a/src/xenia/base/byte_order.h
+++ b/src/xenia/base/byte_order.h
@@ -12,6 +12,7 @@
 
 #include <cstdint>
 
+#include "xenia/base/assert.h"
 #include "xenia/base/platform.h"
 
 #if XE_PLATFORM_MAC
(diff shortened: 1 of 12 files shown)


benvanik:
Wow, looks really good!

I'd like to see the existing module dumping code working, though - that's incredibly valuable to me. Can you please bring that bit back?

DrChat:
It dumps most of the headers - which headers did you want specifically?

benvanik:
All of them? :)

I don't see the exe info (address, entry point, etc), resource info, sections, and most importantly the imports.

DrChat:
It dumps exe info headers, but yeah I can add resource info and sections. Problem with imports is I can't do it without bringing back old code - you sure warnings on undefined imports won't suffice?

benvanik:
Yeah, resource info and sections are helpful (as it makes it easier to cross reference the addresses and see why the game may be hitting them).

Nope, not the same thing - being able to quickly see what a game is using at a glance is extremely valuable when trying to pin down issues that are caused by bad implementations of things, not just missing things.

benvanik:
Awesome, thanks! :heart:

PR #328, opened 2015-07-05
Implement XamLoaderLaunchTitle/XamLoaderTerminateTitle

DrChat:
* Now games can infinitely reload themselves!
* Makes valve games work (which use an app chooser and reload themselves when the user chooses a game)

### DIFF
--- a/src/xenia/cpu/xex_module.cc
+++ b/src/xenia/cpu/xex_module.cc
@@ -297,11 +297,11 @@ bool XexModule::Load(const std::string& name, const std::string& path,
 
 bool XexModule::Unload() {
   // Just deallocate the memory occupied by the exe
-  uint32_t exe_address = 0;
+  xe::be<uint32_t>* exe_address = 0;
   GetOptHeader(XEX_HEADER_IMAGE_BASE_ADDRESS, &exe_address);
   assert_not_zero(exe_address);
 
-  memory()->LookupHeap(exe_address)->Release(exe_address);
+  memory()->LookupHeap(*exe_address)->Release(*exe_address);
 
   assert_not_null(xex_header_); // Unloading a module that wasn't loaded?
   delete[] xex_header_;
--- a/src/xenia/cpu/xex_module.cc
+++ b/src/xenia/cpu/xex_module.cc
@@ -44,7 +44,8 @@ XexModule::XexModule(Processor* processor, KernelState* kernel_state)
       xex_header_(nullptr),
       base_address_(0),
       low_address_(0),
-      high_address_(0) {}
+      high_address_(0),
+      loaded_(false) {}
 
 XexModule::~XexModule() {
   xe_xex2_dealloc(xex_);
@@ -215,7 +216,10 @@ bool XexModule::Load(const std::string& name, const std::string& path,
 
 bool XexModule::Load(const std::string& name, const std::string& path,
                      xe_xex2_ref xex) {
+  assert_false(loaded_);
+  loaded_ = true;
   xex_ = xex;
+
   auto header = xex_header_;
   auto old_header = xe_xex2_get_header(xex_);
 
@@ -296,6 +300,11 @@ bool XexModule::Load(const std::string& name, const std::string& path,
 }
 
 bool XexModule::Unload() {
+  if (!loaded_) {
+    return true;
+  }
+  loaded_ = false;
+
   // Just deallocate the memory occupied by the exe
   xe::be<uint32_t>* exe_address = 0;
   GetOptHeader(XEX_HEADER_IMAGE_BASE_ADDRESS, &exe_address);
--- a/src/xenia/cpu/xex_module.cc
+++ b/src/xenia/cpu/xex_module.cc
@@ -41,20 +41,12 @@ XexModule::XexModule(Processor* processor, KernelState* kernel_state)
       processor_(processor),
       kernel_state_(kernel_state),
       xex_(nullptr),
-      xex_header_(nullptr),
       base_address_(0),
       low_address_(0),
       high_address_(0),
       loaded_(false) {}
 
-XexModule::~XexModule() {
-  xe_xex2_dealloc(xex_);
-
-  if (xex_header_) {
-    delete[] xex_header_;
-    xex_header_ = nullptr;
-  }
-}
+XexModule::~XexModule() { xe_xex2_dealloc(xex_); }
 
 bool XexModule::GetOptHeader(const xex2_header* header, xe_xex2_header_keys key,
                              void** out_ptr) {
@@ -89,7 +81,7 @@ bool XexModule::GetOptHeader(const xex2_header* header, xe_xex2_header_keys key,
 }
 
 bool XexModule::GetOptHeader(xe_xex2_header_keys key, void** out_ptr) const {
-  return XexModule::GetOptHeader(xex_header_, key, out_ptr);
+  return XexModule::GetOptHeader(xex_header(), key, out_ptr);
 }
 
 const xex2_security_info* XexModule::GetSecurityInfo(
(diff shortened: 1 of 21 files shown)


src/xenia/debug/debugger.cc
```diff
@@ -134,67 +135,70 @@ void SendResponse(SOCKET client_socket, flatbuffers::FlatBufferBuilder& fbb,
...
-      sockaddr_in6 client_addr;
-      int client_count = sizeof(client_addr);
-      SOCKET client_socket_id =
-          accept(listen_socket_, reinterpret_cast<sockaddr*>(&client_addr),
-                 &client_count);
-      if (client_socket_id == INVALID_SOCKET) {
-        XELOGE("Failed to accept socket");
-        continue;
-      }
+  if (!accept_thread_running_) {
```
benvanik: If you do `if (accept_thread_running_) { return ; }` here then we can avoid indenting the whole function one level :)

src/xenia/kernel/objects/xthread.cc
```diff
@@ -335,12 +333,11 @@ X_STATUS XThread::Create() {
 }
 
 X_STATUS XThread::Exit(int exit_code) {
+  assert_true(XThread::GetCurrentThread() == this);
+
   // TODO(benvanik): set exit code in thread state block
 
   // TODO(benvanik); dispatch events? waiters? etc?
-  if (event_) {
```
benvanik: I thought this was needed - is it not now that the GetWaitHandle is returning the thread handle?

src/xenia/kernel/objects/xthread.h
```diff
@@ -156,6 +160,8 @@ class XThread : public XObject {
   uint32_t pcr_address_;
   uint32_t thread_state_address_;
   cpu::ThreadState* thread_state_;
+  bool guest_thread_;  // Launched into guest code?
```
benvanik: Why the question? Is there a fuzzy line between host/guest threads? (like host created threads for guests ?)

src/xenia/kernel/objects/xuser_module.cc
```diff
@@ -104,12 +104,22 @@ X_STATUS XUserModule::LoadFromMemory(const void* addr, const size_t length) {
...
   this->xex_module()->GetOptHeader(XEX_HEADER_DEFAULT_STACK_SIZE, &stack_size_);
+  dll_module_ = !!(header->module_flags & XEX_MODULE_DLL_MODULE);
 
   OnLoad();
 
   return X_STATUS_SUCCESS;
 }
 
+X_STATUS XUserModule::Unload() {
+  if (!xex_module()->Unload()) {
```
benvanik: This function should probably have a check to make sure it's not redundant (for manual Unload() followed by the dtor Unload()) -- otherwise anyone listening for OnUnload() will get called twice, possibly.

src/xenia/kernel/xam_info.cc
```diff
@@ -90,53 +92,94 @@ SHIM_CALL XamGetExecutionId_shim(PPCContext* ppc_context,
   SHIM_SET_RETURN_32(X_STATUS_SUCCESS);
 }
 
-SHIM_CALL XamLoaderSetLaunchData_shim(PPCContext* ppc_context,
-                                      KernelState* kernel_state) {
-  uint32_t data_ptr = SHIM_GET_ARG_32(0);
-  uint32_t data_size = SHIM_GET_ARG_32(1);
+dword_result_t XamLoaderSetLaunchData(lpvoid_t data, dword_t size) {
+  auto xam_module = kernel_state()->GetModule("xam.xex");
```
benvanik: We should add a helper like GetKernelMode<T> that does this for you.

src/xenia/kernel/xam_info.cc
```diff
@@ -90,53 +92,94 @@ SHIM_CALL XamGetExecutionId_shim(PPCContext* ppc_context,
...
+dword_result_t XamLoaderGetLaunchDataSize(lpdword_t size_ptr) {
+  auto xam_module = kernel_state()->GetModule("xam.xex");
+  auto xam = kernel::object_ref<kernel::XamModule>(
+      reinterpret_cast<kernel::XamModule*>(xam_module.release()));
 
-  SHIM_SET_RETURN_32(1);
+  *size_ptr = xam->loader_data().launch_data_size;
+
+  // FIXME: What do we return?
+  return 1;
```
benvanik: this should be 0 for success

src/xenia/kernel/xam_info.cc
```diff
@@ -90,53 +92,94 @@ SHIM_CALL XamGetExecutionId_shim(PPCContext* ppc_context,
...
+        kernel_memory()->TranslateVirtual(loader_data.launch_data_ptr);
 
-  SHIM_SET_RETURN_32(0);
+    uint32_t copy_size =
+        std::min(loader_data.launch_data_size, (uint32_t)buffer_size);
+
+    std::memcpy(buffer_ptr, loader_buffer_ptr, copy_size);
+  }
+
+  // FIXME: Unknown return value.
```
benvanik: also 0

src/xenia/kernel/xam_module.h
```diff
@@ -24,7 +24,25 @@ class XamModule : public XKernelModule {
...
 
+  struct LoaderData {
+    LoaderData() {
+      launch_data_ptr = 0;
+      launch_data_size = 0;
+      launch_flags = 0;
+      launch_path = "";
+    }
+
+    uint32_t launch_data_ptr;
```
benvanik: (nit)
We can use C++ default initializers - I've fallen in love with them - so this would become:
```
struct LoaderData {
  uint32_t launch_data_ptr = 0;
  ...
  std::string launch_path; // no need to init, as already ""
};
```

I'm moving things over to this style as I see them.

src/xenia/cpu/xex_module.cc
```diff
@@ -41,11 +41,19 @@ XexModule::XexModule(Processor* processor, KernelState* kernel_state)
...
       base_address_(0),
       low_address_(0),
       high_address_(0) {}
 
-XexModule::~XexModule() { xe_xex2_dealloc(xex_); }
+XexModule::~XexModule() {
+  xe_xex2_dealloc(xex_);
+
+  if (xex_header_) {
+    delete[] xex_header_;
```
benvanik: We should switch this to an std::vector so we don't need any of the cleanup code.
vector can be used like a dumb pointer with `.data()`, so things like:
`std::memcpy(xex_header_.data(), ...)` and such are possible. `xex_header_.resize(new_count);` will reallocate as needed.

benvanik:
This is an awesome set of work! And thank you thank you thank you for splitting it into changes like that. I've been git bisecting a lot lately and it's extremely helpful to have good working intermediate changes!

src/xenia/kernel/objects/xthread.cc
DrChat: Yeah - dunno why we were creating an event.

src/xenia/debug/debugger.cc
DrChat: After this chunk it'll wait for a debugger to attach though

src/xenia/kernel/objects/xthread.h
DrChat: Didn't mean that to sound like it's fuzzy, it just literally means the thread was launched straight into guest code.

src/xenia/debug/debugger.cc
benvanik: Ah true - that's cause I was lazy with this function. Will clean it up later :)

src/xenia/kernel/objects/xthread.h
benvanik: Hah I gotcha - the question mark make me read that as 'Maybe it means it was launched into guest code? *shrug*' :P

src/xenia/cpu/xex_module.cc
DrChat: Bleh, that sounds kind of messy. It'd confuse the crap out of me (and possibly others) later as well.

benvanik: Not messy - quite the opposite - no need for this block of code at all, no need for nullptr init, and the malloc is handled under the covers. In general, anything we can do to remove sources of leaks, uninitialized memory, etc is a win. It's kind of what STL is for :) If we find someplace where the STL type isn't appropriate for some reason we should then at least roll our own that has the same properties, however *that's* what leads to confusion. It's like using std::string instead of naked char*.

benvanik:
:hamburger:

DrChat:
:fries:

PR #330, opened 2015-07-07
gl4: use flag depth_clear_enabled

raven02:
This probably still not sufficiently to clear the depth buffer properly like RR but this helps Raiden IV to render in-game -correctly 

raiden

### DIFF
--- a/src/xenia/gpu/gl4/command_processor.cc
+++ b/src/xenia/gpu/gl4/command_processor.cc
@@ -2849,8 +2849,8 @@ bool CommandProcessor::IssueCopy() {
   // TODO(benvanik): figure out real condition here (maybe when color cleared?)
   // HACK: things seem to need their depth buffer cleared a lot more
   // than as indicated by the depth_clear_enabled flag.
-  // if (depth_clear_enabled) {
-  if (depth_target != kAnyTarget) {
+  if (depth_clear_enabled) {
+    // if (depth_target != kAnyTarget) {
     // Clear the current depth buffer.
     // TODO(benvanik): verify format.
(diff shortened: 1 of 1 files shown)


src/xenia/gpu/gl4/command_processor.cc
```diff
@@ -2849,8 +2849,8 @@ bool CommandProcessor::IssueCopy() {
   // TODO(benvanik): figure out real condition here (maybe when color cleared?)
   // HACK: things seem to need their depth buffer cleared a lot more
   // than as indicated by the depth_clear_enabled flag.
-  // if (depth_clear_enabled) {
-  if (depth_target != kAnyTarget) {
+  if (depth_clear_enabled) {
+    // if (depth_target != kAnyTarget) {
```
benvanik: We want to keep that other condition as well, as otherwise it'll end up clearing random buffers. This should be an &&.

raven02: Yep, i will then do it as 

          if (depth_clear_enabled && depth_target != kAnyTarget) {

raven02: Just retested. Unfortunately this doesn't work for Raiden VI

raven02: Just wonder is ModeControl::kDepth mean clear depth ?

raven02: @benvanik , looks like && is not working here for Raiden IV.

Just wonder why game u tested that need "depth_target != kAnyTarget"?

raven02:
This also fixed in-game graphic in Sonic The Hedgehog 4 Episode 1

benvanik:
Merged in with the conditional from the comments.

PR #336, opened 2015-07-12
gl4: 565 texture fix and GL_BGRA for 8888 texture

raven02:
This fixes wrong color in Naruto Shippuden Ultimate Ninja Storm 2 and Sonic CD


raven02:
Rebase again and should be okay now.

raven02:
@benvanik , forget to ask . this commit looks okay or need some changes?

benvanik:
Merged in the _REV changes. I'm not sure about the second part of this change. We should split that out and figure out why that would be changing behavior (just ignoring swizzle information on any rgba8888 surface seems wrong).

PR #352, opened 2015-07-17
Fix compilation error and Update Python

Zangetsu38:
@benvanik Do not accept it right away.
Accept the other first pull request https://github.com/xenia-project/elemental-forms/pull/24 . 
I will update when the one that will be good for the Submodule :)

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -10,7 +10,7 @@ video drivers for your card.
 
 * Windows 8 or 8.1
 * Visual Studio 2015
-* [Python 2.7](http://www.python.org/download/releases/2.7.6/)
+* [Python 2.7](https://www.python.org/downloads/release/python-2710/)
 * If you are on Windows 8, you will also need the [Windows 8.1 SDK](http://msdn.microsoft.com/en-us/windows/desktop/bg162891)
 
 Ensure Python is in your PATH (`C:\Python27\`).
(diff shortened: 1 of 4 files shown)


Razzile:
iirc TargetPlatformVersion as 8.1 may be needed because before that was added I couldn't target 8.1 at all and couldn't run the built product.
see https://github.com/benvanik/xenia/issues/267#issuecomment-113730375

Zangetsu38:
@Razzile This is precisely the opposite, it is important not to put "TargetPlatformVersion", otherwise it can no longer compiled on Windows 10 with VS 2015.
Visual Strudio 2015 RC sdk install only 10.
So if you put 8.1 ca not find the SDK and therefore ca can not compile.

gibbed:
The problem is that it will cause Xenia to be built for Windows 10 for developers on Windows 8, which is more of a problem.

Sinoxan:
You could try this to see if it works for both Windows 10 and 8.1

<TargetPlatformVersion Condition="'$(TargetPlatformVersion)' > ''">8.0</TargetPlatformVersion>

Zangetsu38:
No, precisely, do no target, that's all, if you use Windows 8.1 VS 2015 sdk 8.1, if you have Windows 10 will use the sdk 10, This stuff is useless, apart we caused the problem on Windows 10.

gibbed:
The problem is when someone has the Windows 10 SDK installed in a Windows 8.1 dev env, Visual Studio opts to (incorrectly) use the 10 SDK instead of 8.1 SDK.

Zangetsu38:
Sln removed, so now not compilled :(

PR #353, opened 2015-07-18
Implement ObLookupThreadByThreadId / ObOpenObjectByPointer

DrChat:
* And some other stuff too

* Should help xenia-project/game-compatibility#99


src/xenia/base/threading_win.cc
```diff
@@ -315,8 +315,7 @@ class Win32Thread : public Win32Handle<Thread> {
   ~Win32Thread() = default;
 
   void set_name(std::string name) override {
-    AssertCallingThread();
-    xe::threading::set_name(name);
+    xe::threading::set_name(handle_, name);
```
benvanik: We need that assert - you can only set the thread name from the thread, not from any other thread. The code calling this needs to change to do that.

DrChat: You sure?
The code calling this is RtlRaiseException. Game created a new thread with ExCreateThread and on the main thread raised an exception to set the new thread's name.

benvanik: See https://github.com/benvanik/xenia/blob/master/src/xenia/base/threading_win.cc#L52.

In this case, you're passing it a handle where it needs a thread ID (which comes from `GetThreadId(handle_)`, and is different than the thread handle). I seem to remember `GetThreadId` behaving oddly (and always return ...FFFE, which is 'current thread id'), so it'd be nice to make sure that's not happening here.

DrChat: That's GetCurrentThread that returns the psuedohandle. This is behaving correctly.

benvanik: You sure about that? Like I said, set_name takes a thread id, and you are passing a thread handle. Are you sure that they are the same? According to the docs they are not expected to be (in that you can have multiple handles that resolve to the same thread id).

hlide:
Id and handle are too different things : the value -1 and -2 are for handle, not for id. One is for current process, the other for current thread.

hlide:
and yes you can have several handles for a unique id.

hlide:
[From microsoft](https://msdn.microsoft.com/en-us/library/xcb2z8hs.aspx):
```
//
// Usage: SetThreadName (-1, "MainThread");
//
#include <windows.h>
const DWORD MS_VC_EXCEPTION=0x406D1388;

#pragma pack(push,8)
typedef struct tagTHREADNAME_INFO
{
   DWORD dwType; // Must be 0x1000.
   LPCSTR szName; // Pointer to name (in user addr space).
   DWORD dwThreadID; // Thread ID (-1=caller thread).
   DWORD dwFlags; // Reserved for future use, must be zero.
} THREADNAME_INFO;
#pragma pack(pop)

void SetThreadName( DWORD dwThreadID, char* threadName)
{
   THREADNAME_INFO info;
   info.dwType = 0x1000;
   info.szName = threadName;
   info.dwThreadID = dwThreadID;
   info.dwFlags = 0;

   __try
   {
      RaiseException( MS_VC_EXCEPTION, 0, sizeof(info)/sizeof(ULONG_PTR), (ULONG_PTR*)&info );
   }
   __except(EXCEPTION_EXECUTE_HANDLER)
   {
   }
}
```

so it is an id but there is a special value (-1) to say to get the id of the caller thread.

PR #361, opened 2015-07-22
Misc fixes

DrChat:
* Mostly cosmetic changes (but those help games run too, right?)

### DIFF
--- a/src/xenia/gpu/gl4/gl4_shader.cc
+++ b/src/xenia/gpu/gl4/gl4_shader.cc
@@ -346,7 +346,7 @@ bool GL4Shader::CompileProgram(std::string source) {
     // Note that we put the translated source first so we get good line numbers.
     FILE* f = fopen(file_name, "w");
     if (f) {
-      fprintf(f, translated_disassembly_.c_str());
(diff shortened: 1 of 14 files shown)


src/xenia/gpu/gl4/gl4_shader.cc
```diff
@@ -345,12 +345,13 @@ bool GL4Shader::CompileProgram(std::string source) {
...
     // Note that we put the translated source first so we get good line numbers.
     FILE* f = fopen(file_name, "w");
-    fprintf(f, "%s", translated_disassembly_.c_str());
-    fprintf(f, "\n\n");
-    fprintf(f, "/*\n");
-    fprintf(f, "%s", ucode_disassembly_.c_str());
-    fprintf(f, " */\n");
-    fclose(f);
+    if (f) {
+      fprintf(f, translated_disassembly_.c_str());
```
benvanik: want the "%s" here as that's a stock clang warning (using a non-const string as a format arg)

DrChat: Hah, got the other one but missed this one

PR #362, opened 2015-07-22
Fix race condition in XThread

DrChat:


src/xenia/kernel/objects/xthread.h
```diff
@@ -25,7 +25,7 @@ namespace kernel {
 class NativeList;
 class XEvent;
 
-constexpr uint32_t X_CREATE_SUSPENDED = 0x00000004;
+constexpr uint32_t X_CREATE_SUSPENDED = 0x00000001;
```
benvanik: One of those 'how did this ever work' moments.

DrChat: Constant wasn't even used before this, so it didn't matter

PR #388, opened 2015-08-16
(Experimental) ELF module support

DrChat:


src/xenia/cpu/elf_module.cc
```diff
@@ -0,0 +1,144 @@
+/**
+******************************************************************************
```
benvanik: bad indent on the file headers

src/xenia/cpu/elf_module.cc
```diff
@@ -0,0 +1,144 @@
...
+* Copyright 2015 Ben Vanik. All rights reserved.                             *
+* Released under the BSD license - see LICENSE in the root for more details. *
+******************************************************************************
+*/
+
+#include "xenia/cpu/elf_module.h"
+
+#include <memory>
+
+#include "xenia/cpu/processor.h"
```
benvanik: headers should be in alphabetical order (xenia/base/ followed by xenia/cpu)

going downstairs to go get the buildbot cpu installed so it'll say this for me :P

Issue #387, opened 2015-08-15
the Kernel Version is not correct - could be the reason for many unplayable games


pharrowboy:
xbox 360 requires a valid kernel for it to run games, i was able to launch a homebrew app called DashLaunch that would install a custom dashboard but it tells me the current kernel version is unsupported. being kernel 2.65535.65535.3 might very well cause a few games not to load correctly. fixing the kernel version so that its the same as an xbox 360 might be helpful.

pharrowboy:
![kernel](https://cloud.githubusercontent.com/assets/11796091/9291009/a531f972-4365-11e5-99b0-cdbbeb20188d.png)
heres a picture.

gibbed:
While it's true the Kernel version specified by Xenia isn't exactly correct, I have yet to see a single game that checks the Kernel version.

Quantum2:
I could not run a lot of my games on my RGH due to the kernel version being
too old, I had to update the dashboard or just give up, not sure if the
game itself performs any checks or if it was a dashboard function

2015-08-15 22:59 GMT+01:00 Rick Gibbed <notifications@github.com>:

> While it's true the Kernel version specified by Xenia isn't exactly
> correct, I have yet to see a single game that checks the Kernel version.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/benvanik/xenia/issues/387#issuecomment-131455875>.
>



-- 
Rafael Francisco

pharrowboy:
alot of games prompt for you to update if youre not on a supported kernel by the game. there is a minimum but no maximum, but if you read the kernel number i listed, the format is incorrect: example: >>> 2.0.17349.0 is the proper format. not 2.65535.65535.3. if you look at all the kernel versions of the past https://en.wikipedia.org/wiki/Xbox_360_system_software, 

2.0.xxxxxx.0   <<< proper format

gibbed:
That sounds like a mechanic of the dashboard, and not the game. If you have an example game that does this, please let us know.

pharrowboy:
If you google it people on the internet say their xbox 360 cant run a new release game without updating. 
http://www.ixtreme.net/ixtreme-lt-lite-touch-firmware-phat-support/31241-playing-new-games-old-dashboard.html

[...]

PR #389, opened 2015-08-16
Giving kernel version an actual number

Sinoxan:
The kernel is only updated every few months so it shouldn't be much trouble to keep this up to date.

Will hopefully help with things such as #387 if anything needs a real kernel version number.

### DIFF
--- a/src/xenia/kernel/xboxkrnl_module.cc
+++ b/src/xenia/kernel/xboxkrnl_module.cc
@@ -110,15 +110,15 @@ XboxkrnlModule::XboxkrnlModule(Emulator* emulator, KernelState* kernel_state)
   // XboxKrnlVersion (8b)
   // Kernel version, looks like 2b.2b.2b.2b.
   // I've only seen games check >=, so we just fake something here.
+  // Current version as of 30 April 2015, 2.0.17349.0
   uint32_t pXboxKrnlVersion = memory_->SystemHeapAlloc(8);
   auto lpXboxKrnlVersion = memory_->TranslateVirtual(pXboxKrnlVersion);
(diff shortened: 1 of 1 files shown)


gibbed:
If we're going to plug an actual kernel version in, this is PR is incomplete. There are other places the kernel version is provided to games from.

However, I would suggest avoiding setting any kernel versions until we know of an actual game that checks against the kernel version.

benvanik:
Agreed @gibbed, if games are checking the kernel version for a minimum number it's likely they are testing for features that we don't support anyway. In fact, it'd probably be better to force a minimum version to see what happens.

Husono:
Speculation: Most games that perform a kernel check won't boot at all or display a warning message unless the kernel is above a certain version. It wouldn't make sense for the game to do something like unlock extra functionality with newer kernel versions. Wouldn't it make more sense to have a high kernel version so games progress past the warning screen even if xenia may not support all the features of that kernel.

hlide:
I dare to say a kernel check in a game is less for extra functionality than forcing gamer to "upgrade" their XBOX kernel the same way they want you to accept new agreements. In the case of PSP, all hackers know that the numerous firmware upgrades were more dictated by anti-hacking than extra functionality.

benvanik:
That may have been the case there, but from what I've seen so far it is very much for extra functionality. Many games are backwards compatible and will check at runtime for functions from more modern kernels (lots of Live features like party mode and other matchmaking, kinect, etc). Others just fix their kernel version then assume those functions exist and have them in the import tables.

The version as specified now is 0xFF.0xFF, the max version, so all games should be getting through.

PR #407, opened 2015-08-30
Rewrote the Audio Decoder

DrChat:
* Can build libav ourselves now
* Decodes frame-by-frame, so no more weirdness if the game specifies a specific offset.
 * Also supports audio loops
* If a frame fails to decode, will skip to the next frame.
* Crapload of error-checking to make sure everything is running smoothly
* Code is now 100% more spaghettified!

* Should fix audio dropping out due to decoder errors. Unfortunately, there's a deadlock in XAudio2 which will still cause audio to die in some games.

### DIFF
--- a/src/xenia/app/premake5.lua
+++ b/src/xenia/app/premake5.lua
@@ -32,7 +32,7 @@ project("xenia-app")
   })
   includedirs({
     project_root.."/third_party/elemental-forms/src",
-	project_root.."/build_tools/third_party/gflags/src",
+    project_root.."/build_tools/third_party/gflags/src",
   })
   local_platform_files()
   files({
@@ -53,21 +53,6 @@ project("xenia-app")
       "xenia-hid-winkey",
       "xenia-hid-xinput",
     })
-
---[[
-  filter("configurations:Checked")
-    local libav_root = "../third_party/libav-xma-bin/lib/Debug"
-    linkoptions({
-      libav_root.."/libavcodec.a",
-      libav_root.."/libavutil.a",
-    })
-  filter("configurations:Debug or Release")
-    local libav_root = "../third_party/libav-xma-bin/lib/Release"
-    linkoptions({
-      libav_root.."/libavcodec.a",
-      libav_root.."/libavutil.a",
-    })
-]]
 
   filter("platforms:Windows")
     -- Only create the .user file if it doesn't already exist.
--- a/src/xenia/app/premake5.lua
+++ b/src/xenia/app/premake5.lua
@@ -32,6 +32,7 @@ project("xenia-app")
   })
   includedirs({
     project_root.."/third_party/elemental-forms/src",
+	project_root.."/build_tools/third_party/gflags/src",
   })
   local_platform_files()
   files({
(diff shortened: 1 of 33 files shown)


src/xenia/app/premake5.lua
```diff
@@ -32,6 +32,7 @@ project("xenia-app")
   })
   includedirs({
     project_root.."/third_party/elemental-forms/src",
+	project_root.."/build_tools/third_party/gflags/src",
```
benvanik: tab

src/xenia/app/premake5.lua
```diff
@@ -52,6 +53,8 @@ project("xenia-app")
       "xenia-hid-winkey",
       "xenia-hid-xinput",
     })
+
+--[[
```
benvanik: No longer needed - can kill this

src/xenia/apu/nop/premake5.lua
```diff
@@ -13,5 +13,6 @@ project("xenia-apu-nop")
   defines({
   })
   includedirs({
+	project_root.."/build_tools/third_party/gflags/src",
```
benvanik: tab (here and in other lua files)

src/xenia/apu/xma_context.cc
```diff
@@ -243,162 +346,319 @@ void XmaContext::DecodePackets(XMA_CONTEXT_DATA* data) {
...
-    uint32_t input_offset_bytes = seq_offset_bytes;
-    auto input_buffer = in0;
-
-    if (seq_offset_bytes >= input_size_0_bytes && input_size_1_bytes) {
-      // Size overlap, select input buffer 1.
-      // TODO(DrChat): This needs testing.
-      input_offset_bytes -= input_size_0_bytes;
-      input_buffer = in1;
-    }
+#ifdef DEBUG
```
benvanik: Doesn't seem too expensive - would be good to keep it always in. Avoiding behavioral changes in debug vs. release builds is really nice.

src/xenia/apu/xma_decoder.cc
```diff
@@ -55,7 +55,34 @@ XmaDecoder::XmaDecoder(cpu::Processor* processor)
 XmaDecoder::~XmaDecoder() = default;
 
 void av_log_callback(void* avcl, int level, const char* fmt, va_list va) {
-  xe::LogLineVarargs('A', fmt, va);
+#ifdef NDEBUG
```
benvanik: Would be nice to have this as a flag instead (like --verbose_libav_logging or something).

src/xenia/apu/xma_helpers.h
```diff
@@ -0,0 +1,48 @@
+/**
+******************************************************************************
```
benvanik: space

src/xenia/apu/xma_context.cc
DrChat: In practice these variables never change, so no point really in having the checks.

benvanik: If it's useful as an assert you could just change them to that (like the buffer size check?) - otherwise kill em :)

PR #408, opened 2015-08-30
vcmpbfp with Rc bit support

DrChat:
@benvanik You'll probably want to double check and make sure this works as intended.

### DIFF
--- a/src/xenia/cpu/frontend/ppc_emit_altivec.cc
+++ b/src/xenia/cpu/frontend/ppc_emit_altivec.cc
@@ -580,20 +580,23 @@ XEEMITTER(vcfpuxws128, VX128_3(6, 624), VX128_3)(PPCHIRBuilder& f,
 
 int InstrEmit_vcmpbfp_(PPCHIRBuilder& f, InstrData& i, uint32_t vd, uint32_t va,
                        uint32_t vb, uint32_t rc) {
+  // if vA or vB are NaN, the 2 high-order bits are set (0xC0000000)
   Value* va_value = f.LoadVR(va);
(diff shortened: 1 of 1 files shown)


src/xenia/cpu/frontend/ppc_emit_altivec.cc
```diff
@@ -580,9 +580,10 @@ XEEMITTER(vcfpuxws128, VX128_3(6, 624), VX128_3)(PPCHIRBuilder& f,
 
 int InstrEmit_vcmpbfp_(PPCHIRBuilder& f, InstrData& i, uint32_t vd, uint32_t va,
                        uint32_t vb, uint32_t rc) {
+  // if vA or vB are NaN, the 2 high-order bits are set (0xC0000000)
   Value* va_value = f.LoadVR(va);
   Value* vb_value = f.LoadVR(vb);
-  Value* gt = f.VectorCompareSGT(va_value, vb_value, FLOAT32_TYPE);
+  Value* gt = f.VectorCompareSGE(va_value, vb_value, FLOAT32_TYPE);
```
benvanik: Change 'gt' to 'ge' now that it's GE.

benvanik:
[qemu](http://copilotco.com/mail-archives/qemu.2009/msg02598.html) does an explicit check for NaN - do you think that's worth it, or does your comment mean this handles that?

DrChat:
I changed the top comment to what it is because I'm pretty sure x86's VectorCompareXX will always return true on NaN checks, right?

benvanik:
Probably :)

PR #417, opened 2015-09-03
Fix crash in memory allocation when there is no space to fit entire page

Rhodan81:
range

Fixes crash when loading SSX.


benvanik:
Nice! Please run 'xb format --all' to fix the style errors and re-push.

benvanik:
:+1: thank you!

PR #423, opened 2015-09-11
Pause Xenia if it crashes in guest code

DrChat:
* Should allow end users to differentiate between host/guest crashes easier.
* Allows developers to attach a debugger after Xenia crashes.


benvanik:
RE travis: weird, must be some linter changes. Otherwise, LGTM!

PR #435, opened 2015-09-22
Fixing some Linux build errors

sephiroth99:
This PR fixes some of the issues when trying to build under Linux with clang. All the third party stuff builds, and parts of xenia too (see issues tagged as cross platform for what doesn't build).

clang++ does not want to compile C code, so C projects were modified in their premake config and clang is used for these.

Three submodules have updates too, please also pull these. 
xenia-project/capstone
xenia-project/libav
xenia-project/elemental-forms

Please do tell if stuff is wrong and I will update ASAP.

For reference, I tested with Debian 8 and clang 3.5.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -45,7 +45,6 @@ filter("configurations:Debug")
     "_NO_DEBUG_HEAP=1",
   })
   runtime("Release")
-  linkoptions({"/NODEFAULTLIB:MSVCRTD"})
 
 filter("configurations:Release")
   runtime("Release")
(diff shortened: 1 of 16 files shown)


sephiroth99:
Hmm lint failed on some of the code I changed, but also on other code. Should I update?

benvanik:
Some good fixes! Will merge and clean up (and then fix travis - it looks like clang was updated and the formatter changed).

benvanik:
Merged with minor changes (some style stuff like header include order and c++-style casts). Since it was a rebasing you'll likely have to force pull your local branch.

Thanks! These are nice! :+1:

PR #437, opened 2015-09-26
Use Clang 3.8 for Linux build.

sephiroth99:
Specify Clang 3.8 for building under Linux. Fix a build error due to upgrading. Finally, update the Linux section of building.md. As mentionned, we can't build with GCC right now because Clang specific "no-error" flags are used. I guess compiler specific flags should be added in premake files.

### DIFF
--- a/building.md
+++ b/building.md
@@ -34,14 +34,13 @@ get helpful spacers/movs in the disassembly.
 
 ### Linux
 
-Linux support is extremely experimental and incomplete.
+Linux support is extremely experimental and presently incomplete.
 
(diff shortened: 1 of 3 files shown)


robhshu:
You could also remove the TODO message about enabling this from the build config (https://github.com/benvanik/xenia/blob/master/.travis.yml#L2) as part of this pull request:

```yml
# TODO(benvanik): re-enable when clang-3.8 is whitelisted.
# https://github.com/travis-ci/apt-package-whitelist/issues/474
#sudo: false
```

sephiroth99:
That comment was about the Travis setup, which was fixed in commit 2298fa8. It is true that the comment should be removed, but it should be in another PR as it is not really related to this PR.

PR #442, opened 2015-10-19
Use guard pages for the MMIO handler rather than no access pages

DrChat:
* Should fix BF1943 and other games that spam textures into memory.

### DIFF
--- a/src/xenia/base/exception_handler.h
+++ b/src/xenia/base/exception_handler.h
@@ -36,7 +36,8 @@ class Exception {
     code_ = Code::kIllegalInstruction;
     thread_context_ = thread_context;
   }
-  void InitializePageGuardViolation(X64Context* thread_context, uint64_t fault_address) {
+  void InitializePageGuardViolation(X64Context* thread_context,
+                                    uint64_t fault_address) {
     code_ = Code::kPageGuardViolation;
     thread_context_ = thread_context;
     fault_address_ = fault_address;
--- a/src/xenia/base/exception_handler.h
+++ b/src/xenia/base/exception_handler.h
@@ -23,6 +23,7 @@ class Exception {
     kInvalidException = 0,
     kAccessViolation,
     kIllegalInstruction,
+    kPageGuardViolation,
   };
 
   void InitializeAccessViolation(X64Context* thread_context,
@@ -34,6 +35,11 @@ class Exception {
   void InitializeIllegalInstruction(X64Context* thread_context) {
     code_ = Code::kIllegalInstruction;
     thread_context_ = thread_context;
+  }
+  void InitializePageGuardViolation(X64Context* thread_context, uint64_t fault_address) {
+    code_ = Code::kPageGuardViolation;
+    thread_context_ = thread_context;
+    fault_address_ = fault_address;
   }
 
   Code code() const { return code_; }
--- a/src/xenia/base/exception_handler_win.cc
+++ b/src/xenia/base/exception_handler_win.cc
@@ -55,6 +55,10 @@ LONG CALLBACK ExceptionHandlerCallback(PEXCEPTION_POINTERS ex_info) {
       ex.InitializeAccessViolation(
           &thread_context, ex_info->ExceptionRecord->ExceptionInformation[1]);
       break;
+    case STATUS_GUARD_PAGE_VIOLATION:
+      ex.InitializePageGuardViolation(
+          &thread_context, ex_info->ExceptionRecord->ExceptionInformation[1]);
+      break;
     default:
       // Unknown/unhandled type.
       return EXCEPTION_CONTINUE_SEARCH;
--- a/src/xenia/base/memory.h
+++ b/src/xenia/base/memory.h
@@ -28,11 +28,14 @@ size_t page_size();
 // This is likely 64KiB.
 size_t allocation_granularity();
 
-enum class PageAccess {
+enum PageAccess {
   kNoAccess = 0,
   kReadOnly = 1 << 0,
   kReadWrite = kReadOnly | 1 << 1,
   kExecuteReadWrite = kReadWrite | 1 << 2,
+  kGuardPage = 1 << 3,
+
+  kGuardReadWrite = kGuardPage | kReadWrite,
 };
 
 enum class AllocationType {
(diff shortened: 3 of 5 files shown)


DrChat:
This may not actually work as intended - closing until I can test this some more.

hlide:
Are you not mixing two different things as Guard Page and Write Watched Page?

As far as I know, Guard Page is just the same as No Access Page but with auto-setting the page accessible normally once after the first exception.

Write Watched Page is just raising an exception the first time a write access is done then auto-sets the page accessible normally (VirtualAlloc-ed with MEM_WRITE_WATCH). I Wonder whether if we can catch it through exception handler.

DrChat:
Yeah, I only need it to fire when a write happens, not when it's accessed...

Could check if it's a read and just drop the exception right away and reinstate the guards, but idk.
I just need the system to drop the protection attribute as soon as a write happens, so another thread won't get caught up and crash.

hlide:
you can use a RO access page and catch it through exception handler so be aware of the write attempt. The exception will set the page with RW access to allow write. When reaching a checkpoint you handle the set of written pages (like uploading texture, etc.) then set those page with RO access back. But that is more or less what functions `ResetWriteWatch` and `GetWriteWatch` do.
On the other end, if you need to set back the RO access after executing the faulty instruction. You could proceed this way too:
1) by default, page is RO access.
2) an instruction is writing to this page.
3) an exception raises: set RW access. Resume instruction by setting EFLAGS.TF to 1 so after executing the instruction back the sing step raises and set R0 access back.
Of course, if you have to memset/memcpy a whole page, it will slow down a lot.
The other possibility is a mix:  you only keep one page with RW access - whenever a write access is done in the same page, only the very first access will raise an exception. If the next access is done to another page, exception raises a set back RO access to the previous written page before setting RW access to the new page.
And so on.
Obviously, none of them are perfect. So you must evaluate which one is the best to your need.

hlide:
I see you suffer form a lack of synchronization between two threads accessing the same page.

DrChat:
Yeah, what we were already doing was setting the pages to RO access and handling the AV exceptions when they occurred. This works if a single thread is accessing said page at a time, but games like BF1943 have 4 or more threads uploading data into memory.

From the existing design, it looks as if it should work fine... but the app is crashing if more than one thread access the same memory. If I hit continue when VS intercepts the exception, Xenia will continue just fine.

hlide:
VS suspends the process when intercepting the exception so no wonder.

Could you elaborate what the conflicted threads are and their purpose? Is the crash due to no thread-safe code in the AV exceptions or due to a call to a WIN32 function?

It sounds like you can block any following exceptions on the same page until the first exception resolves it. The following exceptions will then simply resume execution of the faulty instruction with no need to handle the exception.

DrChat:
Okay, crap - It may just be crashing due to poor design in the code.

The code's all threadsafe - but when the first thread goes through the handler it'll hit the write watch region and then remove the watch. The other threads would be waiting on thread 1, and when they get their turns they'll abort because the region is no longer being watched.

One solution is just don't remove the watch, instead do so later when safe to do so.
Could also rig a way to see if any threads are waiting before removing a write watch.

Probably should have looked into this issue more before submitting a PR, but I was pretty eager to start coding again.

TranzRail:
Regarding the threading issue, if you have multiple threads all doing stuff on things that are shared, wouldn't some sort of thread pool be needed, with a manager essentially regulating who gets to R/W and when?

Not to hijack this thread, but @DrChat do you know if there's a place other than GitHub to discuss this project? I checked IRC and it seemed dead. A forum would be useful. I don't want to start working on something if someone else already is.

Thanks.

DrChat:
Can't do a thread pool - these are game-controlled threads. Would have to write a full scheduler if we wanted to go that route, which I certainly don't.

Also, @TranzRail - the IRC is alive and well. Just ask a question and wait for a response (it could be up to a day, be patient).

hlide:
I tried to read your source but I do not see how it is thread-safe in your PR and I do not see where the other threads are waiting on the first thread. I probably miss something. If the write protection is about MMIO range, then something is not correct: in no way the faulty instruction must be resumed by the other threads - they should all do the same as the first whatever it would be when reading or writing.

Unless you write protection is for something else MMIO.

DrChat:
They're waiting in MMIOHandler::CheckWriteWatch line 172 in mmio_handler.cc.

hlide:
```c++
bool MMIOHandler::CheckWriteWatch(void* thread_state, uint64_t fault_address) {
  ...
  std::list<WriteWatchEntry*> pending_invalidates;
  write_watch_mutex_.lock(); // <--- block the other threads !
  for (auto it = write_watches_.begin(); it != write_watches_.end();) {
    auto entry = *it;
    if (entry->address <= physical_address &&
        entry->address + entry->length > physical_address) {
      // Hit!
      pending_invalidates.push_back(entry);
      ClearWriteWatch(entry);
      auto erase_it = it;
      ++it;
      write_watches_.erase(erase_it);
      continue;
    }
    ++it;
  }
  write_watch_mutex_.unlock(); // <-- free the other threads !!!
  if (pending_invalidates.empty()) { // <-- not under mutex protection !!!!
    // Rethrow access violation - range was not being watched.
    return false;
  }
  while (!pending_invalidates.empty()) { // <-- not under mutex protection !!!!
    auto entry = pending_invalidates.back(); // <-- not under mutex protection !!!!
    pending_invalidates.pop_back(); // <-- not under mutex protection !!!!
    entry->callback(entry->callback_context, entry->callback_data,
                    physical_address);
    delete entry;
  }
  // Range was watched, so lets eat this access violation.
  return true;
}
```
Moreover, another thread may be awaken AFTER pending_invalidates.empty() is true which is totally wrong because instead of resuming the faulty instruction, it will crash because it ends up with EXCEPTION_CONTINUE_SEARCH instead of EXCEPTION_CONTINUE_EXECUTION.

DrChat:
Yeah, I think what I can do is just query the memory region and see if it still has protection constants when the thread acquires the lock.

hlide:
I have just realized that *pending_invalidates* is a local variable so it is not the issue. But yes, looking for the memory protection is the trick indeed.

PR #446, opened 2015-10-24
Fixed mtcrf which was wrongly implemented as mtcr

AlexAltea:
While the instruction `mtocrf` only allows to update one CR field at a time, the instruction `mtcrf` allows a subset of CR fields to be updated. However, Xenia's implementation of `mtcrf` updates all fields, ignoring the `FXM` instruction code. Only `mtcr rS` (simplified mnemonic for `mtcrf 0xFF, rS`) does that.

PS: See page 130 in the [Power ISA docs](https://www.power.org/wp-content/uploads/2012/07/PowerISA_V2.06B_V2_PUBLIC.pdf).
PSS: Since this is a small change, I haven't bothered to build Xenia, let alone running tests before submitting this PR.


benvanik:
Awesome! Thank you!

Razzile:
better late than never :P

PR #447, opened 2015-10-31
Fix ExCreateThread returning a handle when game requests a pointer

DrChat:


DrChat:
FYI: Failing lint check isn't in my code.

PR #449, opened 2015-11-04
Rectangle List Fix (projection issues on right-aligned vertices)

### LINKED ISSUE #304: Fix rect lists
Still seem to be drawing wrong.

DrChat:
* Should fix #304

### DIFF
--- a/src/xenia/gpu/gl4/command_processor.cc
+++ b/src/xenia/gpu/gl4/command_processor.cc
@@ -335,7 +335,7 @@ bool CommandProcessor::SetupGL() {
       "    }\n"
       "    EmitVertex();\n"
       "    EndPrimitive();\n"
-      "  } else if (gl_in[0].gl_Position.y == gl_in[1].gl_Position.y) {\n"
+      "  } else {\n"
       //  0 ------ 1
       //  | -      |
       //  |   \\   |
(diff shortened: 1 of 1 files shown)


src/xenia/gpu/gl4/command_processor.cc
```diff
@@ -335,7 +335,7 @@ bool CommandProcessor::SetupGL() {
       "    }\n"
       "    EmitVertex();\n"
       "    EndPrimitive();\n"
-      "  } else {\n"
+      "  } else if (gl_in[0].gl_Position.y == gl_in[1].gl_Position.y) {\n"
```
benvanik: what about the fall-through case? (if it's even possible?) this shader won't emit anything without the else

benvanik: (this may be ok, as the hardware may cull rects not matching these formats - just curious if you'd seen any missing rects in your experimenting)

DrChat: Yeah, it might be possible. If it falls through nothing will be drawn.

Depends on which will get our attention faster, no rects or corrupt rects on screen

raven02:
Just tested it with another games - Dynasty Warriors Gundam 3-  that have rect list issue .It got fixed as well.

benvanik:
SGTM!

PR #450, opened 2015-11-04
Indirection table fallback

DrChat:
* Not optimized, but shouldn't matter cause the indirection table is almost always available.


benvanik:
Are you finding yourself hitting this a lot? Unless other modules are injecting themselves into your process it shouldn't be likely to ever happen, I think.

DrChat:
Only hit this because I was debugging using Nvidia NSight. Still should be good to have a fallback anyways.

benvanik:
ah, that's a good enough reason for me! :+1:

PR #451, opened 2015-11-06
Fix a rare deadlock in the MMIO handler

DrChat:


benvanik:
Technically yes, though in this function the lock needs to be held for a very specific amount of time. To use RAII you'd need to wrap that all in a scope, but there's also some variables defined inside so it starts to get messy.
Both point to issues in the design, but the whole write watch stuff needs to be rewritten anyway so it's not urgent :P

Issue #292, opened 2015-06-27
Fix texture invalidation


benvanik:
Broke during the great memory rewrite, and it currently doesn't seem to work in atrain or smb.

DrChat:
This is fixed as of #454.

PR #454, opened 2015-11-09
Texture Invalidation Fix

DrChat:
* Fixes: *Everything*
 * Games were freeing texture memory and then reallocating texture memory in the same spot, which neutered the MMIO Handler's watching capabilities. We now purge the entry when the memory is freed or its access protection is changed.


raven02:
Not too sure if this should be same intention to fix for #292

PR #455, opened 2015-11-09
Partially implement XamUserWriteProfileSettings

sabretooth:
Handles writing binary blob settings to the user profile which can then be read
by XamUserReadProfileSettings.

This allows the game "Legend of Spyro: Dawn of the Dragon" to proceed past its "press start" screen. It would otherwise sit in an infinite loop calling XamUserWriteProfileSettings to write some config data and XamUserReadProfileSettings to verify the write.

The implementation is definitely not complete since it's based on observing the behaviour of only one game. That game only writes the BINARY type, so I wasn't able to deduce the format of any of the other datatypes from the structs that it passes to this function.

### DIFF
--- a/src/xenia/kernel/xam/xam_user.cc
+++ b/src/xenia/kernel/xam/xam_user.cc
@@ -253,6 +253,29 @@ SHIM_CALL XamUserReadProfileSettings_shim(PPCContext* ppc_context,
   }
 }
 
+typedef struct {
+  xe::be<uint32_t> from;
+  xe::be<uint32_t> unk_04;
+  xe::be<uint32_t> unk_08;
+  xe::be<uint32_t> unk_0c;
+  xe::be<uint32_t> setting_id;
+  xe::be<uint32_t> unk_14;
+
+  // UserProfile::Setting::Type. Appears to be 8-in-32 field, and the upper 24
+  // are not always zeroed by the game.
+  uint8_t type;
+
+  xe::be<uint32_t> unk_1c;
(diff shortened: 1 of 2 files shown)


src/xenia/kernel/xam/xam_user.cc
```diff
@@ -287,8 +308,56 @@ SHIM_CALL XamUserWriteProfileSettings_shim(PPCContext* ppc_context,
...
+
+  for (uint32_t n = 0; n < setting_count; ++n) {
+    const X_USER_WRITE_PROFILE_SETTING& settings_data = settings_data_arr[n];
+    XELOGD(
+        "XamUserWriteProfileSettings: setting index [%d]:"
+        " from=%d setting_id=%.8X data.type=%d",
+        n, (uint32_t)settings_data.from, (uint32_t)settings_data.setting_id,
+        settings_data.type);
+
+    xam::UserProfile::Setting::Type settingType =
```
benvanik: tip: auto in c++11 is nifty, makes this:

```
auto settingType = static_cast<xam::UserProfile::Setting::Type>(settings_data.type);
```

(as the static_cast names the type already)

sabretooth:
Thanks. I only recently moved to VS2015 (and got C++11 support) so I'm still getting used to having `auto` available.

benvanik:
Yeah it takes a bit of getting used to - very easy to overdo it and end up with totally unreadable code, but when used effectively it's so nice :)

PR #457, opened 2015-11-11
Shader relative constant fix

sabretooth:
This fixes smooth-skinned models' animations in Legend of Spyro: Dawn of the Dragon (and maybe other Criware games? though I don't have any handy to test).


raven02:
This fixes the character in Sonic The Hedgehog 4 Episode 1

PR #464, opened 2015-11-18
Disable GL_CULL_FACES when drawing rectangle lists

DrChat:
* Fixes black screen in Banjo Kazooie

### DIFF
--- a/src/xenia/gpu/gl4/draw_batcher.cc
+++ b/src/xenia/gpu/gl4/draw_batcher.cc
@@ -242,6 +242,7 @@ bool DrawBatcher::CommitDraw() {
 }
 
 bool DrawBatcher::Flush(FlushMode mode) {
+  GLboolean cull_enabled = 0;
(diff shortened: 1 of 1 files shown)


src/xenia/gpu/gl4/draw_batcher.cc
```diff
@@ -289,6 +290,8 @@ bool DrawBatcher::Flush(FlushMode mode) {
         // assert_true(
         // (register_file_->values[XE_GPU_REG_PA_SU_SC_MODE_CNTL].u32
         // & 0x3) == 0);
+        cull_enabled = glIsEnabled(GL_CULL_FACE);
```
benvanik: :scream: queries are really, really bad -- they seem like they'd be fast, but they never are. But, this code will all die eventually, so it's fine!

DrChat: :+1: That's the spirit!

PR #465, opened 2015-11-20
DXT1 Fix (alpha issues in certain games)

JohnGodgames:
Thx to  @JayFoxRox for help.https://www.opengl.org/wiki/S3_Texture_Compression#DXT1_with_1-bit_Alpha


raven02:
thanks. it fixed the alpha issues in bubble bobble live and Sonic Etc

JayFoxRox:
I think you'll have to run 'xb format' (according to the travis logs).

I'd expect the GPU having a flag to switch between RGB and RGBA DXT1 mode, but maybe ATI dropped it because the drivers could always disable the alpha using texture swizzles.
This seems to fix a lot of alpha issues so it's probably a good idea to change to RGBA until we know more.

(I'd probably call this "Use alpha version of DXT1" because it's precisely what it does.)

PR #466, opened 2015-11-20
GPU Tracefile Compression

DrChat:
* Compresses memory reads/writes with zlib if they're above a certain size.
* Significant compression ratio achieved!


src/xenia/gpu/premake5.lua
```diff
@@ -11,6 +11,7 @@ project("xenia-gpu")
     "xenia-base",
     "xenia-ui",
     "xxhash",
+	"zlib",
```
benvanik: spaces not tabs

DrChat: ffff stupid notepad++

PR #467, opened 2015-11-21
Add more texture info to trace viewer (texture format, swizzle)

JohnGodgames:
Thx to @JayFoxRox for it...


DrChat:
:+1: Neat, but could you convert the numbers to strings? Will probably need a table for that.

JayFoxRox:
Hold this change back, this output format is horrible. It was just a quick solution to find the DXT1 issue.
I'll probably make a clean patch in December if nobody else does it.

Textures should have a matrix for the texture swizzle in the preview window so the alpha channel can be visualized easily [rather than the number it is at the moment].
The texture format should be in text as @DrChat already said. There is probably more info which should be included such as the memory address or hash..

PR #468, opened 2015-11-22
Implemented REG_TO_MEM GPU opcode

Kraust:
I've added this opcode based upon what I assume is the proper implementation of it assuming it's name is a proper descriptor of what it does.

Arguments should be a register and a memory address. Swapping the arguments throws an out of range exception on the register side.

I've been sitting on this for a while, just figured that it might be the proper implementation. I could be totally wrong however.


benvanik:
Looks good! The register index should probably be truncated but I'm not sure I'm properly doing that elsewhere now, so I'll fix that up in a separate commit.

PR #469, opened 2015-11-22
Trace Viewer UI Improvements

DrChat:
* Allow seeking forward more than one frame (without playing from the beginning)
* Progress bar while seeking
 * Fixes crash with texture view open while seeking (masks it really - but that should be fine. See https://github.com/benvanik/xenia/commit/7a1d7bd652c1fc5d533a7e4cb532003fa141efe5)
* Disable command list and frame list UI while playing back a trace (to avoid queuing up a crapload of commands)

### DIFF
--- a/src/xenia/gpu/trace_player.h
+++ b/src/xenia/gpu/trace_player.h
@@ -10,6 +10,7 @@
 #ifndef XENIA_GPU_TRACE_PLAYER_H_
 #define XENIA_GPU_TRACE_PLAYER_H_
 
+#include <atomic>
 #include <string>
 
 #include "xenia/gpu/trace_protocol.h"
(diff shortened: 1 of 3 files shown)


src/xenia/gpu/trace_player.h
```diff
@@ -49,6 +56,10 @@ class TracePlayer : public TraceReader {
   GraphicsSystem* graphics_system_;
   int current_frame_index_;
   int current_command_index_;
+  std::atomic<bool> playing_trace_ = false;
```
benvanik: atomic here isn't doing what's intended - better to drop it so it's not confusing

src/xenia/gpu/trace_player.h
```diff
@@ -34,8 +35,14 @@ class TracePlayer : public TraceReader {
   GraphicsSystem* graphics_system() const { return graphics_system_; }
   int current_frame_index() const { return current_frame_index_; }
   int current_command_index() const { return current_command_index_; }
+  bool playing_trace() const { return playing_trace_; }
   const Frame* current_frame() const;
 
+  // Only valid if playing_trace is true.
+  const uint8_t* player_current_ptr() const { return player_current_ptr_; }
```
benvanik: you could expose this as a playback_percent() instead that pulls from a member that is updated during playback. that way less of the internal state is exposed in the API (which has threading implications) and the risk of the data being out of date is low (no one can use the percent as a pointer, so if it's slightly off that's not terrible).

DrChat: You're right - that's a better idea than what I'm doing

PR #471, opened 2015-11-25
Premake: Fix usage of build scripts

sephiroth99:
Please see commit message for details of the problem and fix.

Pull requets for the same problem will be created for the "libavutil" and "elemental-forms" projects.

### DIFF
--- a/src/xenia/apu/premake5.lua
+++ b/src/xenia/apu/premake5.lua
@@ -18,4 +18,3 @@ project("xenia-apu")
     project_root.."/third_party/libav/",
   })
   local_platform_files()
-
--- a/src/xenia/apu/xaudio2/premake5.lua
+++ b/src/xenia/apu/xaudio2/premake5.lua
@@ -1,6 +1,7 @@
 project_root = "../../../.."
 include(project_root.."/build_tools")
 
+group("src")
 project("xenia-apu-xaudio2")
   uuid("7a54a497-24d9-4c0e-a013-8507a04231f9")
   kind("StaticLib")
--- a/src/xenia/hid/xinput/premake5.lua
+++ b/src/xenia/hid/xinput/premake5.lua
@@ -13,6 +13,6 @@ project("xenia-hid-xinput")
   defines({
   })
   includedirs({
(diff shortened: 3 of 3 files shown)


benvanik:
Interesting!

I think the easier solution here would be to fix platform_files.lua so that it doesn't need to be dofile'd. There's nothing about what it does that should make it unable to be included. Is it the local function that's causing the problem?

benvanik:
An alternative would be to make build_tools a module: https://github.com/premake/premake-core/wiki/Introducing-Modules

sephiroth99:
OK I'll look into premake modules. I'd keep this PR open, but close it if you want to.

benvanik:
Merged!

Thanks for taking a look! You can keep this open and just force push over it if you want, or if it's easier open a new one when you're ready and I'll close this one.

sephiroth99:
Allright so I tested the module stuff and it worked, but while looking at that I found the root cause of the issue. When including elemental-forms' premake file, it included it's version of build_tools (I think it was even out of date), so I guess Premake got real confused.

The simplest fix I imagined was to have a local Premake file for EF that does not include EF's build tools. EF's testbed is now excluded from Xenia's build; it's still in the EF project.

I also found some small issues with the premake files (2 whitespace and 1 missing group for apu/xaudio2) that I fixed.

Finally I updated libav to the latest stuff.

benvanik:
Good find on the root cause! Since I run elemental-forms, the ideal fix is to rev it's build_tools version so it can continue to be built as normal. I want the testbed included, etc.

sephiroth99:
OK I'll test that, and check if it fixes it that way too.

sephiroth99:
Allright, updating build tools in elemental-forms fixed it. Removed the commit that created the new premake file. Hope it's for good now!

Do you want to update the submodules in elemental-forms' repo? Or I can send a PR for that on the other repo.

benvanik:
You can send a PR - may as well get some github internet points for the
awesome sleuthing you've done :)

sephiroth99:
Allright when xenia-project/elemental-forms#27 is merged, I'll update this PR with the updated elemental-forms submodule and this should do it.

PR #472, opened 2015-11-25
Replaced all checks for python with python2

ca1ek:
This makes it so, if both Python 2 and 3 are present on the system,
Python 2 will get used. Executing just 'python' will run either the more
recently installed version or the one with a higher version number.
'python2' guarantees that Python 2 will get used if present

Tested with ActivePython 2.7. Please check with normal Python 2.7, even though it should work with no problems on both.

### DIFF
--- a/xb.bat
+++ b/xb.bat
@@ -20,7 +20,7 @@ REM ============================================================================
 REM Trampoline into xenia-build
 REM ============================================================================
 
-python xenia-build %*
+python2 xenia-build %*
 EXIT /b %ERRORLEVEL%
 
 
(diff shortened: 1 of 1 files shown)


benvanik:
The python people are quite insufferable and refuse to add a python2 like ActivePython does: https://bugs.python.org/issue21506

So this script could instead check for python2, and if not found try just python.

ca1ek:
Alright, I'm not experienced at windows batch but will try to make something

ca1ek:
Will close this pull request for now, when I make this batch thing I will open up a new one.

Also, just to say, whoever made Batch was high on... something. Something I wouldn't want to be high on.

benvanik:
Hehe yep :P

ca1ek:
Shouldn't instead be used a Python script that self checks it's version?

benvanik:
It does use python - that's what the `python -c` is - but it first needs to make sure python exists. If you don't have python installed, you can't use python to test itself ;)

benvanik:
Implemented in 57f2504e1f0f905b780325eacbd268c1b02e87b3

ca1ek:
Well, if the file had .py extension most people would check what to run it with, then run it themselves without need for a check script. Anyways, thanks for implementing my idea.

benvanik:
You can do just that - just use 'python xenia-build.py foo' - the xb script
is just a helper so I don't have to do that.

PR #475, opened 2015-12-01
VS2015 Update 1 fix + GLContext small cleanup

sephiroth99:
Installed VS2015U1, fixed compilation. Similar fix to commit 08ae855.

Also, dc_ can be used directly right? Is the dc() function a relic of some older and deleted code? Anyway, I removed it.

Thanks!


sephiroth99:
Hmm looks like something broke Travis...

robhshu:
:+1: for f367c4b

Issue #459, opened 2015-11-13
Save States


DrChat:
* Serialization of the kernel state (and by extension, all XObjects)
* Threads
 * We can reconstruct the host stack by walking the guest stack at runtime.
   * For locals (spilled registers) we could step the thread until they're guaranteed to be in sync with the PPC context.
 * In HLE Functions
   * We can't save a thread that's in a HLE function. We'll have to run the thread until it returns from the HLE function and is back in guest code.
    * If the HLE function is blocking, we'll have to save the thread anyways and just have it call the HLE function again when it's restored.

* File Format
 * Would be neat to pack the state into a .zip file, with different subfiles for different areas (ex. memory file, kernel file, graphics file, etc.)
 * At the very least we should use zlib to compress the memory before writing it into a file.

benvanik:
Blocking host functions almost always end up in a xe::threading::Wait*. We can wrap that when calling from the kernel functions to also include a global 'saving state' event (and always call WaitMultiple/WaitAll/etc - with some tricky logic - AudioSystem does this now by adding the shutdown_event_ to wait_handles_). Then, when a save state is requested, that saving state event can be signaled to wake all waiting threads. They can then check if they are being asked to save, and back out any state they have and bubble up to the syscall. When restoring state, they'll be called back into just as normal.

This requires the wrapping/identification of the waits, probably an assert so that we don't accidentally introduce waits that have guest code on the stack, all waits handling the early-exit on saving to do whatever they need to do, and code cleanup before the waits to ensure that and side-effects happen either after the wait or can be reverted.

If we have that, we can make some assertions about other code that need to be made true:

* Everywhere we can we need to use guest allocations and guest pointers instead of host pointers. This makes the code safe to restoring as addresses will be stable (as PPC addresses don't change). As we won't be trying to restore mid-host function so long as each time a pointer is used it's translated it should always be ok.

* Any host threads that call into guest code (apu/gpu interrupts) must return to host code before we can save.

* Anywhere that can recreate state should be able to (gpu mostly does this now with a cache clear, but not sure the apu does).

* GPU may need to dictate save time - possibly on swap only - as there are waits embedded in the command buffers that will be extremely difficult to restore.

The big complication here is the potential for deadlocks/livelocks between threads in different suspend states.

hlide:
Or using APC with "alertable" wait functions (which include a bigger set of win32 waiting functions that the one cited above).

However, in both case, I wonder how you will be able to implement the Linux and OSx versions.

[...]

PR #476, opened 2015-12-01
Save States

DrChat:
!! DO NOT MERGE YET !!

Save states! Creating this PR now so I can get feedback on my work as well as centralize discussions on this topic. Related issue #459.

### DIFF
--- a/src/xenia/base/byte_stream.h
+++ b/src/xenia/base/byte_stream.h
@@ -32,6 +32,7 @@ class ByteStream {
   }
 
   const uint8_t* data() const { return data_; }
+  uint8_t* data() { return data_; }
   size_t data_length() const { return data_length_; }
 
   size_t offset() const { return offset_; }
--- a/src/xenia/cpu/backend/x64/x64_backend.cc
+++ b/src/xenia/cpu/backend/x64/x64_backend.cc
@@ -143,7 +143,7 @@ bool X64Backend::InstallBreakpoint(Breakpoint* bp) {
 
     auto orig_bytes =
         xe::load_and_swap<uint16_t>(reinterpret_cast<void*>(code + 0x0));
-    bp->backend_data().push_back({code, orig_bytes });
+    bp->backend_data().push_back({code, orig_bytes});
 
     xe::store_and_swap<uint16_t>(reinterpret_cast<void*>(code + 0x0), 0x0F0C);
   }
@@ -170,7 +170,7 @@ bool X64Backend::InstallBreakpoint(Breakpoint* bp, Function* func) {
 }
 
 bool X64Backend::UninstallBreakpoint(Breakpoint* bp) {
-  for (auto pair : bp->backend_data()) {
+  for (auto& pair : bp->backend_data()) {
     xe::store_and_swap<uint16_t>(reinterpret_cast<void*>(pair.first),
                                  uint16_t(pair.second));
   }
--- a/src/xenia/cpu/processor.cc
+++ b/src/xenia/cpu/processor.cc
@@ -422,6 +422,15 @@ bool Processor::InstallBreakpoint(Breakpoint* bp) {
     return false;
   }
 
+  auto functions = FindFunctionsWithAddress(bp->address());
+  if (functions.empty()) {
+    // Go ahead and generate a function.
+    if (!ResolveFunction(bp->address())) {
+      // Could not generate a function - fail!
+      return false;
+    }
+  }
+
   // We need to register the breakpoint before installing it with the backend
   // in-case a thread hits it while we're here.
   breakpoints_.push_back(bp);
--- a/src/xenia/cpu/processor.cc
+++ b/src/xenia/cpu/processor.cc
@@ -408,9 +408,9 @@ void Processor::LowerIrql(Irql old_value) {
 void Processor::BreakpointFunctionDefined(Function* function) {
   std::lock_guard<std::recursive_mutex> lock(breakpoint_lock_);
 
-  for (auto it = breakpoints_.begin(); it != breakpoints_.end(); it++) {
-    if (function->ContainsAddress((*it)->address())) {
-      backend_->InstallBreakpoint(*it, function);
+  for (auto bp : breakpoints_) {
+    if (function->ContainsAddress(bp->address())) {
+      backend_->InstallBreakpoint(bp, function);
     }
   }
 }
@@ -465,9 +465,9 @@ bool Processor::BreakpointHit(uint32_t address, uint64_t host_pc) {
 Breakpoint* Processor::FindBreakpoint(uint32_t address) {
   std::lock_guard<std::recursive_mutex> lock(breakpoint_lock_);
 
-  for (auto it = breakpoints_.begin(); it != breakpoints_.end(); it++) {
-    if ((*it)->address() == address) {
-      return *it;
+  for (auto bp : breakpoints_) {
+    if (bp->address() == address) {
+      return bp;
     }
   }
 
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -344,6 +344,8 @@ bool Emulator::SaveToFile(const std::wstring& path) {
   ByteStream stream(map->data(), map->size());
   stream.Write('XSAV');
 
+  // It's important we don't hold the global lock here! XThreads need to step
+  // forward (possibly through guarded regions) without worry!
   kernel_state_->Save(&stream);
   memory_->Save(&stream);
   map->Close(stream.offset());
@@ -353,6 +355,8 @@ bool Emulator::SaveToFile(const std::wstring& path) {
 }
 
 bool Emulator::RestoreFromFile(const std::wstring& path) {
+  auto lock = global_critical_region::AcquireDirect();
+
   // Restore the emulator state from a file
   auto map = MappedMemory::Open(path, MappedMemory::Mode::kReadWrite);
   if (!map) {
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -355,8 +355,6 @@ bool Emulator::SaveToFile(const std::wstring& path) {
 }
 
 bool Emulator::RestoreFromFile(const std::wstring& path) {
-  auto lock = global_critical_region::AcquireDirect();
-
   // Restore the emulator state from a file
   auto map = MappedMemory::Open(path, MappedMemory::Mode::kReadWrite);
   if (!map) {
@@ -369,6 +367,7 @@ bool Emulator::RestoreFromFile(const std::wstring& path) {
   Pause();
   kernel_state_->TerminateTitle();
 
+  auto lock = global_critical_region::AcquireDirect();
   ByteStream stream(map->data(), map->size());
   if (stream.Read<uint32_t>() != 'XSAV') {
     return false;
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -289,12 +289,15 @@ X_STATUS Emulator::LaunchStfsContainer(std::wstring path) {
 }
 
 void Emulator::Pause() {
-  auto lock = global_critical_region::AcquireDirect();
   if (paused_) {
     return;
   }
   paused_ = true;
 
+  // Don't hold the lock on this (so any waits follow through)
+  graphics_system_->Pause();
+
+  auto lock = global_critical_region::AcquireDirect();
   auto threads =
       kernel_state()->object_table()->GetObjectsByType<kernel::XThread>(
           kernel::XObject::kTypeThread);
@@ -316,6 +319,8 @@ void Emulator::Resume() {
   }
   paused_ = false;
   XELOGD("! EMULATOR RESUMED !");
+
+  graphics_system_->Resume();
 
   auto threads =
       kernel_state()->object_table()->GetObjectsByType<kernel::XThread>(
@@ -346,6 +351,7 @@ bool Emulator::SaveToFile(const std::wstring& path) {
 
   // It's important we don't hold the global lock here! XThreads need to step
   // forward (possibly through guarded regions) without worry!
+  graphics_system_->Save(&stream);
   kernel_state_->Save(&stream);
   memory_->Save(&stream);
   map->Close(stream.offset());
@@ -373,6 +379,9 @@ bool Emulator::RestoreFromFile(const std::wstring& path) {
     return false;
   }
 
+  if (!graphics_system_->Restore(&stream)) {
+    return false;
+  }
   if (!kernel_state_->Restore(&stream)) {
     return false;
   }
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -296,6 +296,7 @@ void Emulator::Pause() {
 
   // Don't hold the lock on this (so any waits follow through)
   graphics_system_->Pause();
+  audio_system_->Pause();
 
   auto lock = global_critical_region::AcquireDirect();
   auto threads =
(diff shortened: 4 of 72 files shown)


benvanik:
Ooo cool! Will start looking at this asap!

src/xenia/base/byte_stream.h
```diff
@@ -0,0 +1,91 @@
...
+
+#ifndef XENIA_BASE_BYTE_STREAM_H_
+#define XENIA_BASE_BYTE_STREAM_H_
+
+#include <cstdint>
+#include <string>
+
+namespace xe {
+
+class ByteStream {
```
benvanik: nice type - we may be able to replace some of the other types with this (StringReader, Arena, etc)

src/xenia/base/threading_win.cc
```diff
@@ -453,6 +453,15 @@ std::unique_ptr<Thread> Thread::Create(CreationParameters params,
   return std::make_unique<Win32Thread>(handle);
 }
 
+std::unique_ptr<Thread> Thread::GetCurrentThread() {
+  HANDLE handle = ::GetCurrentThread();
+  if (handle == INVALID_HANDLE_VALUE) {
+    return nullptr;
+  }
+
+  return std::make_unique<Win32Thread>(handle);
```
benvanik: it'd be nice to use a `thread_local` class static for this instead - that way there's a 1:1 mapping between Win32Thread and the underlying windows thread (and if GetCurrentThread() is called frequently we aren't thrashing memory. You could still have it allocate on demand (when called from a thread that hadn't been created from this file).

src/xenia/cpu/backend/x64/x64_backend.cc
```diff
@@ -116,6 +124,103 @@ std::unique_ptr<GuestFunction> X64Backend::CreateGuestFunction(
...
+  auto orig_bytes =
+      xe::load_and_swap<uint16_t>(reinterpret_cast<void*>(code + 0x0));
+  bp->backend_data().push_back({code, orig_bytes});
+
+  xe::store_and_swap<uint16_t>(reinterpret_cast<void*>(code + 0x0), 0x0F0C);
+  return true;
+}
+
+bool X64Backend::UninstallBreakpoint(Breakpoint* bp) {
+  for (auto pair : bp->backend_data()) {
```
benvanik: (unsure) in general you may want to do `auto&` so that the pair is not copied - I always forget if the compiler implicitly handles that or not, but I do know it'll often copy things

src/xenia/cpu/function.h
```diff
@@ -21,6 +21,10 @@
 
 namespace xe {
 namespace cpu {
+class Export;
```
benvanik: can move this down - no need for duplicate namespace

src/xenia/cpu/processor.cc
```diff
@@ -383,5 +405,74 @@ void Processor::LowerIrql(Irql old_value) {
...
+}
+
+bool Processor::UninstallBreakpoint(Breakpoint* bp) {
+  std::lock_guard<std::recursive_mutex> lock(breakpoint_lock_);
+
+  if (!backend_->UninstallBreakpoint(bp)) {
+    return false;
+  }
+
+  for (auto it = breakpoints_.begin(); it != breakpoints_.end(); it++) {
```
benvanik: fancy c++11 protip: this syntax can be used here (and below):
```
for (auto it : breakpoints_) { ... }
```

src/xenia/cpu/processor.h
```diff
@@ -46,10 +49,11 @@ enum class Irql : uint32_t {
 
 class Processor {
  public:
-  Processor(Memory* memory, ExportResolver* export_resolver,
+  Processor(Emulator* emulator, Memory* memory, ExportResolver* export_resolver,
```
benvanik: what's emulator used for here? It'd be nice not to have this dependency (as then we don't have to think about when emulator is/isn't present, etc)

src/xenia/emulator.cc
```diff
@@ -285,6 +288,112 @@ X_STATUS Emulator::LaunchStfsContainer(std::wstring path) {
...
+    return;
+  }
+  paused_ = true;
+
+  auto threads =
+      kernel_state()->object_table()->GetObjectsByType<kernel::XThread>(
+          kernel::XObject::kTypeThread);
+  for (auto thread : threads) {
+    if (!thread->can_debugger_suspend()) {
+      // Don't pause host threads.
```
benvanik: note that there are a few threads that are considered host threads that we may want to pause - like the vblank one, audio, etc - as they can make callbacks/interrupts into guest code. Maybe we need to tag the threads with a 'kSuspendOnPause' or something instead of using whether it's host or guest.

src/xenia/kernel/kernel_state.cc
```diff
@@ -575,5 +592,60 @@ void KernelState::CompleteOverlappedDeferredEx(
...
+  stream->Write('KRNL');
+
+  // Save the object table
+  object_table_.Save(stream);
+
+  // Save all objects
+  auto objects = object_table_.GetAllObjects();
+  uint32_t* num_objects_ptr = (uint32_t*)(stream->data() + stream->offset());
+  size_t num_objects = objects.size();
+  stream->Write((uint32_t)num_objects);
```
benvanik: nit: prefer static_cast and reinterpret_cast and such instead of C-style casts. At some point I'd like to turn on cpplint to warn on this stuff, so just a heads up :)

src/xenia/kernel/kernel_state.cc
```diff
@@ -329,21 +336,21 @@ object_ref<UserModule> KernelState::LoadUserModule(const char* raw_name) {
 void KernelState::TerminateTitle(bool from_guest_thread) {
   auto global_lock = global_critical_region_.Acquire();
```
benvanik: need to call into the xapps here to have them kill/reset their state (like playlists in the xmp app, etc)

src/xenia/kernel/user_module.cc
```diff
@@ -101,7 +102,8 @@ X_STATUS UserModule::LoadFromMemory(const void* addr, const size_t length) {
     }
 
     // Copy the xex2 header into guest memory.
-    const xex2_header* header = this->xex_module()->xex_header();
+    auto header = this->xex_module()->xex_header();
+    auto security_header = this->xex_module()->xex_security_info();
```
benvanik: no need for this here - can just use the local xex_module variable (here and below)

src/xenia/kernel/util/object_table.h
```diff
@@ -65,6 +76,18 @@ class ObjectTable {
...
+  std::vector<object_ref<T>> GetObjectsByType() {
+    std::vector<object_ref<T>> results;
+    GetObjectsByType(
+        T::kType,
+        reinterpret_cast<std::vector<object_ref<XObject>>*>(&results));
+    return results;
+  }
+
+  std::vector<object_ref<XObject>> GetAllObjects();
+  void PurgeAllObjects(); // Purges the object table of all guest objects
```
benvanik: linter will probably yell about the comment being like this

src/xenia/kernel/xevent.cc
```diff
@@ -70,5 +71,33 @@ int32_t XEvent::Reset() {
 
 void XEvent::Clear() { event_->Reset(); }
 
+bool XEvent::Save(ByteStream* stream) {
+  XELOGD("XEvent %.8X (%s)", handle(), manual_reset_ ? "manual" : "auto");
+  SaveObject(stream);
+
+  // TODO: Write if we're currently signaled.
```
benvanik: you can wait for the event with a zero timeout and check the return code - if it is `WAIT_TIMEOUT` you know it wasn't signaled.

src/xenia/kernel/xobject.cc
```diff
@@ -84,8 +86,68 @@ X_STATUS XObject::Delete() {
...
+    case kTypeSession:
+      break;
+    case kTypeSocket:
+      break;
+    case kTypeThread:
+      return XThread::Restore(kernel_state, stream);
+    case kTypeTimer:
+      break;
+    default:
+      assert_always("No restore handler exists for this object!");
```
benvanik: should probably have all the unimplemented ones fall through to this

src/xenia/kernel/xobject.h
```diff
@@ -128,10 +133,22 @@ class XObject {
...
+  // Returns all associated handles with this object.
+  std::vector<X_HANDLE> handles() const { return handles_; }
+  std::vector<X_HANDLE>& handles() { return handles_; }
+
   const std::string& name() const { return name_; }
   uint32_t guest_object() const { return guest_object_ptr_; }
 
+  // Has this object been created for use by the host?
+  // Host objects are persisted through reloads/etc.
+  bool host_object() const { return host_object_; }
```
benvanik: style tip: bools like this named as is_* (so `is_host_object`) - makes it easy to know what it is (instead of if you just saw `host_object()`, you may think it's actually a pointer to some host object or something)

src/xenia/cpu/processor.cc
paulsapps: don't forget auto& as you noted above ;)

src/xenia/kernel/xthread.cc
```diff
@@ -124,11 +137,15 @@ void XThread::set_last_error(uint32_t error_code) {
 }
 
 void XThread::set_name(const std::string& name) {
-  name_ = name;
+  StringBuffer buff;
```
benvanik: StringBuffer is expensive - it allocs like 16KiB by default -- there's `xe:: format_string("foo%s", ...)` that'll return an std::string you can use without to many extra copies/allocations.

src/xenia/kernel/xthread.cc
```diff
@@ -709,6 +761,376 @@ X_STATUS XThread::Delay(uint32_t processor_mode, uint32_t alertable,
...
+
+uint32_t XThread::StepIntoBranch(uint32_t pc) {
+  cpu::frontend::InstrData i;
+  i.address = pc;
+  i.code = xe::load_and_swap<uint32_t>(memory()->TranslateVirtual(i.address));
+  i.type = cpu::frontend::GetInstrType(i.code);
+
+  auto context = thread_state_->context();
+
+  if (i.type->type & (1 << 4) /* BranchAlways */) {
```
benvanik: there are enums for this so you don't have to bitshift:
https://github.com/benvanik/xenia/blob/ee92f75a842b0b268808e7b502e4b3f982d136de/src/xenia/cpu/frontend/ppc_instr.h#L79

incidentally, what you have here doesn't match :P

src/xenia/memory.cc
```diff
@@ -516,6 +546,82 @@ void BaseHeap::DumpMap() {
...
+    auto& page = page_table_[i];
+    stream->Write(page.qword);
+    if (!page.state) {
+      // Unallocated.
+      continue;
+    }
+
+    // TODO: Write compressed
+    if (page.state & kMemoryAllocationCommit &&
+        page.current_protect & kMemoryProtectRead) {
```
benvanik: we may have to write pages even with noaccess - as a game could put valid data somewhere, Protect() it with no access, and then unprotect it later and expect it to be there.

benvanik:
In general: holy shit this is awesome. Super elegant implementation. Great work!!

src/xenia/cpu/processor.h
DrChat: I added it for my syscall hack. Not really necessary otherwise, but convenient.

benvanik: ah - good to drop now then - it took me awhile to remove the cycles having it introduced between the modules

src/xenia/kernel/xthread.cc
DrChat: Did that cause (for example) the BranchAlways enum matches BranchAlways and Branch - basically matching all branching instructions.

src/xenia/kernel/xevent.cc
DrChat: Question is does this screw with automatic resetting events?

src/xenia/kernel/util/object_table.h
DrChat: Linter's gonna yell at me for a lot of stuff :P

I'll do a commit when I'm done with these changes that lints everything properly.

src/xenia/kernel/user_module.cc
DrChat: Not here - it's a unique ptr that we've moved to the processor.

src/xenia/kernel/xthread.cc
benvanik: ?

`(1 << 4) /* BranchAlways */`

```
  kXEPPCInstrTypeGeneral = (1 << 0),
  kXEPPCInstrTypeBranch = (1 << 1),
  kXEPPCInstrTypeBranchCond = kXEPPCInstrTypeBranch | (1 << 2),
  kXEPPCInstrTypeBranchAlways = kXEPPCInstrTypeBranch | (1 << 3),
  kXEPPCInstrTypeSyscall = (1 << 4),
```

BranchAlways is 1<<3, not 1<<4 - so the comment is wrong?

src/xenia/kernel/xevent.cc
benvanik: you'll have to fix up the event after you query - if it comes back that it was signalled you'll need to re-signal it. because you only do this when nothing else should be waiting on that it should be ok

src/xenia/kernel/user_module.cc
benvanik: ah good catch - blah

src/xenia/kernel/xobject.h
DrChat: :+1: Will fix

src/xenia/kernel/xthread.cc
DrChat: Weird. My local copy has kXEPPCInstrTypeSynchronizeContext at 1<<1 - must not have properly branched out when I started this. I'll go ahead and rebase.

src/xenia/memory.cc
DrChat: True - should be safe to unprotect them, write, then protect them again.

src/xenia/emulator.cc
DrChat: Do we have any XThreads that don't do that? Should probably not be using an XThread for anything that doesn't call into the guest - instead register threads with the emulator.

src/xenia/cpu/processor.cc
hlide: it should be : `for (auto & bp : breakpoints_) {...}` where `bp = *it` ;p

Razzile: would it not matter if bp was a copy or reference if the value is just a pointer to a breakpoint? I might be mistaken though (probably am)

DrChat: Hah, whoops - guess I kind of got carried away after writing UninstallBreakpoint and just copied everything.

DrChat:
For some reason, behavior in most games has regressed - they're freezing (ironically) in the hang detection code for the GPU.

DrChat:
And first milestone reached: Restores work (without audio atm)

Razzile:
It's incredible to see how fast the development of Xenia is progressing. Good job 👍

DrChat:
Saving/restoring mutexes is gonna be a bit weird. There's no way to get the mutex owner, and it's not gonna be trivial to store that information because waits acquire mutex ownership.

Any cool ideas?

robhshu:
A couple of great fixes here independent of this feature. Keep it up!

paulsapps:
@DrChat is the mutex not have some abstraction layer/indirection that can be leveraged for that?

DrChat:
I'd have to constrain the size in bits to a multiple of 64, but other than that there's no reason not to do it.

benvanik:
Closing PR as it's been merged! Woohoo!
:heart:

Margen67:
#665

PR #477, opened 2015-12-02
Compiler fixes

DrChat:
* Fix double-swapping in OPCODE_LOAD
* Add code to enforce page protections (but commented because it introduces a load of constants)

### DIFF
--- a/src/xenia/cpu/compiler/passes/constant_propagation_pass.cc
+++ b/src/xenia/cpu/compiler/passes/constant_propagation_pass.cc
@@ -153,6 +153,14 @@ bool ConstantPropagationPass::Run(HIRBuilder* builder) {
             i->Remove();
           }
           break;
+        case OPCODE_CONVERT:
+          if (i->src1.value->IsConstant()) {
+            TypeName target_type = v->type;
+            v->set_from(i->src1.value);
+            v->Convert(target_type, ROUND_TO_NEAREST);
+            i->Remove();
+          }
+          break;
         case OPCODE_ZERO_EXTEND:
           if (i->src1.value->IsConstant()) {
             TypeName target_type = v->type;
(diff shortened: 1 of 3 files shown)


src/xenia/cpu/hir/value.cc
```diff
@@ -77,8 +77,47 @@ uint64_t Value::AsUint64() {
...
 
 void Value::Cast(TypeName target_type) {
-  // TODO(benvanik): big matrix.
-  assert_always();
+  switch (type) {
+    case INT32_TYPE:
+      switch (target_type) {
+        case FLOAT32_TYPE:
+          // ???
+          constant.f32 = constant.f32;
```
benvanik: No question marks needed - or the assignment - as a constant op this is fine just being the type change

src/xenia/cpu/xex_module.cc
```diff
@@ -281,6 +281,34 @@ bool XexModule::Load(const std::string& name, const std::string& path,
     }
   }
 
+  // Setup memory protection.
+  // TODO: This introduces a load of constants into the JIT, and Xenia isn't
```
benvanik: What other things are missing? this is awesome!

src/xenia/cpu/compiler/passes/constant_propagation_pass.cc
```diff
@@ -153,6 +153,14 @@ bool ConstantPropagationPass::Run(HIRBuilder* builder) {
             i->Remove();
           }
           break;
+        case OPCODE_CONVERT:
+          if (i->src1.value->IsConstant()) {
+            TypeName target_type = v->type;
+            v->set_from(i->src1.value);
+            v->Convert(target_type, ROUND_TO_NEAREST);
```
benvanik: Pass the rounding mode in i->flags

src/xenia/cpu/xex_module.cc
DrChat: Over half of x64_sequences.cc isn't setup to handle constants. We probably need a rewrite or something.

benvanik: Ah yeah - maybe I can help with some of that. I still want to rework that file to make it more readable/etc. Adding a proper LIR would be nice, too.

PR #481, opened 2015-12-07
Fix compilation error of translated GLSL shader.

sephiroth99:
Comparison of uvec2 and int is not possible, explicitly
use a zero-value uvec2.

Error from logs:
```
!> 00000004 Unable to link program: 
Shader Compile Log:
Fragment shader failed to compile with the following errors:
ERROR: 0:125: error(#162) Wrong operand types: no operation "!=" exists that takes a left-hand operand of type "const in highp 2-component vector of uvec2" and a right operand of type "const int" (or there is no acceptable conversion)
ERROR: error(#273) 1 compilation errors.  No code generated
```

Example line: [see line 125 of full shader here](https://gist.github.com/sephiroth99/53b8a9087e8be1bc8424) This example shader is from Forza 2.
```
if (state.texture_samplers[0] != 0) {
```

### DIFF
--- a/src/xenia/gpu/glsl_shader_translator.cc
+++ b/src/xenia/gpu/glsl_shader_translator.cc
@@ -502,7 +502,7 @@ void GlslShaderTranslator::ProcessTextureFetchInstruction(
     case FetchOpcode::kTextureFetch:
       switch (instr.dimension) {
         case TextureDimension::k1D:
-          EmitSourceDepth("if (state.texture_samplers[%d] != 0) {\n",
+          EmitSourceDepth("if (state.texture_samplers[%d] != uvec2(0)) {\n",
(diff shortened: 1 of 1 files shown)


benvanik:
Interesting - are you using an AMD card?

sephiroth99:
Indeed I am.

benvanik:
I'm really hoping SPIR and Vulkan will remove this kind of error - it's shocking the nvidia glsl compiler allowed this when it's clearly wrong!

hlide:
well Nvidia is just using a GLSL translator to their own Cg compiler which is more permissive - I sometimes needed to use an external GLSL validator tool offered by AMD to check whether the generated GLSL was compliant or not.

PR #492, opened 2015-12-22
GL Backend Changes/Fixes

DrChat:
* Shaders:
 * Fix incorrect compare against bool consts
 * Always output the info log if it isn't empty (for shader warnings)
* GL backend:
 * Setup to allow Xenia to detect a graphics TDR on contexts
 * TODO: Dump the command stream on TDR.
 * Minor: Blitter::Draw asserts if src_texture is 0.

### DIFF
--- a/src/xenia/gpu/gl4/gl4_shader.cc
+++ b/src/xenia/gpu/gl4/gl4_shader.cc
@@ -140,8 +140,11 @@ bool GL4Shader::CompileProgram() {
   GLint log_length = 0;
   glGetProgramiv(program_, GL_INFO_LOG_LENGTH, &log_length);
   std::string info_log;
-  info_log.resize(log_length - 1);
-  glGetProgramInfoLog(program_, log_length, &log_length, &info_log[0]);
+  if (log_length > 0) {
+    info_log.resize(log_length - 1);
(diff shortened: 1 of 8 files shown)


src/xenia/gpu/gl4/gl4_shader.cc
```diff
@@ -135,18 +135,21 @@ bool GL4Shader::CompileProgram() {
     return false;
   }
 
+  // Output info log.
+  // log_length includes the null character.
+  GLint log_length = 0;
+  glGetProgramiv(program_, GL_INFO_LOG_LENGTH, &log_length);
+  std::string info_log;
+  info_log.resize(log_length - 1);
```
sephiroth99: When log_length is 0, this fails with "string too long" error.

DrChat: You using AMD or Nvidia?

sephiroth99: AMD...

DrChat: Okay, Nvidia always returns 1 if the log is empty I guess.

PR #493, opened 2015-12-26
Nvidia Optimus support

parkerlamb:
This PR adds support for automatically selecting the discrete GPU on laptops with combined Intel/Nvidia graphics.

I'm not entirely sure if this is the best place to put the exported variable, I'm happy to move it elsewhere if need be.

### DIFF
--- a/src/xenia/app/xenia_main.cc
+++ b/src/xenia/app/xenia_main.cc
@@ -26,6 +26,16 @@
 // Available graphics systems:
 #include "xenia/gpu/gl4/gl4_graphics_system.h"
 
+// Nvidia Optimus/AMD PowerXpress support
+// http://developer.download.nvidia.com/devzone/devcenter/gamegraphics/files/OptimusRenderingPolicies.pdf
+// http://stackoverflow.com/questions/17458803/amd-equivalent-to-nvoptimusenablement
+#if XE_PLATFORM_WIN32
+extern "C" {
+	__declspec(dllexport) unsigned long NvOptimusEnablement = 0x00000001;
+	__declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
+}
+#endif // XE_PLATFORM_WIN32
+
 // Available input drivers:
 #include "xenia/hid/nop/nop_hid.h"
 #if XE_PLATFORM_WIN32


paulsapps:
Do you not need a 

```
extern "C"
  {
    __declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
  }
```

For AMD too?

hlide:
The only requirement is it should be an exported symbol declared in the executable (EXE) - for instance in the same file where the function main is defined - which is the case with this PR. I already tried it for another emulator and it works fine on my Laptop which has a Nvidia Optimus. I also discovered the AMD version which should work the same way as Nvidia version.

hlide:
by the way, does "__declspec(dllexport)" be accepted by clang and gcc? how about Linux version?

parkerlamb:
I updated the commit with the AMD symbol included. If someone with the proper hardware could check that it actually works that would be great.

As for Linux support, I couldn't find any docs/source code on what should be done, everything seems to be Windows-specific. Hybrid graphics support for both AMD and Nvidia seems to be pretty broken on Linux so there may not really be a solution for it. The __declspec(dllexport)s are inside a #if XE_PLATFORM_WIN32 block so there at least shouldn't be any compile errors on gcc/clang.

gibbed:
Yes and no, you can still compile via GCC/clang on Windows. So it'd be best to make it platform independent, export wise.

parkerlamb:
I did some research on gcc/clang - apparently both support __declspec(dllexport) automatically under Windows so the code should be fine as-is. See here http://clang.llvm.org/docs/UsersManual.html#id50 and here https://gcc.gnu.org/wiki/Visibility

gibbed:
Good to know!

src/xenia/app/xenia_main.cc
```diff
@@ -26,6 +26,16 @@
 // Available graphics systems:
 #include "xenia/gpu/gl4/gl4_graphics_system.h"
 
+// Nvidia Optimus/AMD PowerXpress support
+// http://developer.download.nvidia.com/devzone/devcenter/gamegraphics/files/OptimusRenderingPolicies.pdf
+// http://stackoverflow.com/questions/17458803/amd-equivalent-to-nvoptimusenablement
+#if XE_PLATFORM_WIN32
+extern "C" {
+	__declspec(dllexport) unsigned long NvOptimusEnablement = 0x00000001;
```
mirhl: Ehrm.. why unsigned long? I remember in PCSX2/pcsx2#767 it was DWORD

FUNExtreme: DWORD is a 32-bit unsigned integer, so unsigned long and DWORD are the same

benvanik:
Nice!
Merging - may shuffle it around somewhere so that other exes besides just the main one get it.
Thanks for doing it!

PR #504, opened 2015-12-30
Fix various clang compilation issues 

sephiroth99:
Ran the ol' clang again. These fixes are pretty simple so they should not cause any problem. Almost half of it is to handle all enum values in switch-cases. Compiled on Windows and Linux, quick test on Windows.

There is more to come, I'm stubbing the code along so clang can eventually compile all of Xenia.

### DIFF
--- a/src/xenia/base/profiling.cc
+++ b/src/xenia/base/profiling.cc
@@ -37,10 +37,8 @@
 #include "third_party/microprofile/microprofileui.h"
 #endif  // XE_OPTION_PROFILING
 
-#if XE_OPTION_PROFILING_UI
 #undef DrawText
 #include "xenia/ui/microprofile_drawer.h"
-#endif  // XE_OPTION_PROFILING_UI
 
 DEFINE_bool(show_profiler, false, "Show profiling UI by default.");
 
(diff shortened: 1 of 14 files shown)


src/xenia/base/profiling.cc
```diff
@@ -37,10 +37,8 @@
 #include "third_party/microprofile/microprofileui.h"
 #endif  // XE_OPTION_PROFILING
 
-#if XE_OPTION_PROFILING_UI
```
benvanik: This flag doesn't really seem to be useful anymore - since you're removing it here, can you remove it elsewhere? Removing #ifs makes me happy :)

src/xenia/debug/ui/debug_window.cc
```diff
@@ -959,6 +959,8 @@ void DebugWindow::DrawRegistersPane() {
     case RegisterGroup::kHostVector: {
       ImGui::BeginChild("##host_vector");
       for (int i = 0; i < 16; ++i) {
+        float f[4];
```
benvanik: ah, interesting. We aren't using xmm_registers as xmm registers, so they should be `vec128_t` instead. Then we wouldn't need to do this dance to get the values.

benvanik:
Awesome fixes! Just two little comments.
Thanks for doing this cleanup - it'll be nice to get things compiling with clang :)

benvanik:
Manually landing because these are good changes and I want them :)

sephiroth99:
Nice, thanks! Sorry I was not able to do the fixes, I'm away for the New Year.

benvanik:
No worries!

PR #507, opened 2015-12-31
Fix the StfsContainerEntry not being initialized in StfsContainerFile

marc0303:


benvanik:
Good catch! A bit confusing there's an entry_ on this as well as the parent File - will have to clean that up.

PR #508, opened 2016-01-05
Move style_guide.md and building.md to docs/

rlabrecque:
Seems like a more appropriate place for them. Will Squash if preferred.
I'll update building.md later as well.


benvanik:
Rebase please - no need for two commits for a doc change

Issue #506, opened 2015-12-31
Parse embedded xex resources to get game info.


benvanik:
Currently all we have is media/title ID. Would be nice when saving files/titling the main window/etc to have the actual name.

kmather73:
Well one possible quick and easy way to implement this would be to do a Title_ID lookup from http://abgx360.xecuter.com/games_by_titleid.txt      The only problem with this is that it doesn't include any XBLA games and it may not include every disc based game either, in particular I'm thinking that it will miss a lot of the J-NTSC games. 

I'll try and figure how xex work and maybe get something better in the next few days/weeks

SonofUgly:
If you're just talking about simple stuff like this, XBLA game info are easy enough to get from the file:
Name (Region 1): 0x00000412
Name (Region 2): 0x00000512
...
0x000000912

Description (Region 1): 0x00000D12
Description (Region 2): 0x00000E12
...
0x00001212

Game Name: 0x00001692
Game Icon: 0x0000171A

Or you could read from the ArcadeInfo.xml file in them.

jackchentwkh:
xorloser is a great coder/hacker. too bad he never release his source code. I asked him few times regarding to his reverse engineering finds of the backward compatibility on 360, but never answered.
here is a source code of older version of xextool which is not supposed to be xorloser's code. but as I remembered, xorloser did mentioned that his xextool was inspired by somebody's xextool code. maybe this is it. 

https://github.com/Intline9/i360gm/tree/master/dep/xextool_old

the tool is capable for retrieving all xex infos, including a decrypted PE which one can explore more resources including the game name/game icon, etc. hope it help.

btw, xorloser packed his xextool with upx with some tricks, so one needs to unpack it manually to get  a clear exe file for idapro to reverse. I managed to unpack the exe file, and the unpacked file works fine under winxp x86, but not working under x64 which I don't know why. maybe it's because I unpack and rebuild the import section under winxp x86. the unpacked exe can be reversed in idapro, and the hexray decompiler works with it. but seems xorloser uses many low level code tricks, it's not that easy for me to reverse it clearly even with the decompiler. if any one is interested in it, I will post the unpacked version of the xextool.

btw, free60's wiki got some error with the xex file format, especially regarding to the optional header entry data size. on the other hand the source code of the xextool should be correct.

x1nixmzeng:
Good for contributors?

@jackchentwkh do we need the source?

jackchentwkh:
[...]

PR #509, opened 2016-01-07
Rebrand the Xenia window with module metadata

x1nixmzeng:
Replaces window title and icon with the resources in the current module - part of #506 

Example:

limbo_example

### DIFF
--- a/src/xenia/ui/window.h
+++ b/src/xenia/ui/window.h
@@ -53,7 +53,7 @@ class Window {
     return true;
   }
 
-  virtual bool SetIconFromBuffer(void *buffer, size_t size) = 0;
+  virtual bool SetIconFromBuffer(void* buffer, size_t size) = 0;
 
   virtual bool is_fullscreen() const { return false; }
   virtual void ToggleFullscreen(bool fullscreen) {}
--- a/src/xenia/ui/window.h
+++ b/src/xenia/ui/window.h
@@ -53,7 +53,7 @@ class Window {
     return true;
   }
 
-  virtual bool set_icon_from_buffer(void *buffer, size_t size) = 0;
+  virtual bool SetIconFromBuffer(void *buffer, size_t size) = 0;
 
   virtual bool is_fullscreen() const { return false; }
   virtual void ToggleFullscreen(bool fullscreen) {}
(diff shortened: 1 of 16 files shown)


x1nixmzeng:
Also (as I don't have a voice on #xenia-dev for whatever reason :disappointed:):

* Do you have any suggestions on where to store off the metadata for other uses (save system, etc)
* Is it too early to allow setting the region? (and provide regional game names)

DrChat:
Ooh, this is really cool! Also, to talk on #xenia-dev, you just need to be authenticated by NickServ.

rlabrecque:
I don't like that the title no longer says xenia and doesn't contain the branch info anymore, otherwise I love it.

Razzile:
@rlabrecque maybe it would be a good idea for rolling builds to show build number but actual releases to revert to game metadata

x1nixmzeng:
I considered creating a flag which allows the title to be completely replaced.

For now, there's a silly bug where `s` isn't incremented if the XDBF isn't found. I also want to figure out some other headers, as there should be achivements somewhere.

src/xenia/ui/window.h
```diff
@@ -53,6 +53,8 @@ class Window {
     return true;
   }
 
+  virtual bool set_icon_from_buffer(void *buffer, size_t size) = 0;
```
DrChat: This should be named SetIconFromBuffer. Only use lowercase underscored names for functions that directly return a variable.

x1nixmzeng:
Latest:

xenia-2

Off topic, I don't understand the linting errors. I'm staging and running `.\xb.bat format` before committing changes, but then travis rearranges the pointers? Do we not share the same formatting rules?

DrChat:
About that - @benvanik, why not automatically lint pull requests on merge?

Also @x1nixmzeng, there's an automatic linter for VS that I use. It's documented in the style guide.

DrChat:
:+1:

benvanik:
Cool stuff!

Will land and clean up some style stuff.

The linter on travis is a weird version - no longer matches head, and I haven't wanted to have our build process build clang from source :/ hoping to stabilize on 3.8 (which in theory is released in feb)

PR #510, opened 2016-01-09
Fixed rare crash when accessing library import by name

x1nixmzeng:
Found with Astropop (xenia-project/game-compatibility/issues/2).

The library name indexes are `0xCD00` and `0xCD01` which were masked to 8-bits in the previous implementation (https://github.com/benvanik/xenia/blob/ea99ba8e3bbac03386c95efe1496d5f4b03f5456/src/xenia/kernel/util/xex2.cc#L272).

I opted to just cap the index by the number of libraries. If there is a hard limit of 255 imports then the `xe_xex2_import_library_t` will need changing :smile:

### DIFF
--- a/src/xenia/cpu/xex_module.cc
+++ b/src/xenia/cpu/xex_module.cc
@@ -242,13 +242,15 @@ bool XexModule::Load(const std::string& name, const std::string& path,
   // FIXME: Don't know if 32 is the actual limit, but haven't seen more than 2.
   const char* string_table[32];
   std::memset(string_table, 0, sizeof(string_table));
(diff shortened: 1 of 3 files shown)


DrChat:
Have you verified that the name index is supposed to wrap around like this?

x1nixmzeng:
@DrChat It should never need to wrap. This particular XEX header must have been setup using an 8-bit index, and the high bits were uninitialized, causing the bug.

DrChat:
Ah - instead of using modulo, use a bitwise-and with 0xFF. If the value is greater than the string table count, assert.

PR #514, opened 2016-01-13
X64 Speedups

DrChat:
* Removed quite a few commonly hit emulated instructions
* PACK_TYPE_FLOAT16_3 (speculative, not confirmed by testing on 360 but seems to work ok-ish)

### DIFF
--- a/src/xenia/cpu/compiler/passes/constant_propagation_pass.cc
+++ b/src/xenia/cpu/compiler/passes/constant_propagation_pass.cc
@@ -161,6 +161,13 @@ bool ConstantPropagationPass::Run(HIRBuilder* builder) {
             i->Remove();
           }
           break;
+        case OPCODE_ROUND:
+          if (i->src1.value->IsConstant()) {
+            v->set_from(i->src1.value);
+            v->Round(RoundMode(i->flags));
+            i->Remove();
+          }
+          break;
         case OPCODE_ZERO_EXTEND:
           if (i->src1.value->IsConstant()) {
             TypeName target_type = v->type;
@@ -188,6 +195,7 @@ bool ConstantPropagationPass::Run(HIRBuilder* builder) {
 
         case OPCODE_LOAD:
           if (i->src1.value->IsConstant()) {
+            assert_false(i->flags & LOAD_STORE_BYTE_SWAP);
             auto memory = processor_->memory();
             auto address = i->src1.value->constant.i32;
             auto mmio_range =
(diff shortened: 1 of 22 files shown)


src/xenia/cpu/ppc/ppc_emit_memory.cc
```diff
@@ -732,8 +735,9 @@ int InstrEmit_stwcx(PPCHIRBuilder& f, const InstrData& i) {
 
   Value* ea = CalculateEA_0(f, i.X.RA, i.X.RB);
   Value* rt = f.ByteSwap(f.Truncate(f.LoadGPR(i.X.RT), INT32_TYPE));
-  f.Store(ea, rt);
-  f.StoreContext(offsetof(PPCContext, cr0.cr0_eq), f.LoadConstantInt8(1));
+  Value* res = f.ByteSwap(f.Truncate(f.LoadReserved(), INT32_TYPE));
+  Value* v = f.AtomicCompareExchange(ea, res, rt);
```
benvanik: these changes scare me a bit - are you sure they doing the right thing? if we aren't 100% sure it's correct it'd be better to leave it as it was until we know

src/xenia/cpu/compiler/passes/constant_propagation_pass.cc
```diff
@@ -253,12 +261,23 @@ bool ConstantPropagationPass::Run(HIRBuilder* builder) {
...
+              } else if (i->src2.value->IsConstant() &&
+                         i->src3.value->IsConstant()) {
+                // TODO
+                // v->set_from(i->src2.value);
+                // v->Select(i->src3.value, i->src1.value);
+                // i->Remove();
+              }
             } else {
-              v->set_from(i->src3.value);
+              // TODO
```
benvanik: when you add TODO's you should put a description of what it is you need to do - like, 'TODO(DrChat): handle constant vec128 in src1' or something - that way, people scanning can find them and have a hope of implementing them easier

src/xenia/cpu/ppc/ppc_emit_memory.cc
DrChat: Yeah - they're working as intended - though these suffer from the ABA problem it doesn't matter cause they're still under the global lock.

I've found that, even with the ABA problem, games do not appear to become any more unstable than they already are. Probably a side effect of being developed on PC.

I intended to allow users to disable the global lock in these blocks because that seems to cause a few hang-ups.

benvanik: coool - would be good to add a note to that effect, for others finding this code later - it looks like it should be doing the right thing, but because it's not (still under the lock) it could be confusing

PR #520, opened 2016-01-18
Fix shaders reading 10_11_11 bitfields backwards.

DrChat:
* Also rewrote GL4 Shaders and stuff

### DIFF
--- a/src/xenia/gpu/gl4/gl4_shader.cc
+++ b/src/xenia/gpu/gl4/gl4_shader.cc
@@ -18,9 +18,7 @@ namespace gl4 {
 
 GL4Shader::GL4Shader(ShaderType shader_type, uint64_t data_hash,
                      const uint32_t* dword_ptr, uint32_t dword_count)
-    : Shader(shader_type, data_hash, dword_ptr, dword_count),
-      program_(0),
-      vao_(0) {}
+    : Shader(shader_type, data_hash, dword_ptr, dword_count) {}
 
 GL4Shader::~GL4Shader() {
   glDeleteProgram(program_);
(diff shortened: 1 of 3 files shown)


src/xenia/gpu/glsl_shader_translator.cc
```diff
@@ -189,6 +191,43 @@ out gl_PerVertex {
 };
 layout(location = 0) flat out uint draw_id;
 layout(location = 1) out VertexData vtx;
+
+vec3 get_10_11_11_u(const uint data_in) {
+  vec3 vec;
+  vec.x = bitfieldExtract(data_in, 0, 10);
+  vec.y = bitfieldExtract(data_in, 10, 11);
+  vec.z = bitfieldExtract(data_in, 21, 11);
+
```
benvanik: kill 'dem extra whitespaces

src/xenia/gpu/gl4/gl4_shader.cc
```diff
@@ -120,81 +127,135 @@ bool GL4Shader::PrepareVertexArrayObject() {
...
+  if (!shader_) {
+    return "GL4Shader::GetShaderInfoLog(): Program is NULL";
+  }
+
+  std::string log;
+  GLint log_length = 0;
+  glGetShaderiv(shader_, GL_INFO_LOG_LENGTH, &log_length);
+  if (log_length > 0) {
+    log.resize(log_length - 1);
+    glGetShaderInfoLog(shader_, log_length, &log_length, &log[0]);
```
benvanik: use log.data() instead of &[0] - as a pattern, data() is more explicit and won't assert in checked builds like [0] can

DrChat: log.data() is const unfortunately - we'd have to cast that away.

Razzile: Whenever I see code that gets logs from gl functions, they tend to store the log in a vector<char> instead of a std::string. i guess that's to avoid casting like DrChat said

raven02:
Dragon ball Z Burst limit bootup wihtout graphic now since this commit.

SakataGintokiYT:
<h3>Boot up without graphic only sound intro and menu (color screens )</h3>
*Naruto Shippuden: Ultimate Ninja Storm Generations
*Naruto Shippuden: Ultimate Ninja Storm  2
*Naruto Shippuden: Ultimate Ninja Storm Revolution
*Naruto Shippuden: Ultimate Ninja Storm 3
*Gal Gun

sawnikthederphog:
Will this be reverted or fixed?

benvanik:
@DrChat what did this fix? sounds like it broke a bunch

DrChat:
It did fix BF1943's weapon drawing overflowing a matrix, but maybe by coincidence.
I know now that I was using bitfieldExtract backwards (it starts from the LSB, I thought it started from the MSB)

jackchentwkh:
https://www.opengl.org/wiki/Vertex_Specification

in openGL, for vertex format 2_10_10_10_rev and 10_11_11_rev, the elements are in reverse order. the lowest bits are the first element. so 2_10_10_10_rev represents w_z_y_x or a_b_g_r, and 10_11_11_rev represents z_y_x. 

in direct 3D, with 2_10_10_10 vertex format, it represents w_x_y_z, per the description in the wiki link above. there is a special vertex size GL_BGRA for D3D compatibility to use this D3D vertex type directly. but I found no further instructions for D3D vertex format 10_11_11. 

I am not a expert in the graphic field. just a quick search result and hope it help.

SakataGintokiYT:
@DrChat check this :
https://github.com/xenia-project/game-compatibility/issues/132

PR #522, opened 2016-01-20
Linux compilation/link and compatibility fixes

sephiroth99:
Some small fixes for compilation under Clang. There are two bigger fixes:
1- Using 'hardware_concurrency': implements logical_processor_count for all platforms. On my i5 it reported all the cores, I don't have an i7 with hyper threading to test.
2- Reworking RtlTime: uses C library functions in order to be portable. Tested with Minecraft game, at startup it does some conversions and the results are the same.

Thanks!

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
@@ -25,8 +25,10 @@
 #include "xenia/kernel/xevent.h"
 #include "xenia/kernel/xthread.h"
 
-// For FileTimeToSystemTime and SystemTimeToFileTime:
+#if XE_PLATFORM_WIN32
 #include "xenia/base/platform_win.h"
+#define timegm _mkgmtime
+#endif
 
 namespace xe {
 namespace kernel {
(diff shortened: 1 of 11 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
```diff
@@ -25,8 +25,10 @@
 #include "xenia/kernel/xevent.h"
 #include "xenia/kernel/xthread.h"
 
-// For FileTimeToSystemTime and SystemTimeToFileTime:
+#if XE_PLATFORM_WIN32
```
benvanik: is this needed anymore with the new methods?

sephiroth99: If you mean the platform check, yes it is needed. Windows implements timegm but calls it _mkgmtime, so the #define is used for that.

janisozaur:
On HT-enabled i5 (2 cores / 4 threads) `std::thread::hardware_concurrency()` returns 4.

If you expect the number to be different on i7, I can check as well later on.

sephiroth99:
@janisozaur It should be fine, for an i7 (4c/8t) I would expect it to return 8. If it's not too much trouble for you to check, go ahead, but yeah it should be fine.

janisozaur:
4c/8t i7 returns 8, as expected.

PR #524, opened 2016-01-23
Fix Depth Clear / few shader fixes

DrChat:
* Depth clear wasn't working cause we weren't getting the depth target if it wasn't the source.
* Shaders:
 * Fix 2_10_10_10 fetch being completely wrong (now only partially wrong)
 * Runtime check for vertex fetch index
 * Fix trying to make a vec1


benvanik:
I need to get a bot running on pcmaker's traces - any time the depth buffer stuff has changed things have both broken and gotten better - I'm hoping this fixes them all :P

PR #532, opened 2016-02-08
Decouple MicroProfilerFlip from Profiler::Present

DrChat:
* Makes moving the mouse around on-screen not start a new profiler frame


DrChat:
Appveyor build failed because of a timeout, not because of any compiler errors.

Issue #416, opened 2015-09-03
gpu-gl4: glDrawArraysInstancedBaseInstance generates GL_INVALID_OPERATION on AMD GPUs


sephiroth99:
**Problem**
When OpenGL debug logging is enabled, the following error is logged in pretty much any game:

`!> 1390 GL4 OpenGL: error(high) 1000: glDrawArraysInstancedBaseInstance has generated an error (GL_INVALID_OPERATION)`

**Possible cause**
It seems that the error occurs when the bound program pipeline has a Geometry Shader associated. For example, by removing the GS stage from the pipelines in [CommandProcessor::UpdateShaders](https://github.com/benvanik/xenia/blob/master/src/xenia/gpu/gl4/command_processor.cc#L1701), the error is not generated anymore (but, of course, what is displayed on screen is not correct/is garbage).

**Further analysis**
_Error conditions_
According to the [OpenGL wiki/manual](https://www.opengl.org/wiki/GLAPI/glDrawArraysInstancedBaseInstance), there are two cases when it can occur:
1. GL_INVALID_OPERATION is generated if a geometry shader is active and mode​ is incompatible with the input primitive type of the geometry shader in the currently installed program object.
_I verified what mode is configured when glDraw... is called and it is correct, so that is not it._
2. GL_INVALID_OPERATION is generated if a non-zero buffer object name is bound to an enabled array and the buffer object's data store is currently mapped.
_I don't know enough OpenGL to say if that happens or not._
3. I'm sure there are more undocumented cases...

_OpenGL debugger_
I also ran xenia under AMD CodeXL, and it detected the error before calling the function, but it did not give any more specific info on why it could happen

For reference, here is what it says:

```
OpenGL Error: glDrawArrays - GL_INVALID_OPERATION
Break
Reason
OpenGL Error
Break on
glDrawArrays(GL_TRIANGLES , 0 , 3)
Error Code
GL_INVALID_OPERATION
Error Description
The specified operation is not allowed in the current state. The offending function is ignored, having no side effect other than to set the error flag.
* Stopped before function execution
```

If it helps, I can also provide a function call history until the error (about 66 calls).

_Equivalent function_
As per the wiki, glDrawArraysInstancedBaseInstance behaves the same as glDrawArrays​. I replaced [the call](https://github.com/benvanik/xenia/blob/master/src/xenia/gpu/gl4/draw_batcher.cc#L321) with the following:

```
        assert(cmd->instance_count == 1);
[...]

PR #533, opened 2016-02-09
More shader changes

sephiroth99:
2 things:
1- minor syntax fix
2- changes from commit f75a4cab0e2ec6d7acde47bea38fbd994994ac54 introduced usage of ```mix(vec4, vec4, bool)``` which is a GLSL 4.50 function. The shaders do not compile on AMD anymore, so make it a failure to not support GLSL 4.50 (like it was before). Also, this probably means that issue #416 is not caused by incomplete GLSL support...


DrChat:
:| Sorry AMD users, didn't realize that didn't compile.

May not really matter anyways. I hear Vulkan's going to be out either this month or next month, and we're going to be all over that when it happens.

PR #538, opened 2016-02-16
Trace Viewer Zoomed View / Fix Alpha Test

DrChat:
* Zoomed views for color targets / depth targets
* Fixes invisible images due to their alpha.
* Also correctly flags the alpha test as disabled/enabled

### DIFF
--- a/src/xenia/ui/gl/gl_immediate_drawer.cc
+++ b/src/xenia/ui/gl/gl_immediate_drawer.cc
@@ -270,6 +270,16 @@ void GLImmediateDrawer::End() {
   }
 }
 
+void GLImmediateDrawer::EnableAlphaTest(bool enable) {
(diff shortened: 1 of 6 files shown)


src/xenia/ui/gl/gl_immediate_drawer.cc
```diff
@@ -270,6 +270,16 @@ void GLImmediateDrawer::End() {
   }
 }
 
+void GLImmediateDrawer::EnableAlphaTest(bool enable) {
```
benvanik: side-channeling this is not good - make it part of the API if it's required. add `bool alpha_blend` to ImmediateDraw, have imgui_drawer populate it, and then have the gl drawer ::Draw switch its mode based on that (being smart about only switching if needed)

PR #542, opened 2016-02-19
Fix Validator Error (with swapchains)

DrChat:


src/xenia/ui/vulkan/vulkan_swap_chain.cc
```diff
@@ -37,9 +37,13 @@ VulkanSwapChain::~VulkanSwapChain() { Shutdown(); }
 bool VulkanSwapChain::Initialize(VkSurfaceKHR surface) {
   surface_ = surface;
 
+  VkBool32 surface_supported = false;
+  auto err = vkGetPhysicalDeviceSurfaceSupportKHR(
+      *device_, device_->queue_family_index(), surface, &surface_supported);
+
```
benvanik: CheckResult that err and assert_true surface_supported - if we get here it should be true and succeed (unless my assumptions are wrong :)

PR #553, opened 2016-02-22
More SPIR-V Opcodes

DrChat:


src/xenia/gpu/spirv_shader_translator.cc
```diff
@@ -785,6 +1088,152 @@ void SpirvShaderTranslator::ProcessScalarAluInstruction(
...
+
+      b.setBuildPoint(kill_block);
+      b.createNoResultOp(spv::Op::OpKill);
+
+      b.setBuildPoint(continue_block);
+      dest = b.makeFloatConstant(0.f);
+    } break;
+
+    case AluScalarOpcode::kLogc: {
+    } break;
```
benvanik: worth asserting in here/the others so we remember? or you gonna do these next?

DrChat: I figure it sticks out enough now - will add asserts in later

PR #562, opened 2016-03-10
Fix usage of mix in fragment shader

sephiroth99:
The boolean value must be a vector too, where each bool component
selects whether the value of the output vector component is from the
first or the second source vector.


sephiroth99:
Also of note, AMD now (finally) supports GLSL 4.50! woo!

nulcearc:
Wow! Now it works with AMD gpu!

PR #589, opened 2016-06-22
Dump shaders in binary mode

sephiroth99:
Use binary mode when writing shaders to file, so that 0x0a is not
considered a LF and then converted to CRLF.


DrChat:
:+1: Nice catch!

PR #598, opened 2016-07-07
Fix dead links to Google style guide

jakehamilton:
A few of the links pointing to Google's cpp style guide were pointing to dead pages where the old Google code project was. The project has since switched to Github and hosts itself as html pages. This commit updates the references to point to the new urls.


DrChat:
:+1: Awesome - thanks!

Issue #529, opened 2016-02-04
XNA games based on .NET not supported


SakataGintokiYT:
@x1nixmzeng 
@benvanik 
@gibbed 

Emulator closes after starting the game
*You need to add support for this format
-Arkedo Series: 01 JUMP!
-Arkedo Series: 02 SWAP!
-Arkedo Series: 03 PIXEL!

jackchentwkh:
As I know the XBLIG indie games are created with XNA studio which is based on .NET framework. The execution file is actually a managed file including IL, not machine code. So unless the Xenia can emulate the whole system function of the xbox360 including the IL interpretation, we won't expect the emulation of these XNA titles any sooner. on the other hand, there is one 360 emulator dedicated for this IL type XBLA related titles, but unfortunately the emulator was abandoned.

jackchentwkh:
there are two ways to run the titles built with XNA, one is to decompile it back to source code with ILspy or dotpeek, the recompile the source code with the publicly available XNA sdk on PC and fix the related errors in the source. the other way is a high end emulation which converts the xex files back to windows PE file format first, then introduces the windows version XNA runtimes, and then uses self-developed DLLs to fixup the missing function between 360 and PC. here is a unfinished HLE project. 

https://code.google.com/archive/p/ex360e/

x1nixmzeng:
This is perhaps another case for trying to read `GameInfo.xml` or `GameInfo.bin` before launching `default.xex`. But as @jackchentwkh mentions there are bigger differences with XNA games.

x1nixmzeng:
I was thinking of `ArcadeInfo.xml` in my last post.

The launch target in this series is `Arkedo Series - 02 SWAP!.exe` - which is a valid Windows executable... It's worth taking a look at the work done for Fez. I think there was an attempt to convert it to PC somewhere!

x1nixmzeng:
So we can actually decompile this to IL using [ildasm](https://msdn.microsoft.com/en-us/library/f7dy01k1%28v=vs.110%29.aspx).

Running this command:
> ildasm /out=xenia_arkedo "Arkedo Series - 02 SWAP!.exe"

Produces:
* Microsoft.Xna.Framework.RuntimeProfile
* xenia_arkedo
* xenia_arkedo.res

Where work can begin :smiling_imp:

benvanik:
[...]

PR #604, opened 2016-07-23
Find correct launch module for XNA-based games

x1nixmzeng:
Added utility to parse game information and updated launch logic to use it.

Addresses Ben's comment in #529 until we can throw up error messages


x1nixmzeng:
@DrChat 

Where is best to throw up the error messages? I want to avoid `xe::FatalError` because it is non-critical. I could use use `xe::ui::ImGuiDialog::ShowMessageBox`, but I'd need to store a "last error" somewhere. Any ideas :grinning: ?

PR #620, opened 2016-09-17
Bump suggested Python 2.7 version in xb.bat

voltagex:
Probably not that important, but updating it anyway


Blackbird88:
Hmm. Any reason it shouldn't be changed to x64 version? Since Xenia is already x64 only? Just a thought.

PR #627, opened 2016-11-12
Vulkan: point primitive rendering improvements

sabretooth:
Support point primitive size as read from PA_SU_POINT_SIZE on a per-draw-call basis, as well as point size export from the vertex shader. Wire the viewport dimensions (in pixels) into the push constants for the geometry shader using the unused zw components of window_scale and use that to fix the size computation for points. Finally, fix the point sprite coordinate generation logic -- gl_PointCoord is meaningless when we're using a geometry shader to expand points into quads, so we need to generate the texture coordinates ourselves.


src/xenia/gpu/spirv_shader_translator.cc
```diff
@@ -292,6 +298,24 @@ void SpirvShaderTranslator::StartTranslation() {
...
     }
 
+    point_size_ = b.createVariable(spv::StorageClass::StorageClassOutput,
+                                   float_type_, "point_size");
+    b.addDecoration(point_size_, spv::Decoration::DecorationLocation, 17);
+    // Set default point-size value (-1.0f, indicating to the geometry shader
+    // that the register value should be used instead of the per-vertex value)
+    b.createStore(point_size_, b.makeFloatConstant(-1.0f));
+
+    point_coord_ = b.createVariable(spv::StorageClass::StorageClassOutput,
```
DrChat: Do we need to declare point_coord in vertex shaders if the value is never used?

sabretooth: The fragment shader always generates code that reads point_coord (the ps_param_gen code path). We might be able to get rid of it if we knew what the primitive type was at shader translation time, since afaict it only applies to point lists, but the shader translator doesn't seem to have that information available.

In any case, it's probably cheap enough to just load a extra zero into an output to not really worry about.

sabretooth:
I was working on this patch for missing draws in Geometry Wars RE2. It's got more rendering glitches after this, though -- lots of color swapping, more missing line draws, and post processing issues.

DrChat:
I've started on a blitter for Vulkan that can swap the color channels on
copy.
Colors are being swapped because the xenos renders BGR to render targets
but we render RGB.

See: xenia/ui/vulkan/blitter

sabretooth:
That should certainly help. I'll ignore any leads I follow that lead me towards the bitter, then, until you get that wired in.

sabretooth:
Updated patch just switches the if-chain to an else-if in pipeline_cache.cc.

PR #638, opened 2016-12-01
Update kernel convention for xboxkrnl_io.cc

maxton:


DrChat:
There are some potential buffer overflow issues I'm spotting here - but they're not your fault, they already existed here. I'll fix those myself.

Otherwise, LGTM 👍

PR #643, opened 2016-12-06
* statics initialization order fix

inckie:
Depending on the link order, static object xboxkrnl_exports could be not initialized yet (have zero size) when DECLARE_XBOXKRNL_EXPORT is called.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
@@ -140,9 +140,13 @@ XboxkrnlModule::XboxkrnlModule(Emulator* emulator, KernelState* kernel_state)
       });
 }
 
-std::vector<xe::cpu::Export*> xboxkrnl_exports(4096);
+static auto& get_xboxkrnl_exports() {
+  static std::vector<xe::cpu::Export*> xboxkrnl_exports(4096);
+  return xboxkrnl_exports;
+}
 
 xe::cpu::Export* RegisterExport_xboxkrnl(xe::cpu::Export* export_entry) {
+  auto& xboxkrnl_exports = get_xboxkrnl_exports();
   assert_true(export_entry->ordinal < xboxkrnl_exports.size());
   xboxkrnl_exports[export_entry->ordinal] = export_entry;
   return export_entry;
@@ -158,6 +162,7 @@ void XboxkrnlModule::RegisterExportTable(
 #include "xenia/kernel/xboxkrnl/xboxkrnl_table.inc"
   };
 #include "xenia/kernel/util/export_table_post.inc"
+  auto& xboxkrnl_exports = get_xboxkrnl_exports();
   for (size_t i = 0; i < xe::countof(xboxkrnl_export_table); ++i) {
     auto& export_entry = xboxkrnl_export_table[i];
     assert_true(export_entry.ordinal < xboxkrnl_exports.size());


DrChat:
Looks like you just wrap xboxkrnl_exports. What exactly does this do?

inckie:
DrChat, it resolves the classic C++ statics initialization problem, more info could be found there
[https://isocpp.org/wiki/faq/ctors#static-init-order](https://isocpp.org/wiki/faq/ctors#static-init-order)
This stuff is very annoying, especially in debug builds, when initialization order changes from build to build due to incremental linking.

hlide:
Only my two cents rant here. That's not the best solution. Having the static instances be instantiated at first use through a function means implicite addition of thread-safe code with all the whistles and bells with recent compilers. That's just another bullshit when Construct-On-First-Use-Idiom is just used for preventing from Static-Initialization-Order-Fiasco. As for me, I would put all the singletons in one .cpp file with a clear name in the order I want so I can be sure their initialization are done in the order I gave. One way you can do it:
```
MyHeader1.h:
#ifndef MyHeader1_h
#define MyHeader1_h
...
#include "MyHeader2.h"
...
class MyInstance1 : ...
{
...
};
#ifdef NO_STATIC_INITIALIZATION_ORDER_FIASCO
MyInstance myInstance1(...);
#endif // NO_STATIC_INITIALIZATION_ORDER_FIASCO
#endif // MyHeader1_h
```
```
MyHeader2.h:
#ifndef MyHeader2_h
#define MyHeader2_h
...
class MyInstance2 : ...
{
...
};
#ifdef NO_STATIC_INITIALIZATION_ORDER_FIASCO
MyInstance myInstance2(...);
#endif // NO_STATIC_INITIALIZATION_ORDER_FIASCO
#endif // MyHeader2_h
```
and so on
```
instances.cpp:
#define NO_STATIC_INITIALIZATION_ORDER_FIASCO
#include "MyHeader1.h" // this one uses MyHeader2.h so MyIntance2 instantiation will come before MyInstance1 instantiation in order
#include "MyHeader2.h" // no instantiation because it already done by MyHeader1.h
#include "MyHeader3.h"
#include "MyHeader4.h"
```

inckie:
Yeah, I pretty much understand the complications. Also, there are some nasty side effects of static auto-registrators like DECLARE_XBOXKRNL_EXPORT when used in the libraries with intensive linking optimization. I can go on pretty much far on this topic, but I only did a minimal fix for application to run for me.
Not sure about threading, have not heard of the compilers with multi-threaded statics initialization, and this code is seems to be used in statics only for now. But some warnings could be added, just in case.
Btw, there is the same code in XamModule.

hlide:
Example:
```
#include <iostream>

struct A {
  A() { std::cout << "A::A" << std::endl; }
  void f() { std::cout << "A::f" << std::endl; } 
};

A & a() {
  static A instance;
  return instance;
}
 
struct B {
  B() { std::cout << "B::B" << std::endl; a().f(); }
};
```

Assembly result of a():
```
a():
        push    rbp
        mov     rbp, rsp
        push    r12
        push    rbx
        movzx   eax, BYTE PTR guard variable for a()::instance[rip]
        test    al, al
        sete    al
        test    al, al
        je      .L4
        mov     edi, OFFSET FLAT:guard variable for a()::instance
        call    __cxa_guard_acquire
        test    eax, eax
        setne   al
        test    al, al
        je      .L4
        mov     r12d, 0
        mov     edi, OFFSET FLAT:a()::instance
        call    A::A()
        mov     edi, OFFSET FLAT:guard variable for a()::instance
        call    __cxa_guard_release
.L4:
        mov     eax, OFFSET FLAT:a()::instance
        jmp     .L9
        mov     rbx, rax
        test    r12b, r12b
        jne     .L7
        mov     edi, OFFSET FLAT:guard variable for a()::instance
        call    __cxa_guard_abort
.L7:
        mov     rax, rbx
        mov     rdi, rax
        call    _Unwind_Resume
.L9:
        pop     rbx
        pop     r12
        pop     rbp
        ret
```
Oh God.... all that code is useless in our case.

Assembly result of B::B():
```
B::B():
        push    rbp
        mov     rbp, rsp
        sub     rsp, 16
        mov     QWORD PTR [rbp-8], rdi
        mov     esi, OFFSET FLAT:.LC2
        mov     edi, OFFSET FLAT:std::cout
        call    std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*)
        mov     esi, OFFSET FLAT:std::basic_ostream<char, std::char_traits<char> >& std::endl<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&)
        mov     rdi, rax
        call    std::basic_ostream<char, std::char_traits<char> >::operator<<(std::basic_ostream<char, std::char_traits<char> >& (*)(std::basic_ostream<char, std::char_traits<char> >&))
        call    a()
        mov     rdi, rax
        call    A::f()
        nop
        leave
        ret
```
*facepalm* totally stupid call to a() with all that whistles and bells...

inckie:
Well, nice to know that they've finally protected statics initialization.  I hope we can survive this performance hit in all 144 calls we have, so program will at least able to start.
I would actually won't use singletons and globals in the first place.

paulsapps:
You can disable thread safe init of statics if you wish (in MSVC at least), probably not recommended though. 

Function local static is the usual workaround instead of really invasive changes of auto generating code or removing all globals...

hlide:
Sorry to say, I don't care whether Xenia uses it. Usage of Construct-On-First-Use-Idiom is plain wrong: I had a lot of worm cans because of them. By the way, when destructors of "new-ed" instances in those COFUI functions are called? never! they are just bad practice, period.

inckie:
I did not use 'new', so destructros will be called in atexit(), as usual.
From the assembly I can see that the check of the statics initialization flag is done before the mutexes, i.e. a double check locking is performed, so no real performance hit. And this place is definitely not a bottleneck in any way. I find this workaround perfectly fine in that particular case. If we want to discus practical side of this pattern in general, this place does not seems to be a good place for it.
I would be more concerned about DECLARE_XBOXKRNL_EXPORT macro because if it happens to occur in somewhat isolated compilation unit (no calls to any functions in the same file from the outside code) in a static link library, it will be stripped out during linking stage, and that will take a while to trace.

hlide:
"I did not use 'new', so destructros will be called in atexit(), as usual." That's worst, you may get a static deinitialization disaster.

inckie:
As I said, pretty much familiar with statics behavior and trade-offs.  There are **books** on that topic. But all this is completely irrelevant is this case. I don't think someone will get a 'disaster' on std::vector of pointers. Btw, why not an std::array, since there is no resize, anyway?  
I maybe new to github, but is it always like that? What the point of the discussion?

DrChat:
Could you remove that last commit? Looks like that's intended for another PR.

inckie:
Oops, I did not intend to push that branch. Removed.

DrChat:
Gonna merge anyways, because this fixes an annoying issue when working on xboxkrnl exports.

inckie:
Yay!

Issue #631, opened 2016-11-19
Convert old-style kernel convention to new-style convention


DrChat:
Just rewrite exports using the old shim convention to the new convention.

```
SHIM_CALL XAudioGetSpeakerConfig_shim(PPCContext* ppc_context,
                                      KernelState* kernel_state) {
  uint32_t config_ptr = SHIM_GET_ARG_32(0);

  XELOGD("XAudioGetSpeakerConfig(%.8X)", config_ptr);

  SHIM_SET_MEM_32(config_ptr, 0x00010001);

  SHIM_SET_RETURN_32(X_ERROR_SUCCESS);
}

...

SHIM_SET_MAPPING("xboxkrnl.exe", XAudioGetSpeakerConfig, state);
```

Becomes

```
dword_result_t XAudioGetSpeakerConfig(lpdword_t config_ptr) {
  *config_ptr = 0x80030001;
  return X_ERROR_SUCCESS;
}
DECLARE_XBOXKRNL_EXPORT(XAudioGetSpeakerConfig, ExportTag::kImplemented);

// (Call to SHIM_SET_MAPPING removed)
```

Fairly easy to do, just use others already converted as guidelines if you get lost. Be sure to add appropriate ExportTags and test the exports after conversion to make sure they still work!

mmdanggg2:
Hey, I was trying to do this but I'm just getting an exception after loading an iso. Not too sure how to debug it, It's crashing at shim_utils.h L423 and the string_buffer has got XAudioGetVoiceCategoryVolumeChangeMask in it so I assume that's where the problem is.

Unhandled exception at 0x00007FF622FB9F12 in xenia.exe: 0xC0000005: Access violation reading location 0x0000000241550000.

You can check the code https://github.com/mmdanggg2/xenia/blob/master/src/xenia/kernel/xboxkrnl/xboxkrnl_audio.cc

Thoughts/Suggestions?

DrChat:
You're dereferencing driver_ptr, which is deliberately an invalid pointer.

mmdanggg2:
Don't think that's what's causing this problem, even with both lines commented out and just a return, it crashes the same.

Nukem9:
@mmdanggg2 I'd assume all instances of `*driver_ptr` need to be replaced like this or something similar:

`assert_true((*driver_ptr & 0xFFFF0000) == 0x41550000);`
`assert_true((driver_ptr.guest_address() & 0xFFFF0000) == 0x41550000);`

[...]

PR #649, opened 2016-12-20
Update kernel convention for xboxkrnl_rtl.cc, xboxkrnl_ob.cc

maxton:
Progress on #631

I'm pretty sure these changes are correct, although what could be helpful in the future is unit testing all these kernel exports to make sure there aren't regressions. Or maybe there is a list somewhere of which games use which exports so I can do some cursory checks?

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
@@ -232,7 +232,7 @@ dword_result_t RtlUnicodeStringToAnsiString(
 DECLARE_XBOXKRNL_EXPORT(RtlUnicodeStringToAnsiString, ExportTag::kImplemented);
 
 // https://msdn.microsoft.com/en-us/library/ff553113
-dword_result_t RtlMultiByteToUnicodeN(pointer_t<uint16_t> destination_ptr,
+dword_result_t RtlMultiByteToUnicodeN(lpword_t destination_ptr,
                                       dword_t destination_len,
                                       lpdword_t written_ptr,
                                       pointer_t<uint8_t> source_ptr,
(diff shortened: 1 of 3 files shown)


SakataGintokiYT:
I found regressions in Bomberman games
**[Bomberman Live]**
Fotoszok.pl
**[Bomberman Battlefest]**
Fotoszok.pl
**[Bomberman Live.Log:](https://gist.github.com/SakataGintokiYT/66deaecb2f230cc3045b1da1303f1774)**
**[Bomberman Battlefest.Log:](https://gist.github.com/SakataGintokiYT/acb09bdf62e5d71d214f11bf5d19fe38)**

src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
```diff
@@ -188,106 +188,97 @@ void RtlFreeUnicodeString(pointer_t<X_UNICODE_STRING> string) {
...
   copy_len = copy_len < destination_len ? copy_len : destination_len;
 
   // TODO(benvanik): maybe use UnicodeToMultiByte on Win32?
-
-  auto source = reinterpret_cast<uint16_t*>(SHIM_MEM_ADDR(source_ptr));
-  auto destination = reinterpret_cast<uint8_t*>(SHIM_MEM_ADDR(destination_ptr));
   for (uint32_t i = 0; i < copy_len; i++) {
-    uint16_t c = xe::byte_swap(*source++);
-    *destination++ = c < 256 ? (uint8_t)c : '?';
+    uint16_t c = source_ptr[i];
```
mrRosset: Shouldn't there still be a byte swap here ?

maxton: The byte swap happens transparently because source_ptr is a pointer to a xe::be<uint16_t>

maxton: Oh, wait, you're right, that only seems to happen for uint32_t

maxton: Ok, I added the type lpword_t to handle this the same way as lpdword_t, etc... now it is actually doing the byte-swapping

DrChat:
Looks pretty solid to me 👍

PR #650, opened 2016-12-24
Regression fix (Air Conflicts Secret Wars)

Nmzik:
Game doesn't boot past loading screen (The guest has crashed). Also fixes game "Just Cause 2".

### DIFF
--- a/src/xenia/memory.cc
+++ b/src/xenia/memory.cc
@@ -24,7 +24,7 @@
 // TODO(benvanik): move xbox.h out
 #include "xenia/xbox.h"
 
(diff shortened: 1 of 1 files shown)


gibbed:
This is not a valid fix for the issue. Games attempting to write to the 0 page indicate bad behavior of some sort which needs to be corrected.

If you want to boot said games doing bad behavior, use --protect_zero=false on the command-line.

Margen67:
Is there a list of launch parameters somewhere?

gibbed:
I think you can give Xenia --help to list all of them. The DEFINE_* macros are defining available command-line arguments.

PR #655, opened 2017-01-06
Make vfs::Entry time stamps consistent (as Windows FILETIME)

inckie:
vfx::Entry originated from filesystem_win.cc holds file time fields in FILETIME format (100-nanosecond intervals since January 1, 1601), while StfsContainerDevice provides UNIX epoch time (or something close to that), and DiscImageDevice just sets it to zero. This leads RtlTimeToTimeFields  to crash later.

This can be fixed in filesystem_win.cc and NtQueryFullAttributesFile/NtQueryFullAttributesFile, but this place seems to be a better fit.

I have it tested on two games only (both were crashing before), so it would be nice for someone to try some more discs/images.

### DIFF
--- a/src/xenia/vfs/devices/disc_image_device.cc
+++ b/src/xenia/vfs/devices/disc_image_device.cc
@@ -127,9 +127,11 @@ bool DiscImageDevice::ReadEntry(ParseState* state, const uint8_t* buffer,
   entry->attributes_ = attributes | kFileAttributeReadOnly;
   entry->size_ = length;
   entry->allocation_size_ = xe::round_up(length, bytes_per_sector());
-  entry->create_timestamp_ = 0;
-  entry->access_timestamp_ = 0;
-  entry->write_timestamp_ = 0;
+
+  // Set to January 1, 1601 (UTC) in 100-nanosecond intervals
+  entry->create_timestamp_ = 10000 * 11644473600000LL;
+  entry->access_timestamp_ = 10000 * 11644473600000LL;
+  entry->write_timestamp_ = 10000 * 11644473600000LL;
 
   if (attributes & kFileAttributeDirectory) {
     // Folder.
--- a/src/xenia/vfs/devices/disc_image_device.cc
+++ b/src/xenia/vfs/devices/disc_image_device.cc
@@ -128,7 +128,7 @@ bool DiscImageDevice::ReadEntry(ParseState* state, const uint8_t* buffer,
   entry->size_ = length;
   entry->allocation_size_ = xe::round_up(length, bytes_per_sector());
 
-  // Set to January 1, 1601 (UTC) in 100-nanosecond intervals
+  // Set to January 1, 1970 (UTC) in 100-nanosecond intervals
   entry->create_timestamp_ = 10000 * 11644473600000LL;
   entry->access_timestamp_ = 10000 * 11644473600000LL;
   entry->write_timestamp_ = 10000 * 11644473600000LL;
--- a/src/xenia/vfs/devices/stfs_container_device.cc
+++ b/src/xenia/vfs/devices/stfs_container_device.cc
@@ -16,6 +16,11 @@
 #include "xenia/base/math.h"
 #include "xenia/vfs/devices/stfs_container_entry.h"
 
+#if XE_PLATFORM_WIN32
+#include "xenia/base/platform_win.h"
+#define timegm _mkgmtime
+#endif
+
 namespace xe {
 namespace vfs {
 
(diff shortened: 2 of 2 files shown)


src/xenia/vfs/devices/disc_image_device.cc
```diff
@@ -127,9 +127,11 @@ bool DiscImageDevice::ReadEntry(ParseState* state, const uint8_t* buffer,
   entry->attributes_ = attributes | kFileAttributeReadOnly;
   entry->size_ = length;
   entry->allocation_size_ = xe::round_up(length, bytes_per_sector());
-  entry->create_timestamp_ = 0;
-  entry->access_timestamp_ = 0;
-  entry->write_timestamp_ = 0;
+
+  // Set to January 1, 1601 (UTC) in 100-nanosecond intervals
```
DrChat: Wouldn't 0 be the value for January 1, 1601? Did you mean to write Jan 1, 1970?

DrChat:
This seems to be a pretty solid fix - one question though:

Did loading games in their STFS containers cause them to get hung up on this issue before applying this fix?
I don't think Unix time can reach can reach past Jan 1, 1970 when it was being improperly scaled.

inckie:
Games were crashing in RtlTimeToTimeFields, not hung. I've temporary suppressed the crash by hacking the RtlTimeToTimeFields and everything worked fine. Unpacking the .iso and running .xex directly worked as well. I've got curious on that, but did not investigate the reasons beyond that point until tonight.

After applying the fix there were no changes in the behavior from workarounds above.
The reason of the crash is adjusting made in RtlTimeToTimeFields by subtracting a constant. It's not just linear scaling: subtracting from zero (and STFS time stamp as well) gives negative times. I've actually checked that time values from vfs::Entry did occur in RtlTimeToTimeFields later.
According to [http://www.arkem.org/xbox360-file-reference.pdf](http://www.arkem.org/xbox360-file-reference.pdf), STFS holds timestamp in FAT format, but proper parsing still gave me dates about 7 years off from the release date, so more investigations needed.

src/xenia/vfs/devices/disc_image_device.cc
inckie: Nope, it's 1601 for some strange reasons: [https://msdn.microsoft.com/en-us/library/windows/desktop/ms724284(v=vs.85).aspx](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724284(v=vs.85).aspx)

wellerallergy:
Thanks for taking a look at this. Looking forward to it being merged. You wouldn't happen to have a compiled version with the fix applied that you could share? Would like to try it on Scott Pilgrim (which is what originally made me open up the issue). Would attempt to apply your fix and compile myself but I'm going to be away from my VS install for a while.

inckie:
Now FAT (or XTAF, whatever) timestamps are parsed properly.
It would be nice to have a file for adapters like #define timegm _mkgmtime though.

inckie:
Not sure how I did receive the message and can't see the post in the thread (was it deleted?).
There is a binary:
https://drive.google.com/file/d/0B7brqi7-nXdeWXRMazF1WlBWT0U/view?usp=sharing

rlabrecque:
Wasn't just you that got that, don't worry!

src/xenia/vfs/devices/disc_image_device.cc
DrChat: This comment should indeed say January 1, 1970.

inckie: Nope.
If we track the usage of this data, it will end in the pair of functions NtQueryInformationFile and NtQueryFullAttributesFile, that have respective WinAPI originals, and they are using structure X_FILE_NETWORK_OPEN_INFORMATION, that just a copy of FILE_NETWORK_OPEN_INFORMATION, and if we check the [docs](https://msdn.microsoft.com/en-us/library/windows/hardware/ff545822(v=vs.85).aspx), we will read "100-nanosecond intervals since the start of the year 1601 in the Gregorian calendar.". 
Next, we can see that values are just copied:
```
auto file_info = file_info_ptr.as<X_FILE_NETWORK_OPEN_INFORMATION*>();
file_info->creation_time = file->entry()->create_timestamp();
```
And these two functions are only one who uses that information.
Therefore, there is no gain in recalculating it to Unix timestamps, and zero is definitely January 1, 1601. Also, filesystem_win.cc just stores values from FILETIME as it is, and it is 1601-based too.

DrChat: Yes it should - you're using the magic number to set the time to Jan 1, 1970 from the timebase of Jan 1, 1601.

inckie: Ops, my bad. I thought about another point in the code. Already forgot that I've done there. I've added the offset so later time conversion won't crash, yes,

PR #662, opened 2017-01-21
third_party: vulkan: loader: fix Linux build

resetnow:
* Remove Win32 dirent implementation from Linux build
* Fix a bunch of clang warnings
* Add default paths that would otherwise be passed from CMake

### DIFF
--- a/third_party/vulkan/loader/premake5.lua
+++ b/third_party/vulkan/loader/premake5.lua
@@ -2,7 +2,7 @@ group("third_party")
 project("vulkan-loader")
   uuid("07d77359-1618-43e6-8a4a-0ee9ddc5fa6a")
   kind("StaticLib")
-  language("C++")
+  language("C")
 
   defines({
     "_LIB",
@@ -22,4 +22,11 @@ project("vulkan-loader")
     characterset("MBCS")
     defines({
       "VK_USE_PLATFORM_WIN32_KHR",
+    })
+  filter("platforms:not Windows")
+    removefiles("dirent_on_windows.c")
+  filter("platforms:Linux")
+    defines({
+      [[SYSCONFDIR="\"/etc\""]],
+      [[DATADIR="\"/usr/share\""]],
     })


third_party/vulkan/loader/loader.c
```diff
@@ -3789,8 +3789,7 @@ VkResult loader_create_instance_chain(const VkInstanceCreateInfo *pCreateInfo,
                 continue;
             if ((fpGIPA = layer_prop->functions.get_instance_proc_addr) ==
                 NULL) {
-                if (layer_prop->functions.str_gipa == NULL ||
-                    strlen(layer_prop->functions.str_gipa) == 0) {
+                if (strlen(layer_prop->functions.str_gipa) == 0) {
```
x1nixmzeng: This will crash if
```cpp
    layer_prop->functions.str_gipa
```
is **NULL** - you will need to keep the null-check before using `strlen`!

resetnow: Correct me if I'm wrong, but `layer_prop->functions.str_gipa` is an array of fixed size allocated as a part of `functions` struct — so if `functions` is not NULL, str_gipa isn't either.

resetnow: https://github.com/benvanik/xenia/blob/8d476fc845f9535e4f2d3ef15089a4bdab364ac8/third_party/vulkan/loader/loader.h#L115

paulsapps: Yup, checking an array for NULL is pointless and probably wrong

x1nixmzeng: I should have checked the context before commenting! Thanks for clarifying!

DrChat:
Nice work - though I may not be able to merge the changes to the third party source, as we just copy this code directly from the Vulkan repository.

Could you check the latest version of the Vulkan loader with Xenia and see if it addresses linux? If not, definitely submit a pull request for them and then I'll update this code.

However, I can pull the premake file if you make that the only change in this PR.

resetnow:
The upstream vulkan repo seems to have these fixes already: https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/commit/91b85f4715990437d728bfd80266cd7eb9f09866 Should I remove these changes from PR so you can pull it from there?

resetnow:
As for default paths, I'll try to pass them from premake.

resetnow:
I have updated the PR to only contain the premake change, it should build after upstream vulkan changes are pulled.

PR #663, opened 2017-01-25
Linux + clang build fixes

resetnow:
* Fixed missing includes (cmath, cfloat)
* Added const specifiers to some functions in SPIR-V builder so there aren't any lvalue-references to temporary objects in spirv_shader_translator
* Fixed crossed variable initialization
* xb format did quite a few changes

FTR I was building with clang 3.8.0 on Fedora 25

### DIFF
--- a/src/xenia/gpu/trace_viewer.cc
+++ b/src/xenia/gpu/trace_viewer.cc
@@ -404,7 +404,7 @@ int TraceViewer::RecursiveDrawCommandBufferUI(
         }
 
         ImGui::PushID(int(i));
-        if (ImGui::TreeNode((void*)0, "Indirect Buffer %d", i)) {
+        if (ImGui::TreeNode((void*)0, "Indirect Buffer %" PRIu64, i)) {
           ImGui::Indent();
           auto id = RecursiveDrawCommandBufferUI(
               frame, buffer->commands[i].command_subtree.get());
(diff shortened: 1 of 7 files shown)


resetnow:
Regarding third party code (SpvBuilder) — these changes aren't in the upstream repository, should I make a PR there?

resetnow:
SpvBuilder change was merged upstream (https://github.com/KhronosGroup/glslang/pull/702), so now it's a matter of updating the code from there.

src/xenia/gpu/trace_viewer.cc
```diff
@@ -404,7 +404,7 @@ int TraceViewer::RecursiveDrawCommandBufferUI(
         }
 
         ImGui::PushID(int(i));
-        if (ImGui::TreeNode((void*)0, "Indirect Buffer %d", i)) {
+        if (ImGui::TreeNode((void*)0, "Indirect Buffer %zu", i)) {
```
DrChat: Change this to use PRIX64

src/xenia/gpu/trace_viewer.cc
```diff
@@ -1563,7 +1563,7 @@ void TraceViewer::DrawStateUI() {
       vertices.resize(size / 4);
       QueryVSOutput(vertices.data(), size);
 
-      ImGui::Text("%d output vertices", vertices.size() / 4);
+      ImGui::Text("%zu output vertices", vertices.size() / 4);
```
DrChat: Ditto

resetnow:
Updated. 

>PRIx64

This assumes size_t is 64-bit, right? Is this because VS doesn't support C99 "z" specifier?

DrChat:
Yes - and I made a mistake, I meant to say PRIu64 rather than PRIx64.

resetnow:
Updated again, so why not use `z`?

DrChat:
I couldn't see anything that says visual studio supports the `z` specifier.

PR #667, opened 2017-02-11
Regression fix

MattKackles:
Should fix GTA IV crash, probably goes ingame once again.
Fixes some games not rendering anymore.


MattKackles:
Sorry about the stupid commit name, forgot that I was making this a pr.
Anyways, it should compile now.
:)

rlabrecque:
Can you rebase those two commits into one and fix the commit name at the same time, then force push?

DrChat:
I made the code differentiate between internal format and format for a good reason I can't remember now.

What you can do instead is if the format is e.g. 2_10_10_10_unknown, convert it to 2_10_10_10.
I believe there's 2 unhandled cases of that nature.

MattKackles:
@DrChat 
I am closing this PR, it actually does more harm than it does good.
I thought it was the internal format causing GTA IV to crash, but it seems its something else.

PR #676, opened 2017-02-24
Various fixes

parkerlamb:
These are fixes for some small things I came across:

- Some games load code from separate .dlls which do not have import headers. I added a check for this and a warning is written to the log.
- Games may intentionally ask for too much memory (see PR #376) but there is an assert every time it happens, making it difficult to do other debugging. This assert is removed here.
- Some games have enormous shaders (~1000 lines disassembled) which were causing an overflow in the logger format buffer.

### DIFF
--- a/src/xenia/base/logging.cc
+++ b/src/xenia/base/logging.cc
@@ -228,11 +228,23 @@ void InitializeLogging(const std::wstring& app_name) {
   logger_ = new (mem) Logger(app_name);
 }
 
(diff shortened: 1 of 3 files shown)


DrChat:
Thanks for bringing these issues up 👍.

Committed these changes in [this set of changes](https://github.com/benvanik/xenia/compare/ed56f7e96bf3...ad5748f3e782).

PR #682, opened 2017-03-19
Faster Calculations in Texture Info

Ershany:
I thought this might be positive fix


DrChat:
Unfortunately - I won't be pulling this PR because @gibbed and I feel that these changes will obfuscate the code, plus the compiler can make these optimizations automatically.

Appreciate the try though - Good effort 👍

rlabrecque:
I'd recommend realworld perf comparisons if you're doing optimization work too!

PR #684, opened 2017-03-26
Ringbuffer init fix

cesys:
Must reset read_ptr_index_ in CommandProcessor::InitializeRingBuffer becaue a game can initialize the ring buffer multiple times.
Left the new condition on the assertion as the value 0xFFFFFF seems to acceptable and it's preventing 4B4F07FE to go ahead.


DrChat:
I'll let it slide for now - but next time, don't combine separate changes to one commit

Nice fix though - good work :)

cesys:
Sure. In this case the second fix was hidden by the first. But I understand.

PR #573, opened 2016-05-12
Linux porting work

dougvj:
Here is a first stab of implementing several backends for the linux port, most notably porting UI code to GTK+3, and abstracting the GL context creation. 

Currently the code compiles on Linux, but there are several linker errors preventing full compilation and preventing testing of the code. There is a lot of work left to be done, but it's progress.

I don't have access to a windows build environment so I need these changes tested for that platform as well.

PR #694, opened 2017-04-27
Building xenia-app on Linux (almost)

sephiroth99:
This PR makes xenia-app build on Linux as far as it can before the linker complains about undefined references to stuff not implemented yet for Linux (mainly xe::loop, xe::window, xe::threading, xe::filesystem, and some other things). I already have some code for some of that missing stuff so expect another PR "soon".

I am planning on using XCB for window/event management (X server interface basically), so you see some initial references to that in the PR.

I disabled the gl4 video backend on Linux for two reasons:
1. GLX requires Xlib, so by wanting to use XCB, you are left with implementing a Xlib+XCB hybrid, which does not seem super inviting...
2. Mesa does not support all extensions required by Xenia yet anyway.

Finally about commit 247bf6d, turns out std::identity is not a thing anymore, so I replaced it with an example from a Microsoft help page about that. I'm open on suggestions to fix this issue differently, right now it's only done that way so it compiles.

Comments welcome.


TotalCaesar659:
Hello!

1. Could you add a manual for compiling on Linux like

> sudo apt-get install <...>
> mkdir build
> cd build
> ../configure --prefix=/usr/
> make

or something like this.
2. It would be good to make a list of OpenGL extensions that Mesa doesn't support now (with checkboxes). Maybe somebody try to ask them about implementing.
3. Many of Linux distributions will go from X11 to Wayland, but XCB is based on X11. This is strange moment, I guess.

resetnow:
>Could you add a manual for compiling on Linux

Last time I tried to compile on Linux, it was `./xenia-build build`

sephiroth99:
@TotalCaesar659 Like @resetnow said, there is no special procedure to build the Linux version. Right now it is configured to use clang 3.8 by default but you can override CC and CXX environment variables to a newer clang without (much) issue. Also to note that some of the targets don't build at all (e.g. trace viewers), so they can fail the build early before xenia-app had the chance to completely finish building, so you can build only xenia-app by calling the build script like so
`./xenia-build build --target=xenia-app`

As for your second point, I said extension**s** but really the only one I know about is GL_ARB_bindless_texture. I am not an expert on the gl4 backend so maybe more extensions are needed.

Finally, regarding Wayland, I agree that some distributions are moving (or planning to move) to Wayland. I thought of going Wayland only (I do use it on my machine), but then people staying on X would never have a way to run Xenia. With XWayland, people using Wayland-based distros would be able to run Xenia in addition to people still using X.

src/xenia/base/bit_field.h
```diff
@@ -29,7 +34,7 @@ struct bf {
   // For enum values, we strip them down to an underlying type.
   typedef
       typename std::conditional<std::is_enum<T>::value, std::underlying_type<T>,
-                                std::identity<T>>::type::type value_type;
+                                xe::identity<T>>::type::type value_type;
```
DrChat: Why swap from std::identity? Does gcc not support this?

src/xenia/gpu/gl4/premake5.lua
```diff
@@ -21,6 +21,7 @@ project("xenia-gpu-gl4")
   includedirs({
     project_root.."/third_party/gflags/src",
   })
+  filter("platforms:Windows")
```
DrChat: Looks like you're disabling this project - could just do this from where it's included instead.

src/xenia/base/bit_field.h
resetnow: std::identity is not in a standard. It was present in some pre c++11 drafts, but was removed later. So now it's specific to VS compiler.

paulsapps: I thought std::forward replaced it?

sephiroth99: Yeah LLVM's libc++ and GCC's libstdc++ do not have std::identity as it was never finalised in a standard, but MS kept it in their c++ std library...

I'll look into std::forward.

src/xenia/gpu/gl4/premake5.lua
sephiroth99: I'll look if I can make gl4 Windows only from the main premake file, similar to xaudio2 and friends.

TotalCaesar659:
> As for your second point, I said extensions but really the only one I know about is GL_ARB_bindless_texture. I am not an expert on the gl4 backend so maybe more extensions are needed.
@sephiroth99 Today ARB_bindless_texture has landed to Mesa Git.
https://cgit.freedesktop.org/mesa/mesa/commit/?id=65d1e4d1eb08dda16d9b5967b4a6a414ba775815

Margen67:
@TotalCaesar659 OpenGL is deprecated.

TotalCaesar659:
@Margen67 I know, just for the information.

sephiroth99:
Allright I did not forget about this!

I made the change about how gl4 is now Windows only.

About std::identity, I did some research and in the end, it seems you can use std::remove_reference. For non-references it does exactly the same thing, and for reference types it just remove the reference. In the specific case of the bitfield struct, from my quick check it doesn't seem like any type used is of the reference type, so there is no issue. The other alternatives would be std::decay (it does, among other things, std::remove_reference and other changes) or std::common_type with only one type (but that's the same thing as std::decay).

I added one commit for a missing function in Profiler when disabled.

Also, I rebased on latest master and there were some additional changes needed.

Thanks!

tetration:
Cant your guys use Vulkan for the linux version instead of OpenGL?

Razzile:
@tetration opengl backend is deprecated anyway so that's the most likely outcome

DrChat:
@sephiroth99 Has this PR been merged with #573 already?

PR #697, opened 2017-05-02
Specify python2 in xenia-build

Kriskras99:
Otherwise it breaks on Arch Linux

### DIFF
--- a/xenia-build
+++ b/xenia-build
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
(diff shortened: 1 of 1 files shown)


zephirusgit:
download link? :⭕️

Margen67:
Isn't this just a comment? Shouldn't it have no effect?

DrChat:
@Margen67: See [this](https://en.wikipedia.org/wiki/Shebang_(Unix)).

PR looks good to me 👍, pulling.

PR #700, opened 2017-05-08
Actually open the file picker when Ctrl+O is pressed

### LINKED ISSUE #698: CTRL+O shortcut doesn't function
I'm finding that when I open Xenia and press CTRL+O, it does not launch the File>Open function even though the File menu indicates that keyboard shortcut is valid.  To rule out an issue with my keyboard (one of those where function lock is forced enabled by default in Windows 10) I verified that this key sequence works in other applications.

tgsm:
Closes #698.

I'm somewhat new to C++, however this should be okay.

### DIFF
--- a/src/xenia/app/emulator_window.cc
+++ b/src/xenia/app/emulator_window.cc
@@ -88,6 +88,11 @@ bool EmulatorWindow::Initialize() {
   window_->on_key_down.AddListener([this](KeyEvent* e) {
     bool handled = true;
     switch (e->key_code()) {
+      case 0x4F: {  // O
+        if (window_->modifier_cntrl_pressed_) {
+          FileOpen();
+        }
+      } break;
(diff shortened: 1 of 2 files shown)


DrChat:
It isn't a good idea to use the modifiers on the Window - the KeyEvent could be stored and accessed later.

Rather than making modifiers public, put them into KeyEvent. You can modify the constructor of KeyEvent to take those modifiers as bools, and then where the event is created in Win32Window::HandleKeyboard, try using [VkKeyScan](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646329(v=vs.85).aspx) to scan for the modifiers.

Nice first try though :)

tgsm:
Thank you for the feedback!

I'm gonna go ahead and close this PR for now, whenever I get time to work on a proper fix I will just open a new one.

Issue #693, opened 2017-04-26
Xenia build fails with Windows SDK 10.0.14393.0


Margen67:
If you go into each of them and change the SDK used to 14393 it gets reverted unless you set them to read only. https://gist.github.com/Margen67/67be8626e0c8d5d2aae3e5bb31d7369b

By default this happens:
```C:\Users\marge\Documents\GitHub\xenia>
xb build
Building debug...

- running premake...
** Warning: the flags value Unicode has been deprecated.
   Use `characterset "Unicode"` instead
** Warning: the flags value Symbols has been deprecated.
   Use `symbols "On"` instead
Building configurations...
    Baking capstone...
    Baking gflags...
    Baking glew...
    Baking glslang-spirv...
    Baking imgui...
    Baking libavcodec...
    Baking libavutil...
    Baking snappy...
    Baking spirv-tools...
    Baking vulkan-loader...
    Baking xxhash...
    Baking yaml-cpp...
    Baking xenia-core...
    Baking xenia-app...
    Baking xenia-apu...
    Baking xenia-apu-nop...
    Baking xenia-base...
    Baking xenia-base-tests...
    Baking xenia-cpu...
    Baking xenia-cpu-tests...
    Baking xenia-cpu-ppc-tests...
    Baking xenia-cpu-backend-x64...
    Baking xenia-debug-ui...
    Baking xenia-gpu...
    Baking xenia-gpu-shader-compiler...
    Baking xenia-gpu-null...
    Baking xenia-gpu-gl4...
    Baking xenia-gpu-gl4-trace-viewer...
    Baking xenia-gpu-gl4-trace-dump...
    Baking xenia-gpu-vulkan...
    Baking xenia-gpu-vulkan-trace-viewer...
    Baking xenia-gpu-vulkan-trace-dump...
    Baking xenia-hid...
    Baking xenia-hid-demo...
    Baking xenia-hid-nop...
    Baking xenia-kernel...
    Baking xenia-ui...
    Baking xenia-ui-gl...
    Baking xenia-ui-window-gl-demo...
    Baking xenia-ui-spirv...
    Baking xenia-ui-vulkan...
    Baking xenia-ui-window-vulkan-demo...
    Baking xenia-vfs...
    Baking xenia-apu-xaudio2...
    Baking xenia-hid-winkey...
    Baking xenia-hid-xinput...
Running action 'vs2017'...
Done (978ms).

- building (all):debug...
[...]

PR #714, opened 2017-06-08
[RFC] Build against Windows 10 SDK

parkerlamb:
Xenia currently builds on Windows 10, though premake defaults to using the Windows 8.1 SDK when generating VS projects. This PR allows for building Xenia with only a Windows 10 SDK installed.

There are two issues which need to be resolved before merging:

1. The code in this PR selects the most recent Windows 10 SDK on the system by looking at the registry. This is flexible but tends to imply that Xenia supports _all_ of the SDKs which may not be a good idea. Alternatively, we could pick a version and stick with it, though at the rate Microsoft puts out new Windows 10 builds it'd probably have to get bumped pretty frequently.

2. The premake version needs to be bumped. The binary at /tools/build/bin/premake5.exe is fairly out of date and doesn't report OS versions correctly (it reports Windows 10 as Windows 8 on my system). It could be replaced with a newer version or it could be deleted and the project could build premake from third_party/premake_core automatically. The premake build system needs some tweaks to work on Windows 10 (pretty much identical to what's been done here) but since Xenia has its own version of premake anyway this doesn't seem unreasonable. Building premake from source will probably be required for Linux builds so it may make sense to do it now.

Other things: 
- XInputEnable() has been deprecated under Windows 10, though from looking through the docs it doesn't seem like it was being used correctly here anyway. XInput still works fine with XInputEnable() being removed.
- flags({"WinMain"}) is no longer required for newer versions of premake, kind("WindowedApp") now implies this.

Comments welcome!

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -185,6 +185,13 @@ solution("xenia")
     platforms({"Linux"})
   elseif os.is("windows") then
     platforms({"Windows"})
+
+	-- Determine the Windows SDK version. Adapted from https://github.com/premake/premake-core/issues/651
+	if os.getversion().majorversion == 10 then
+		local win10SDK = os.getWindowsRegistry( "HKLM:SOFTWARE\\Wow6432Node\\Microsoft\\Microsoft SDKs\\Windows\\v10.0\\ProductVersion" )
+		-- It needs an extra ".0" to be recognized by VS
+		if win10SDK ~= nil then systemversion( win10SDK .. ".0" ) end
+	end
   end
   configurations({"Checked", "Debug", "Release"})
 
--- a/premake5.lua
+++ b/premake5.lua
@@ -7,7 +7,7 @@ objdir(build_obj)
 
 -- Define an ARCH variable
 -- Only use this to enable architecture-specific functionality.
-if os.is("linux") then
+if os.istarget("linux") then
   ARCH = os.outputof("uname -p")
 else
   ARCH = "unknown"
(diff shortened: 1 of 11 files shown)


parkerlamb:
Forgot to mention, this fixes #693.

DrChat:
It appears that your change from "os.is" to "os.istarget" has broken the latest builds.
Have you tested this against the premake version we're using?

parkerlamb:
I tested this against the latest commit (52b3b5f) in xenia-project/premake-core. This is not the commit Xenia is currently tracking for the submodule, which is 61d799c8.

The 61d799c8 version does not report OS versions correctly, which is why it needs to be updated. "os.is" has been deprecated in the 52b3b5f version, hence the errors.

The buildbot will fail because the build script prioritizes the precompiled binary at /tools/build/bin/premake5.exe, which is the older version.

As I mentioned above, a decision needs to be made as to whether the precompiled binary should be updated or just move to building premake from source. The submodule commit needs to be bumped either way.

The lint errors are my fault, I'll fix them once everything else is sorted out.

DrChat:
Ah - don't build against premake-core's master branch. We're building against the release branch to keep bleeding-edge bugs away.

parkerlamb:
This PR needs the premake-core master branch to work because of the OS version reporting issue (at least up to commit 0a7b45ce which has the patch for it). If you don't want Xenia to track a newer commit then this PR will have to wait until the premake devs release a new version.

Margen67:
Why was this closed?

parkerlamb:
It's basically a Visual Studio issue. There's a somewhat philosophical discussion on the Premake bug tracker here: [https://github.com/premake/premake-core/issues/935](https://github.com/premake/premake-core/issues/935).

TLDR:
Visual Studio defaults to using the Windows 8.1 SDK when one isn't specified in the config files. There isn't really a good way to determine which SDK to use for Windows 10 because users may have different ones installed on their systems, plus Microsoft pushes a new Windows build every 6 months which means the recent version constantly changes. If we picked a particular SDK to target, it would need to be constantly bumped to match or be out of date on newer Windows builds.

It's my view that this is something Microsoft needs to fix, not hack-fixed in the project files. If you (or anyone else) wants to continue with this PR, feel free to use my changes.

PR #718, opened 2017-06-11
PPC disassembly fixes

parkerlamb:
A few fixes for the debugger PPC disassembly:

- "mtmspr" should be "mtspr"
- FormatDS decode data was missing padding, causing instructions to have incorrect register/constant offsets
- Some of the autogenerated ppc_opcode_* files had been edited directly. ppc-table-gen will now reflect these changes in generated code.


DrChat:
Changes look good to me - nice catch on that padding issue!

PR #722, opened 2017-06-12
Added GPU Register Definition

tmccaff:
Added 0x1924 = 0x6490 according to this referenced here https://github.com/Free60Project/libxenon/blob/master/libxenon/drivers/xenos/xenos.h  and https://www.x.org/docs/AMD/old/RRG-216M56-03oOEM.pdf as DC_LUT_PWL_DATA. Had issue with the other pull request.


DrChat:
@tmccaff If you're making changes - don't close and make a new pull request. They will automatically show up here.

tmccaff:
@DrChat ok I noticed I pulled a request and than made a change that's why I canceled the other request. It's been a long time since I used GitHub. Thanks

Muhammad2324:
I want xenia

PR #729, opened 2017-06-15
Format Shim Conversion Xam_user.cc

drillsar:
Still work has to be done to this file. I wasn't sure what to do with the rest.
Reference #631


maxton:
You know, opening a new GitHub account is _not_ the right way to avoid getting blocked from the project. Instead, you could try listening to the advice that you've been given, or maybe take some time to actually learn how to use git and GitHub.

drillsar:
I did take advice. I guess I won't contribute screw it. I just learned to use git now.

paulsapps:
You didn't take advice at all. There are now 10 pull requests. How hard is it to understand that there should be ONE pull request which you keep UPDATING?

If you don't get it then create your own test repos and play around with it, read git tutorials, it really isn't that hard. In fact creating new PR's is harder. When you have a PR push to your branch and the PR is updated.. as you seem to have done here and in other PRs. Then after a while you create yet another one.. it makes no sense.

PR #732, opened 2017-06-19
Format shim conversion xboxkrnl_threading.cc; xam_user.cc; xam_input.cc

tmccaff:
Still some more work to do removing shim conversions to new format on this file.
#631


Razzile:
you seriously need to stop opening so many pull requests. Just have one and stick with it

tmccaff:
@Razzile  I figured it out now what I was doing wrong. Yes very sorry.

PR #737, opened 2017-06-27
xboxkrnl_threading.cc: PVS-Studio: fixed return value.

SvyatoslavRazmyslov:
We have found and fixed a bug using [PVS-Studio](https://www.viva64.com/en/pvs-studio/) analyzer. 

Return the variable _old_head_, which is always zero.

Analyzer warning: V561 It's probably better to assign value to 'old_head' variable than to declare it anew. Previous declaration: xboxkrnl_threading.cc, line 1256.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_threading.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_threading.cc
@@ -1259,7 +1259,7 @@ pointer_result_t InterlockedPushEntrySList(
     new_hdr.depth = old_hdr.depth + 1;
     new_hdr.sequence = old_hdr.sequence + 1;
 
-    uint32_t old_head = old_hdr.next.next;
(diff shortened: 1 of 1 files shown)


VHGaming:
when will you put xenia back on the website

tmccaff:
@VHGaming what you talking about?

VHGaming:
i'm talking about there being a copy of the emulator on the website they took it down and i know there fixing things right now but i wanna know when they'll be done so we can have the emulator to download from the website

tmccaff:
You always can get the latest release on the website.

DrChat:
Nice catch!

resetnow:
@SvyatoslavRazmyslov was this the only significant bug found or there might be an article about Xenia later?

SvyatoslavRazmyslov:
@resetnow 

Perhaps later I will write an article. Now I don't have much time, so I decided to fix some bugs myself.

PR #741, opened 2017-06-28
Xaudio2 bug fix

### LINKED ISSUE #739: Not sure if this is a bug in Xaudio2
I am troubleshooting audio issues in certain games. I came across this:

In Xenia/Src/xenia/apu/xaudio2/xaudio2_audio_system.cc

On line:
46 it has  xdriver->Shutdown();
47 has assert_not_null(xdriver);

isn't it supposed to be the other way around? 
46 assert_not_null(xdriver);
47  xdriver->Shutdown();

drillsar:
Fixes #739. Call assert_not_null(xdriver) is not needed here. static_cast is not dynamic cast if the pointer is not null it will remain non-zero from what I understand. The assert_not_null(xdriver) should be removed.

### DIFF
--- a/src/xenia/apu/xaudio2/xaudio2_audio_system.cc
+++ b/src/xenia/apu/xaudio2/xaudio2_audio_system.cc
@@ -44,7 +44,6 @@ void XAudio2AudioSystem::DestroyDriver(AudioDriver* driver) {
   assert_not_null(driver);
   auto xdriver = static_cast<XAudio2AudioDriver*>(driver);
   xdriver->Shutdown();
-  assert_not_null(xdriver);
   delete xdriver;
 }
 


VHGaming:
is everything on the emulator bugged out cause this is gonna take a while and i hope you guys turn this in to a top notch emulator

drillsar:
I think most of the issues is Vulkan as this is the rendering engine. The Vulkan code isn't easy at all to code and to understand. I am just looking at Vulkan now and trying to understand it. I would be guessing here you should see a lot of improvements to games in a year or more.

VHGaming:
well i hope we have multiple gpu settings to use and not just vulkin i think using vulkin open GL and DirectX 12 would be good

tmccaff:
OpenGL is no longer supported as the faqs should be updated. Vulkan is better just not fully implemented yet in this project. Direct 12 I think will go as well not sure on that yet.

tmccaff:
You can use OpenGL however using the flag in Xenia like this Xenia --gpu=gl4.

DrChat:
This isn't a bug.

Issue #702, opened 2017-05-13
Crash when attempting to load a file immediately after Xenia starts


AmadeusMop:
Whenever I launch Xenia, there's a short window of time (about 2 seconds) during which loading a game with File->Open will crash Xenia.

As far as I can tell, the crash happens because `Emulator::CompleteLaunch()` is trying to access a kernel module before `Emulator::Setup()` has loaded them.
I have no idea how to fix it, though.

AmadeusMop:
Here's a more thorough explanation of what's going on:

In `xenia_main.cc`, the emulator window is being created _before_ the call to `Emulator::Setup()`.
```c++
  // Create the emulator but don't initialize so we can setup the window.
  auto emulator = std::make_unique<Emulator>(L"");

  // Main emulator display window.
  auto emulator_window = EmulatorWindow::Create(emulator.get());

  // Setup and initialize all subsystems. If we can't do something
  // (unsupported system, memory issues, etc) this will fail early.
  X_STATUS result =
      emulator->Setup(emulator_window->window(), CreateAudioSystem,
                      CreateGraphicsSystem, CreateInputDrivers);
```

As a result, the File menu is accessible while the emulator is still initializing.

Since mouse events are handled on a separate thread, an early click on File->Open can lead to `Emulator::CompleteLaunch()` being called while `Emulator::Setup()` is still executing.

In particular, if the click is fast enough, `CompleteLaunch()` will attempt to access `xam.xex`:
```c++
X_STATUS Emulator::CompleteLaunch(const std::wstring& path,
                                  const std::string& module_path) {
  // Allow xam to request module loads.
  auto xam = kernel_state()->GetKernelModule<kernel::xam::XamModule>("xam.xex");
```
before `Setup()` even has a chance to load it!
```c++
...
  // HLE kernel modules.
  kernel_state_->LoadKernelModule<kernel::xboxkrnl::XboxkrnlModule>();
  kernel_state_->LoadKernelModule<kernel::xam::XamModule>();
  kernel_state_->LoadKernelModule<kernel::xbdm::XbdmModule>();
```

(Note: This only applies to games opened using the File->Open command. Games passed in through command-line arguments are unaffected, because `xenia_main()` only processes command-line arguments after `Emulator::Setup()` has finished.)

DrChat:
Nice catch! I'll have to look into fixing this later.

Margen67:
Does this still occur?

YeimiK:
[...]

PR #743, opened 2017-07-01
Prevent race condition during emulator startup (issue #702)

parkerlamb:
The menu bar on the main window is now greyed out until the emulator is completely initialized.

This is a slightly hacky fix, it would probably be better to make the dependency between the emulator and the window one-way so a proper wait can be done during the initialization phase. However, currently they are both pretty entangled so I'm not sure that'd be worth doing until Xenia has a more permanent UI.

Fixes issue #702

### DIFF
--- a/src/xenia/app/emulator_window.cc
+++ b/src/xenia/app/emulator_window.cc
@@ -266,6 +266,8 @@ bool EmulatorWindow::Initialize() {
 
   window_->Resize(1280, 720);
 
+  window_->DisableMainMenu();
+
   return true;
 }
 
--- a/src/xenia/app/xenia_main.cc
+++ b/src/xenia/app/xenia_main.cc
@@ -178,6 +178,9 @@ int xenia_main(const std::vector<std::wstring>& args) {
     evt->Set();
   });
 
+  // Enable the main menu now that the emulator is properly loaded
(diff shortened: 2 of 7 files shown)


parkerlamb:
Turns out this was two separate race conditions, I somehow didn't hit the on_launch() problem during testing...

Leaving the fixes as two separate commits (since they are technically different problems), I can squash if need be.

drillsar:
@parkerlamb I don't want to hijack this thread, but I used PVS Studio analyzer and a lot of errors show up. One error that caught my eye was this:
V780: The object 'out_info' of a non-passive (non-PDS) type cannot be initialized using the memset function. filesystem_win.cc 164

```
bool GetInfo(const std::wstring& path, FileInfo* out_info) {
  std::memset(out_info, 0, sizeof(FileInfo)); //V780 error on this line
  WIN32_FILE_ATTRIBUTE_DATA data = {0};
  if (!GetFileAttributesEx(path.c_str(), GetFileExInfoStandard, &data)) {
    return false;
  }

parkerlamb:
Did those changes, Window is now passed instead of NativeWindowHandle.

DrChat:
Excellent - thanks for the changes!

PR #744, opened 2017-07-02
32-bit value fix to the 64-bit type.

drillsar:
PVS Studio found error //V629 32-bit value with a subsequent expansion to the 64-bit type


tmccaff:
What?

DrChat:
Nice catch! 👍

Issue #746, opened 2017-07-06
Problems in code from static analyzer log 


drillsar:
PVS Studio found these High errors in the code:

V730 Not all members of a class are initialized inside the constructor. Consider inspecting: handle. immediate_drawer.h 39
V614 The 'backend' smart pointer is utilized immediately after being declared or reset. It is suspicious that no value was assigned to it. emulator.cc 113
V575 The null pointer is passed into 'move' function. Inspect the first argument. emulator.cc 131
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: display_window_. emulator.cc 51
V763 Parameter 'top_down' is always rewritten in function body before being used. memory.cc 1151
V512 A call of the 'memcpy' function will lead to the '& ex_info->ContextRecord->Rax' buffer becoming out of range. exception_handler_win.cc 42
V512 A call of the 'memcpy' function will lead to the '& ex_info->ContextRecord->Xmm0' buffer becoming out of range. exception_handler_win.cc 44
V780 The object 'out_info' of a non-passive (non-PDS) type cannot be initialized using the memset function. filesystem_win.cc 164
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: bSelected. microprofileui.h 1726
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: loaded_, entry_point_. elf_module.cc 22
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: address_reference_count_, instruction_result_count_. function_debug_info.cc 20
V570 The 'constant.i8' variable is assigned to itself. value.cc 1372
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: irql_. processor.cc 80
V512 A call of the 'memcpy' function will lead to overflow of the buffer '& thread_context.Rax'. stack_walker_win.cc 179
V512 A call of the 'memcpy' function will lead to overflow of the buffer '& thread_context.Xmm0'. stack_walker_win.cc 181
V512 A call of the 'memcpy' function will lead to the '& thread_context.Rax' buffer becoming out of range. stack_walker_win.cc 189
V512 A call of the 'memcpy' function will lead to the '& thread_context.Xmm0' buffer becoming out of range. stack_walker_win.cc 191
V569 Truncation of constant value 3735929054. The value range of unsigned char type: [0, 255]. xex_module.cc 393
V569 Truncation of constant value 4027445261. The value range of unsigned char type: [0, 255]. xex_module.cc 409
[...]

PR #749, opened 2017-07-08
It is incorrect to compare the variable of BOOL type with TRUE

drillsar:
Fix V675 main_win.cc 32 It is incorrect to compare the variable of BOOL type with TRUE
Reference #746


drillsar:
I fix this tomorrow.

paulsapps:
I think the fix is just "bool has_console = ::AttachConsole(ATTACH_PARENT_PROCESS) != FALSE;"

PR #750, opened 2017-07-08
variable of BOOL type with TRUE fix

drillsar:
It is incorrect to compare the variable of BOOL type with TRUE. Reference #746 Fix for V676


drillsar:
Not sure why it failed test..

PR #751, opened 2017-07-08
Multiple fixes (Issue #746)

drillsar:
Fixed V796 Missing break statement
Fixed V555 the expression '32 - (offset + count)'
Fixed V668 The 'fetch_mask' argument is a non-constant bug
Fixed V629 32-bit value fix to the 64-bit type

### DIFF
--- a/src/xenia/gpu/vulkan/texture_cache.cc
+++ b/src/xenia/gpu/vulkan/texture_cache.cc
@@ -1309,7 +1309,7 @@ bool TextureCache::UploadTextureCube(VkCommandBuffer command_buffer,
 }
 
 void TextureCache::HashTextureBindings(
-    XXH64_state_t* hash_state, uint32_t& fetch_mask,
+    XXH64_state_t* hash_state, uint32_t fetch_mask,
     const std::vector<Shader::TextureBinding>& bindings) {
   for (auto& binding : bindings) {
     uint32_t fetch_bit = 1 << binding.fetch_constant;
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_strings.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_strings.cc
@@ -580,6 +580,7 @@ int32_t format_core(PPCContext* ppc_context, FormatData& data, ArgList& args,
           // ANSI_STRING / UNICODE_STRING
           case 'Z': {
             assert_always();
+            break;
           }
 
           default: { assert_always(); }
--- a/src/xenia/gpu/spirv_shader_translator.cc
+++ b/src/xenia/gpu/spirv_shader_translator.cc
@@ -1018,7 +1018,7 @@ spv::Id SpirvShaderTranslator::BitfieldExtract(spv::Id result_type,
   spv::Id base_type = b.getTypeId(base);
 
   // <-- 32 - (offset + count) ------ [bits] -?-
-  if (32 - (offset + count) > 0) {
+  if (32 != offset + count) {
     base = b.createBinOp(spv::Op::OpShiftLeftLogical, base_type, base,
                          b.makeUintConstant(32 - (offset + count)));
   }
--- a/src/xenia/gpu/vulkan/texture_cache.h
+++ b/src/xenia/gpu/vulkan/texture_cache.h
@@ -159,7 +159,7 @@ class TextureCache {
                          VkFence completion_fence, Texture* dest,
                          const TextureInfo& src);
 
-  void HashTextureBindings(XXH64_state_t* hash_state, uint32_t& fetch_mask,
+  void HashTextureBindings(XXH64_state_t* hash_state, uint32_t fetch_mask,
                            const std::vector<Shader::TextureBinding>& bindings);
   bool SetupTextureBindings(
       VkCommandBuffer command_buffer, VkFence completion_fence,
--- a/src/xenia/ui/gl/circular_buffer.cc
+++ b/src/xenia/ui/gl/circular_buffer.cc
@@ -81,7 +81,7 @@ CircularBuffer::Allocation CircularBuffer::Acquire(size_t length) {
 
 bool CircularBuffer::AcquireCached(uint32_t key, size_t length,
                                    Allocation* out_allocation) {
-  uint64_t full_key = key | (length << 32);
+  uint64_t full_key = key | (uint64_t(length) << 32);
   auto it = allocation_cache_.find(full_key);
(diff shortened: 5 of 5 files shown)


iwanttobethebee:
btw idk

drillsar:
@iwanttobethebee you don't know what?

drillsar:
Travis.cl is complaining that I need to format.
I have a question with this line:
The original code
```
XXH64_state_t* hash_state, uint32_t& fetch_mask,
```
you noticed the & with uint32_t instead of with fetch_mask which I thought perhaps a mistype. However, travis.cl says I need to format and when I do it goes back.

I changed the code to this which I feel is correct since now analyzer don't throw this error anymore.
```
XXH64_state_t* hash_state, uint32_t &fetch_mask,
```

iwanttobethebee:
... don't trigger I...I was just... sorry for commenting

El 8 jul. 2017 10:43 p. m., "drillsar" <notifications@github.com> escribió:

@iwanttobethebee <https://github.com/iwanttobethebee> you don't know what?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/benvanik/xenia/pull/751#issuecomment-313898310>, or mute
the thread

maxton:
# Please, _please_ stop opening and closing so many pull requests. Why are you still doing this?

drillsar:
@maxton this is my 3rd request. For some reason travis test wouldn't pass. One of the fixes which I thought was a fix won't let me, because of Travis unless it's a false error.
here is the code:
```
XXH64_state_t* hash_state, uint32_t& fetch_mask,
```
I was figuring it was:
```
XXH64_state_t* hash_state, uint32_t &fetch_mask,
```
It kind of don't matter since it's part of Vulkan and probably better if that's fixed from vulkan project.

resetnow:
this is also a second time a newly created account posts some nonsense to your PR, how accidental is that?

drillsar:
@resetnow I know

src/xenia/base/exception_handler_win.cc
```diff
@@ -40,9 +40,9 @@ LONG CALLBACK ExceptionHandlerCallback(PEXCEPTION_POINTERS ex_info) {
   thread_context.rip = ex_info->ContextRecord->Rip;
   thread_context.eflags = ex_info->ContextRecord->EFlags;
   std::memcpy(thread_context.int_registers, &ex_info->ContextRecord->Rax,
-              sizeof(thread_context.int_registers));
+              sizeof(*thread_context.int_registers));
```
benvanik: This is not correct; the size here should be the full size of the int_registers array, not the size of each element.

drillsar: Ok Fixed. PVS Studio gave it a V512 error. I am guessing that it was a false error. I removed that fix

gibbed: @drillsar Information output from tools such as PVS Studio are meant to be guidance with regards to warnings/errors about code. It cannot perfectly understand every scenario especially in a project such as Xenia.

drillsar: @gibbed got ya. The one V668 Fetch_Mask looks like an error, but I wasn't sure on this either.

paulsapps: I think you can shut up PVS and also make it more clear by copying from  &ex_info->ContextRecord->int_registers instead of Rax?

drillsar: @gibbed @benvanik what about the other fixes?

drillsar:
@gibbed @benvanik what about the other fixes?
I also trying to figure out if this is indeed an error or a false positive.
V569 Truncation of constant value 3735929054. The value range of unsigned char type: [0, 255]. xex_module.cc line 393.

src/xenia/gpu/vulkan/texture_cache.cc
```diff
@@ -1309,7 +1309,7 @@ bool TextureCache::UploadTextureCube(VkCommandBuffer command_buffer,
 }
 
 void TextureCache::HashTextureBindings(
-    XXH64_state_t* hash_state, uint32_t& fetch_mask,
+    XXH64_state_t* hash_state, uint32_t fetch_mask,
```
DrChat: Not a bug, but the reference is no longer necessary.

src/xenia/base/main_win.cc
```diff
@@ -29,7 +29,7 @@ bool has_console_attached_ = true;
 bool has_console_attached() { return has_console_attached_; }
 
 void AttachConsole() {
-  bool has_console = ::AttachConsole(ATTACH_PARENT_PROCESS) == TRUE;
+  bool has_console = ::AttachConsole(ATTACH_PARENT_PROCESS) != FALSE;
```
DrChat: Don't do this - this makes this more confusing to read.

src/xenia/cpu/stack_walker_win.cc
```diff
@@ -209,7 +209,7 @@ class Win32StackWalker : public StackWalker {
            stack_walk_64_(IMAGE_FILE_MACHINE_AMD64, GetCurrentProcess(),
                           thread_handle, &stack_frame, &thread_context, nullptr,
                           XSymFunctionTableAccess64, XSymGetModuleBase64,
-                          nullptr) == TRUE) {
+                          nullptr) != FALSE) {
```
DrChat: Ditto.

src/xenia/kernel/xboxkrnl/xboxkrnl_strings.cc
```diff
@@ -580,6 +580,7 @@ int32_t format_core(PPCContext* ppc_context, FormatData& data, ArgList& args,
           // ANSI_STRING / UNICODE_STRING
           case 'Z': {
             assert_always();
+            break;
```
DrChat: This is a good catch!

src/xenia/base/main_win.cc
drillsar: OK so keep it True?

DrChat: Yes - keep it as it was.

gibbed: To elaborate: it's highly unlikely we'll ever get junk in this case. If we do, there's going to be something drastically wrong with things anyway that this check failing to work as intended is the least of worries.

drillsar:
Ok removed that change.

drillsar:
Everything look ok?

PR #757, opened 2017-07-19
Fixed Vulkan's dark color issue.

AllanCat:
It turns out the wrong color format for DXT caused this issue.
Before:
vulkan_shadow
After:
lightfix

The water reflection is still incorrect, will look into that later.


gibbed:
Good catch!

PR #758, opened 2017-07-19
Fixed white edge on some textures(text display).

AllanCat:
Set borderColor to transparent to get rid of white edges on some textures(mainly text).

Before:
fontfix
fontfix5

After:
fontfix2
fontfix3


VHGaming:
how do you guys get test builds

Margen67:
@VHGaming https://ci.appveyor.com/project/benvanik/xenia Use the "Release" configuration

DrChat:
@AllanCat Nice catch - and it also looks like games can actually control the border color.

If you want to look into that - see [this](https://github.com/benvanik/xenia/blob/master/src/xenia/gpu/xenos.h#L84) enum, which is referenced [here](https://github.com/benvanik/xenia/blob/master/src/xenia/gpu/xenos.h#L460). For now, I'll go ahead and pull this PR as a temporary fix.

PR #788, opened 2017-09-14
Update memory.cc

wolfokami321:
This may prevent tenchu Z from crashing after the start menu


gibbed:
Sorry, but this will not be accepted. Page zero writes are *never* valid, if a game is currently doing it, it has other issues.

If you wish to change this behavior from the default, use the command-line argument `--protect_zero=false` to set it to false.

PR #789, opened 2017-09-19
Rendering and emulation fixes for Geometry Wars RE2

sabretooth:
A couple of minor changes (2e4413e and d82f37a) to get the game booting, and then rendering fixes. 

Commit b34350e9 (TextureCache: Remove distinction between resolve textures and regular textures) broke all of the render-to-texture effects in this game (and presumably others), so reimplement that feature in e1d576d.

This game uses a depth-only render pass with a fullscreen quad at z=0 to zero out relevant EDRAM tiles prior to color rendering. The EDRAM tile store isn't emulated well enough for this to work yet, but a quick hack in a2ab3b3 gets the game rendering with correct clearing.

After applying these changes, the game boots and renders (nearly) correctly with all visual effects applied. There are still a couple of issues: some (mainly UI) elements are the wrong color, there are missing audio channels in the menu background music, and the whole audio mix runs about 500ms behind gameplay.

### DIFF
--- a/src/xenia/gpu/vulkan/vulkan_command_processor.cc
+++ b/src/xenia/gpu/vulkan/vulkan_command_processor.cc
@@ -678,6 +678,41 @@ bool VulkanCommandProcessor::IssueDraw(PrimitiveType primitive_type,
                      vertex_offset, first_instance);
   }
 
+  // Track candidate "real" sizes for resolve operations. We don't have enough
+  // information during resolve to know how much of the viewport was actually
+  // rendered to (just the rounded-up tile size), so keep track of the scissor
+  // rects used during rendering to make an educated guess.
+  {
+    uint32_t window_scissor_tl = regs[XE_GPU_REG_PA_SC_WINDOW_SCISSOR_TL].u32;
+    uint32_t window_scissor_br = regs[XE_GPU_REG_PA_SC_WINDOW_SCISSOR_BR].u32;
+    MsaaSamples surface_msaa =
+        MsaaSamples((regs[XE_GPU_REG_RB_SURFACE_INFO].u32 >> 16) & 0x3);
+
+    uint32_t window_x1 = window_scissor_tl & 0x7FFF;
+    uint32_t window_y1 = (window_scissor_tl >> 16) & 0x7FFF;
+    uint32_t window_x2 = window_scissor_br & 0x7FFF;
+    uint32_t window_y2 = (window_scissor_br >> 16) & 0x7FFF;
+    switch (surface_msaa) {
+      case MsaaSamples::k1X:
+        break;
+      case MsaaSamples::k4X:
+        window_x1 *= 2;
+        window_x2 *= 2;
+      // fallthrough
+      case MsaaSamples::k2X:
+        window_y1 *= 2;
+        window_y2 *= 2;
+        break;
+    }
+
+    if (window_x1 == 0 && window_y1 == 0) {
+      // the scissor extent will probably only be relevant if we're covering the
+      // full frame
+      VkExtent2D candidate = {window_x2, window_y2};
+      resolve_sizes_.push_back(candidate);
+    }
+  }
+
   return true;
 }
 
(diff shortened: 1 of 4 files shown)


AllanCat:
Very nice work :D You can set kMaximumQueuedFrames lower in audio_system.h to get better audio latency(32 in my case).

DrChat:
These are some exciting changes - but oh man, resolve textures are a complete hack (they're no different from regular textures; afaik they were implemented because we thought we didn't have enough info in IssueCopy to completely fill a TextureInfo struct).
I'd _really_ prefer not going back to resolve textures, instead could you investigate what exactly is causing the issue as it is now?

Also, great find on the depth clears! I actually think RDR might be doing the same thing. Could you split that commit from this PR and submit it separately?

Same thing with the packet disasm and kernel fixes, but you can group those together.
Nice work!

sabretooth:
The only thing preventing this game from using the textures as created in IssueCopy is that the dimensions are incorrect. I'll take another look at the code there and see if I can figure out how to fix the dimensions at create time; this code path would then not be required if the game is reading a full texture created during IssueCopy. That wouldn't help if a game was reading a subregion of an IssueCopy-created texture (and there are some comments in there about RDR doing that?), though; I think you'd still need this code (or something like it) to copy out the relevant region.

sabretooth:
Opened a new pull request #794 with just the kernel fixes.

sabretooth:
Replaced the resolve texture copying code with some logic to try and figure out the correct size during IssueCopy(). It's pretty simple, but it covers all of the cases in this game at least.

src/xenia/gpu/vulkan/vulkan_command_processor.cc
```diff
@@ -995,6 +1030,34 @@ bool VulkanCommandProcessor::IssueCopy() {
   int32_t dest_max_y = int32_t(
       (std::max(std::max(dest_points[1], dest_points[3]), dest_points[5])));
 
+  // Search recently used window scissor sizes for the actual dimensions of
+  // this copy (instead of the rounded-out tile size we currently have)
```
DrChat: Hmm... In IssueCopy, we're given the exact dimensions of the texture in one of the copy registers. Those dimensions are rounded up to the nearest tile size when fetching from the render cache.

Perhaps there is an issue with finding the correct image in the render cache?

sabretooth: The RB_COPY_* registers don't contain the true image width, just the surface pitch (which is too large). The RB_COPY_DEST_PITCH register does contain the true height, so I should use that. Still, I've not found a register that contains the true width.

My understanding is that the copy command doesn't actually need the true (pixel-accurate) width because the GPU stores the texture in a padded/aligned format anyway, rounding up to the next block size, and that the cropping from surface pitch to actual width is handled at texture fetch time.

DrChat: So the EDRAM source information is stored in register `RB_SURFACE_INFO`, and the pitch of the image (in EDRAM tiles) is specified here.

The logical width and height of the destination texture is stored in RB_COPY_DEST_PITCH in pixels. This cannot be incorrect otherwise it would screw up later reads.

Is the game resolving to a texture of a certain size then aliasing that texture with a different size?

sabretooth:
Updated the texture size matching to use the known-accurate RB_COPY_DEST_PITCH register copy_dest_height field.

src/xenia/gpu/vulkan/vulkan_command_processor.cc
sabretooth: I guess you could say that, yeah. It does some typical postprocessing blur operations where you sample a texture and render to an RT that's half the size of the input. It ends up needing some oddly-sized textures in that process.

Initial render: 150 x 100, resolved to 152x104
Pass 2: 75x50, resolved as 80x56
Pass 3: 37x25, resolved as 40x32
Pass 4: 18x12, resolved as 24x16
Pass 5: 9x6, resolved as 16x8
Pass 6: 4x3, resolved as 8x8
Pass 7: 2x1, resolved as 8x8

sabretooth:
Ran through some more tests and this appears to break games that render using predicated tiling. I'll update this PR when I've fixed that.

DrChat:
I'm not gonna pull these changes quite yet - let's figure out the size differences and what we can do about that specifically.

src/xenia/gpu/vulkan/vulkan_command_processor.cc
DrChat: Okay cool, your data is very useful - thanks.

Can you also see what texture size the game requests when attempting to sample the resolved textures?

sabretooth: The first column of that data is both the window scissor size when rendering and the requested texture size when sampling -- they always match.

sabretooth:
Updated the patch to fix predicated tiling. My test game for that feature was Ice Age 3; no render regressions there now.

As a bonus, I found that with this patch applied Zoids Infinity EX NEO no longer immediately crashes inside the graphics driver and makes it to the title screen. Has some threading issues preventing it from being playable, though.

sabretooth:
I'm going to withdraw this PR and try and come up with a better solution for cropping the resolve textures. While the current patch works with my limited test cases, it seems like it's going to become a pain in the future.

PR #794, opened 2017-09-24
GW:RE2 game fixes (kernel/libraries only)

Fraxul:
Separated from the graphics fixes in #789 that need rework.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
@@ -418,10 +418,14 @@ DECLARE_XBOXKRNL_EXPORT(RtlTryEnterCriticalSection,
                         ExportTag::kImplemented | ExportTag::kHighFrequency);
 
 void RtlLeaveCriticalSection(pointer_t<X_RTL_CRITICAL_SECTION> cs) {
+  if (cs->recursion_count == 0) {
+    // This CS is not locked. API misuse, but not fatal.
+    return;
+  }
+
   assert_true(cs->owning_thread == XThread::GetCurrentThread()->guest_object());
 
   // Drop recursion count - if it isn't zero we still have the lock.
-  assert_true(cs->recursion_count > 0);
   if (--cs->recursion_count != 0) {
     assert_true(cs->recursion_count >= 0);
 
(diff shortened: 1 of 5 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
```diff
@@ -418,7 +418,10 @@ DECLARE_XBOXKRNL_EXPORT(RtlTryEnterCriticalSection,
                         ExportTag::kImplemented | ExportTag::kHighFrequency);
 
 void RtlLeaveCriticalSection(pointer_t<X_RTL_CRITICAL_SECTION> cs) {
-  assert_true(cs->owning_thread == XThread::GetCurrentThread()->guest_object());
+  if (cs->owning_thread != XThread::GetCurrentThread()->guest_object()) {
```
DrChat: You're right - games that do this are totally abusing the API but xboxkrnl will release the critical section regardless of the owner, so this if statement should be axed.

sabretooth: I went and checked in the debugger and the actual issue is that the game is releasing a critical section that's not owned by _anyone_ (owning_thread=0, lock_count=0). I'll change this patch to a check for that, since it would just fail the assertion on line 424.

DrChat: Huh, that's interesting. Doing so natively will cause the recursion count to underflow and wrap around, which would corrupt the critical section.
Is the game misbehaving?

sabretooth: Doesn't seem like it, the game plays fine. I set some conditional breakpoints and afaict nobody ever calls RtlEnterCriticalSection on that CS, so even if it is corrupted then maybe nobody ever noticed.

sabretooth:
changed the RtlLeaveCriticalSection patch to ignore leave on unlocked critical section, instead of ignoring an owner mismatch

sabretooth:
Need me to make any more changes to these patches, or can we close this out?

DrChat:
Hmm.. Needs more investigation. xboxkrnl does not have this branch, and will decrement the recursion count even if it underflows.

sabretooth:
That behaviour should work fine for this game, since RtlEnterCriticalSection is never called on that CS.

DrChat:
Indeed - it may not be a problem for this game, but if a game does some shady under-the-radar BS it's gonna have a bad day.

sabretooth:
You'd think they would've caught basic API usage errors like these in cert...

DrChat:
Indeed you'd think so, but it doesn't look like this one was caught. Or there's an error elsewhere in Xenia that's making the game think the CS is taken.

illusion98:
Latest master now in-game without any additional fixes.
https://github.com/xenia-project/xenia/commit/3238fae75849aac0637d61b087994177dd8ca5a8
 should this be closed?

Margen67:
@Fraxul 👀

TGLTClutch:
Close or fix 🤔

Issue #795, opened 2017-09-25
Games sample from resolve textures with different dimensions


DrChat:
As seen in #789.

Geometry Wars RE2
> Initial render: 150 x 100, resolved to 152x104
> Pass 2: 75x50, resolved as 80x56
> Pass 3: 37x25, resolved as 40x32
> Pass 4: 18x12, resolved as 24x16
> Pass 5: 9x6, resolved as 16x8
> Pass 6: 4x3, resolved as 8x8
> Pass 7: 2x1, resolved as 8x8

Textures are sampled with the dimensions from the first column.

gibbed:
When games are doing this are they asking for a scaled version of that texture, or a sliced portion of the texture? (I'm assuming the former?)

DrChat:
For our purposes let's assume the latter.
I don't think it's possible for games to scale a texture by requesting smaller dimensions, and I know RDR already slices images doing this.

sabretooth:
I think the right way to do this is to generate cropped views of the resolve surface at demand time / when it's attached to the sampler. I'll try and come up with a changeset for method that's less terrible than my previous reimplementation of the resolve texture feature.

DrChat:
I ran into this problem before and I thought of duplicating the texture with the sub-dimensions, and assigning it to the same writewatch as the original (so it's updated alongside the original).

sabretooth:
Took a stab at generating cropped views of textures. Commit is here if you want to look: https://github.com/sabretooth/xenia/commit/a8228a5fc96f414f7716ee9e90abafd34ce2a6a3

Works for GW:RE2 at least!

DrChat:
That actually looks a lot like what I had in mind :)

Open a PR with these changes and I'll review and test them.

sabretooth:
Opened the PR as #800. I updated the patch to address your comments on the previous revision (about the feature flags).

sabretooth:
You get a chance to look at that pull request yet?

DrChat:
@sabretooth I looked at it briefly but haven't had time as of yet to give it a test.
I'll test it soon, but as of now it looks good!

Triang3l:
Supported by the Direct3D 12 backend with generalized physical memory that resolve does tiling directly to.

PR #800, opened 2017-10-02
Texture region splitting

sabretooth:
Implementing a potential fix for #795 by generating cropped views of textures on-demand.

### DIFF
--- a/src/xenia/gpu/vulkan/render_cache.cc
+++ b/src/xenia/gpu/vulkan/render_cache.cc
@@ -592,6 +592,7 @@ bool RenderCache::dirty() const {
   dirty |= cur_regs.rb_color2_info.value != regs[XE_GPU_REG_RB_COLOR2_INFO].u32;
   dirty |= cur_regs.rb_color3_info.value != regs[XE_GPU_REG_RB_COLOR3_INFO].u32;
   dirty |= cur_regs.rb_depth_info.value != regs[XE_GPU_REG_RB_DEPTH_INFO].u32;
+  dirty |= cur_regs.rb_color_mask != regs[XE_GPU_REG_RB_COLOR_MASK].u32;
   dirty |= cur_regs.pa_sc_window_scissor_tl !=
            regs[XE_GPU_REG_PA_SC_WINDOW_SCISSOR_TL].u32;
   dirty |= cur_regs.pa_sc_window_scissor_br !=
@@ -610,6 +611,7 @@ const RenderState* RenderCache::BeginRenderPass(VkCommandBuffer command_buffer,
   current_command_buffer_ = command_buffer;
 
   // Lookup or construct a render pass compatible with our current state.
+  auto previous_render_pass = current_state_.render_pass;
   auto config = &current_state_.config;
   CachedRenderPass* render_pass = nullptr;
   CachedFramebuffer* framebuffer = nullptr;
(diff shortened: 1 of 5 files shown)


src/xenia/gpu/vulkan/vulkan_command_processor.cc
```diff
@@ -854,7 +856,7 @@ bool VulkanCommandProcessor::PopulateSamplers(VkCommandBuffer command_buffer,
 
   std::vector<xe::gpu::Shader::TextureBinding> dummy_bindings;
   auto descriptor_set = texture_cache_->PrepareTextureSet(
-      setup_buffer, current_batch_fence_, vertex_shader->texture_bindings(),
+      command_buffer, current_batch_fence_, vertex_shader->texture_bindings(),
```
DrChat: Q: Why did you change this to command_buffer? AFAIK uploads cannot be completed inside of a render pass.

src/xenia/gpu/vulkan/render_cache.cc
```diff
@@ -713,6 +716,28 @@ const RenderState* RenderCache::BeginRenderPass(VkCommandBuffer command_buffer,
   vkCmdBeginRenderPass(command_buffer, &render_pass_begin_info,
                        VK_SUBPASS_CONTENTS_INLINE);
 
+  // XXX HACK: clear the color target if the previous render pass was a
+  // depth-stencil pass targeting the same EDRAM tile
+  // (some games clear color targets with a depth pass, since it's slightly
+  // faster)
+  if (previous_render_pass && previous_render_pass->config.depth_stencil.used &&
```
DrChat: This may need to be axed for now. It's causing a regression in banjo-kazooie for me as of now.

DrChat:
Good work! Just need a couple of things addressed for now.

src/xenia/gpu/vulkan/vulkan_command_processor.cc
sabretooth: The setup_buffer is built through the entire frame and executed in its entirety before the command_buffer. The blits that we add here for carving up the resolve textures depend on the render passes completed in command_buffer; putting them in setup_buffer would have the blits execute before the render targets they read from are written.

src/xenia/gpu/vulkan/render_cache.cc
sabretooth: Good to know, I'll look at that game and see if I can fix the regression.

src/xenia/gpu/vulkan/texture_cache.cc
```diff
@@ -325,17 +350,21 @@ bool TextureCache::FreeTexture(Texture* texture) {
     vkDestroyFramebuffer(*device_, texture->framebuffer, nullptr);
   }
 
-  for (auto it = texture->views.begin(); it != texture->views.end();) {
-    vkDestroyImageView(*device_, (*it)->view, nullptr);
-    it = texture->views.erase(it);
+  for (auto region_it = texture->regions.begin();
```
DrChat: You don't appear to be freeing the base_region here

src/xenia/gpu/vulkan/vulkan_command_processor.cc
DrChat: Okay, that's fair but you'll need to end the render pass when doing an upload/blit (as per vulkan spec).

I'd say pass both the setup buffer and command buffer, do uploads from memory in the setup buffer and uploads from other textures in the command buffer.

src/xenia/gpu/vulkan/texture_cache.cc
DrChat: Also - there's a regression somewhere causing a texture to be freed twice in this branch, but I haven't tracked it down quite yet.

src/xenia/gpu/vulkan/texture_cache.cc
```diff
@@ -404,13 +433,23 @@ TextureCache::Texture* TextureCache::DemandResolveTexture(
       },
       this, texture);
 
+  // Invalidate any textures that share this memory location
+  for (auto it = textures_.begin(); it != textures_.end(); ++it) {
```
DrChat: This block of code is causing a texture to be freed twice - it appears that an access watch fires and then this block of code runs over the texture as well.

DrChat:
Update: Keep an eye on [the texsplit branch](https://github.com/benvanik/xenia/tree/texsplit) - I'm making changes there.

DrChat:
Continued in new PR.

Issue #356, opened 2015-07-19
Add abstracted ABI handling in CPU.


benvanik:
Right now everything is fairly Win64-specific. Most of the JIT'ed code can remain the same, but the `X64ThunkEmitter` functions in `x64_backend.cc` will need to change to handle proper guest-to-host and host-to-guest transitions. This may then be useful to describe on `xe::cpu::backend::MachineInfo`, as then the CodeCache and other types can pull unwind info/etc from that without having to have magic constants now like `kUnwindInfoSize`.

PR #803, opened 2017-10-24
[Linux] xenia-cpu-ppc-tests

### LINKED ISSUE #670: xenia-cpu-ppc-tests Linux Support
**Goal**: Get the PowerPC tests to succeed on Linux hosts.

As of now, a bunch of posix functionality is stubbed out in xenia-base. You can fill in functionality as it's needed for the tests.
Code for posix functionality can be found in src/xenia/base/*_posix.cc
Code for the tester lives in src/xenia/cpu/ppc/testing

uytvbn:
Hi,

This pull request fixes #670 (and also partially #356). I've tried to change as little as possible in Windows code.

> $ ./build/bin/Linux/Debug/xenia-cpu-ppc-tests --log_file stdout
> ...
>i> 0000293A Total tests: 1457
>i> 0000293A Passed: 1457
>i> 0000293A Failed: 0

Comments:
- src/xenia/base/exception_handler_posix.cc - straightforward implementation based almost entirely on Windows version,
- src/xenia/base/math.h - off-by-one error, ffs contrary to _BitScanForward returns one-based index,
- src/xenia/base/memory_posix.cc - porting this exactly is probably not possible as there are significant differences in virtual memory handling between Windows and Linux, tried to follow the behavior of Windows implementation as closely as possible,
- src/xenia/base/string_buffer.cc - fixed non-standard behavior that happens to work with MSVC,
- src/xenia/cpu/backend/x64/x64_backend.cc - conform to Linux calling convention,
- src/xenia/cpu/backend/x64/x64_sequences.cc - the only change here that affects (code-wise) MSVC, it forces Linux to pass XMM values by pointer which is expected by codegen (default calling convention on Linux uses XMM registers), I've checked and it doesn't affect code generated by Visual Studio 2017, could probably hide it behind a typedef if you want to feel extra safe - though you might end up with different types for input and output,
- src/xenia/cpu/hir/value.cc - fixes Value::MulHi for 64 bit values, returned low not high qword,
- xenia-build - changes binutils path to use the one you get when building in third-party/binutils.

Thanks.


uytvbn:
Reformatted and rebased. It looks like I should have really used clang-3.8 before.

DrChat:
Thanks for the PR! This looks pretty exciting - give me some time and I'll review it later.

uytvbn:
Thank you for the review. I'm sorry but I won't be able to continue working on this. Unfortunately I'm swamped right now and it doesn't look like it will change anytime soon. Please just take whatever, if anything, you find useful.

PR #805, opened 2017-11-01
Missing break statement

drillsar:
Missing break statement in xboxkrnl_strings.cc


maxton:
`assert_always()` calls `assert(false)`, which terminates the program, so the break statement doesn't matter here.

DrChat:
@maxton You're correct in the case of Debug builds - in Release builds, all asserts are removed.

I'll go ahead and pull this - thanks @drillsar!

PR #806, opened 2017-11-05
Fixed V555 the expression '32 - (offset + count)'

drillsar:


src/xenia/gpu/spirv_shader_translator.cc
```diff
@@ -1015,7 +1015,7 @@ spv::Id SpirvShaderTranslator::BitfieldExtract(spv::Id result_type,
   spv::Id base_type = b.getTypeId(base);
 
   // <-- 32 - (offset + count) ------ [bits] -?-
-  if (32 - (offset + count) > 0) {
+  if (32 != offset + count) {
```
x1nixmzeng: The previous code is checking this value is within a valid range (0-31). You're effectively checking one edge case - what does this solve?

maxton: This should be `if (32 > (offset + count))`

PR #808, opened 2017-11-05
Reference no longer necessary

drillsar:


x1nixmzeng:
Marge the .h and .cc changes into one pull request. The CI be failing on both.

drillsar:
ok got ya. I will do

PR #813, opened 2017-12-04
Recognize OpenGL 4.60

joaozin23:
A suitably easy fix is to change line 289 of gl_context.cc to:

if ((glsl_version.find("4.5") != 0) && (glsl_version.find("4.6") != 0))


raynebc:
My original recommended change was better, this commit would prevent any version of OpenGL other than 4.6x from being accepted.

PR #814, opened 2017-12-04
Recognize OpenGL 4.60

joaozin23:
..\src\xenia\ui\gl\gl_context.cc(289): error C2059: syntax error: ')' [C:\dev\xenia\build\xenia-ui-gl.vcxproj]


shadowbrony33:
That's better

PR #816, opened 2017-12-04
Detect OpenGL 4.60

Kitessencial:
This change allows Xenia to detect if the system has OpenGL 4.6

### DIFF
--- a/src/xenia/ui/gl/gl_context.cc
+++ b/src/xenia/ui/gl/gl_context.cc
@@ -286,7 +286,7 @@ void GLContext::AssertExtensionsPresent() {
   auto glsl_version_raw =
       reinterpret_cast<const char*>(glGetString(GL_SHADING_LANGUAGE_VERSION));
   std::string glsl_version(glsl_version_raw);
-  if (glsl_version.find("4.50") != 0 && glsl_version.find("4.60") != 0) {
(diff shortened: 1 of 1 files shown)


joaozin23:
Not ah that problem aki n

raynebc:
Consider using my previous suggestion of:
if ((glsl_version.find("4.5") != 0) && (glsl_version.find("4.6") != 0))

Which should work with OpenGL 4.5x and 4.6x instead of just 4.50 and 4.60

Kitessencial:
Indeed, it is a better approach. I didn't know they were versions between 4.50 and 4.60.

x1nixmzeng:
Tiny nitpick, but a find result of 0 is a valid index. `std::string::npos` is defined as `-1` when the string is not found.

maxton:
@x1nixmzeng - I think the code is correct as-is; the version number will always be at the start of the string so checking for `!= 0` is equivalent to checking for `== -1` in this case

Kitessencial:
True, the code should work fine as is, the current release uses != 0.
But, for clarity, I think it's better to adopt the == -1, maybe a new OpenGL version is added in the future that changes how the string starts, so it would be easier to add detection for it as well.

gibbed:
You should be comparing the index against std::string::npos, not -1 (even if npos is technically -1).

Kitessencial:
You mean like this?

PR #817, opened 2017-12-04
(Possible ?)Crash workaround

Kitessencial:
This change ignores a crash that happened with me.
Fix for a crash that happens after the intro cutscene in Sonic: The Hedgehog. Playing with OpenGL
It may fix some other crashes as well.

Can someone confirm whether the changed lines are needed or they can just cause a crash? Maybe we can develop a way to bypass these crashes if we know when they happen.

### DIFF
--- a/src/xenia/cpu/mmio_handler.cc
+++ b/src/xenia/cpu/mmio_handler.cc
@@ -200,13 +200,13 @@ void MMIOHandler::CancelAccessWatch(uintptr_t watch_handle) {
 
   // Remove from table.
   auto it = std::find(access_watches_.begin(), access_watches_.end(), entry);
-  assert_false(it == access_watches_.end());
+  //assert_false(it == access_watches_.end());
 
   if (it != access_watches_.end()) {
     access_watches_.erase(it);
   }
 
-  delete entry;
+  //delete entry;
 }
 
 void MMIOHandler::InvalidateRange(uint32_t physical_address, size_t length) {


gibbed:
Disabling these sanity checks is only hiding the problem and leaking memory.

gibbed:
(didn't mean to close, sorry)

Kitessencial:
I know it's not good practice, but at least I got past the cutscene. By the way, I haven't noticed any memory leakage while using this change. Process memory stays almost constant and the game somehow works. Still heavy graphical glitches and only half the screen gets 3d rendered, but better than before for me.
As far as I know, there were very few (maybe even only one) commands that triggered the breakpoint, so it worked. Wonder how it would be for a game that throws would execute many of those.

Kitessencial:
More bad practice: 
Commenting lines 1553, 1555, 1571, 1573, 1575 and 1577 of file gl4_command_processor.cc allows Dante's Inferno to go past the second cutscene (after the first mission) without crashing.
First mission plays in black and white with OpenGL, and with Vulkan displays a black screen with occasional blue flashes. Seems functional in Vulkan as well, just with no graphics.
Second mission with OpenGL seems to start in color but gradually reverts to B/W during some seconds. Testing other missions now.
Still no apparent memory leakage to be worried about, xenia.exe sitting with about 1GB of RAM.

Again, I know this is not a fix, just a workaround to maybe allow for better testing and debugging.

DrChat:
Sorry, we can't accept workarounds in this style :(

This is a good starting point if you want to look into the issue - it looks like an access watch is deleted twice if removing those lines "fixes" the issue.

Kitessencial:
Thank you for the info. I will look into it in more detail, but first there are some things I need to learn.

PR #1069, opened 2018-01-08
UI/Vulkan: Fixed graphics context recreation on HiDPI systems

parkerlamb:
This prevents the graphics context from being constantly recreated on HiDPI systems. The window resize check was not using the scaled window dimensions so the context was recreated on every frame as the dimensions were always different.

### DIFF
--- a/src/xenia/ui/vulkan/vulkan_context.cc
+++ b/src/xenia/ui/vulkan/vulkan_context.cc
@@ -150,8 +150,8 @@ void VulkanContext::BeginSwap() {
   // If it has been, we'll need to reinitialize the swap chain before we
   // start touching it.
   if (target_window_) {
-    if (target_window_->width() != swap_chain_->surface_width() ||
-        target_window_->height() != swap_chain_->surface_height()) {
+    if (target_window_->scaled_width() != swap_chain_->surface_width() ||
+        target_window_->scaled_height() != swap_chain_->surface_height()) {
       // Resized!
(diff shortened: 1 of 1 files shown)


DrChat:
Neat! Have you verified that it works OK with normal DPI systems still?

parkerlamb:
No, unfortunately I don't have a standard DPI display to test with.

However, the DPI scaling factor is calculated here:
https://github.com/benvanik/xenia/blob/d512d4127fa12a710e5d1d6d8067fe050b3445e7/src/xenia/ui/window.h#L69

and the actual DPI is gotten here:
https://github.com/benvanik/xenia/blob/d512d4127fa12a710e5d1d6d8067fe050b3445e7/src/xenia/ui/window_win.cc#L305-L312

The DPI value returned by Windows is dependent on the monitor because Xenia declares itself `PROCESS_PER_MONITOR_DPI_AWARE` elsewhere. So anyone with a standard DPI display should end up with a scaling factor of 1.

It may be a good idea to stash the DPI and only get it when it changes to reduce the number of API calls. Xenia seems to listen for the `WM_DPICHANGED` event but it doesn't look like it's actually used (as far as I can tell). I can modify the PR to do this if you like.

DrChat:
I don't think the API is called often enough to benefit from caching (normally, anyways - this bug is an exception :)

DrChat:
Verified no regressions on my standard DPI-scaled screen. Merged!

PR #1073, opened 2018-01-12
Reference no longer necessary

drillsar:


DrChat:
Actually, this PR points out a minor bug with texture hashing. Will fix in a bit.

DrChat:
Fixed in b38b9e43ea1b52e151864e7445144134bc7996f9.

PR #1074, opened 2018-01-12
Fixed V555 the expression

drillsar:


DrChat:
I can understand why it's recommending this change, but I think it'll make the code a little bit more confusing - so I'm gonna drop this. Appreciate the changes!

PR #1076, opened 2018-01-14
Linux: Fix vulkan init and GTK abstraction to run window-vulkan-demo

dougvj:
- Fix GTK abstractions on linux
- Fix Vulkan init support on linux
- Miscellaneous fixes for linux build/support
- Abstract the keyboard events so that nothing relies on hardcoded windows key codes. 

xenia-ui-window-vulkan-demo compiles and runs. 

xenia-app compiles, runs briefly and crashes with an assert as there is still missing threading functionality, however the GTK window is created correctly.

xenia-ui-window-vulkan-demo

### DIFF
--- a/src/xenia/ui/window_win.cc
+++ b/src/xenia/ui/window_win.cc
@@ -707,7 +707,7 @@ bool Win32Window::HandleMouse(UINT message, WPARAM wParam, LPARAM lParam) {
 
 bool Win32Window::HandleKeyboard(UINT message, WPARAM wParam, LPARAM lParam) {
   auto e = KeyEvent(
-      this, VirtualKey(wParam), lParam & 0xFFFF0000, !!(lParam & 0x2),
+      this, VirtualKey(wParam), lParam & 0xFFFF, !!(lParam & 0x400000000),
       !!(GetKeyState(VK_SHIFT) & 0x80), !!(GetKeyState(VK_CONTROL) & 0x80),
       !!(GetKeyState(VK_MENU) & 0x80), !!(GetKeyState(VK_LWIN) & 0x80));
   switch (message) {
--- a/src/xenia/ui/window_win.cc
+++ b/src/xenia/ui/window_win.cc
@@ -447,6 +447,7 @@ void Win32Window::Close() {
   closing_ = true;
   OnClose();
   DestroyWindow(hwnd_);
+  hwnd_ = 0;
 }
 
 void Win32Window::OnMainMenuChange() {
--- a/premake5.lua
+++ b/premake5.lua
@@ -77,6 +77,11 @@ filter({"configurations:Debug", "platforms:Linux"})
     "_GLIBCXX_DEBUG",   -- make dbg symbols work on some distros
   })
 
+filter({"configurations:Debug", "platforms:Linux"})
+  buildoptions({
+    "-g"
+  })
+
 filter("configurations:Release")
   runtime("Release")
   defines({
--- a/src/xenia/app/premake5.lua
+++ b/src/xenia/app/premake5.lua
@@ -73,7 +73,6 @@ project("xenia-app")
       "X11",
       "xcb",
       "X11-xcb",
-      "vulkan",
       "SDL2",
     })
 
--- a/src/xenia/base/main_posix.cc
+++ b/src/xenia/base/main_posix.cc
@@ -9,6 +9,8 @@
 
 #include <stdio.h>
 #include <unistd.h>
+#include "build/version.h"
+#include "xenia/base/main.h"
 
 #include "xenia/base/cvar.h"
 #include "xenia/base/main.h"
@@ -40,6 +42,9 @@ extern "C" int main(int argc, char** argv) {
 
   // Initialize logging. Needs parsed FLAGS.
   xe::InitializeLogging(entry_info.name);
+
+  // Print version info.
+  XELOGI("Build: " XE_BUILD_BRANCH " / " XE_BUILD_COMMIT " on " XE_BUILD_DATE);
 
   // Call app-provided entry point.
   int result = entry_info.entry_point(args);
--- a/src/xenia/cpu/ppc/testing/premake5.lua
+++ b/src/xenia/cpu/ppc/testing/premake5.lua
@@ -13,6 +13,7 @@ project("xenia-cpu-ppc-tests")
     "xenia-core",
     "xenia-cpu-backend-x64",
     "xenia-cpu",
+    "xenia-ui",
     "xenia-base",
   })
   files({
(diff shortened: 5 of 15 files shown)


dougvj:
The tests were unhappy with some kind of DNS faliure on APT so I closed and opened it again to trigger a rerun

src/xenia/ui/window_win.cc
```diff
@@ -680,9 +680,200 @@ bool Win32Window::HandleMouse(UINT message, WPARAM wParam, LPARAM lParam) {
...
+    case '8':
+      return KeyEvent::Key::k8;
+    case '9':
+      return KeyEvent::Key::k9;
+    case '0':
+      return KeyEvent::Key::k0;
+    case VK_OEM_MINUS:
+      return KeyEvent::Key::kMinus;
+    case VK_OEM_PLUS:
+      return KeyEvent::Key::kEquals;
```
x1nixmzeng: Is this correct?

x1nixmzeng:
Looks good to me! You even made some useful fixes for both platforms :+1:

src/xenia/ui/window_win.cc
dougvj: Good question. That one confused me too. 

I think it is: 

https://www.gamedev.net/forums/topic/293214-virtual-keycode-hell---or-where-is-vk_equals/

x1nixmzeng:
Any comments on this PR @DrChat /@gibbed ?

DrChat:
@dougvj The changes look great at a glance! I'll review this in-depth as I get time.

src/xenia/base/clock_posix.cc
```diff
@@ -15,17 +15,17 @@
...
 
 uint64_t Clock::host_tick_frequency() {
-  timespec res;
-  clock_getres(CLOCK_MONOTONIC_RAW, &res);
-
-  return uint64_t(res.tv_sec) + uint64_t(res.tv_nsec) * 1000000000ull;
+  // Even if our resolution is less, CLOCK_MONOTONIC_RAW will always return
+  // ns values. Most of the time the resolution is pretty close to 1ns anyway
+  // due to the presence of hr timers, so just pass along ns ticks
+  return 1000000000ull;
```
DrChat: Even though I can understand your reasoning - is there a specific reason you decided to hardcode this value rather than use `clock_getres`?

src/xenia/ui/vulkan/vulkan_util.cc
```diff
@@ -28,7 +28,7 @@ Version Version::Parse(uint32_t value) {
   Version version;
   version.major = VK_VERSION_MAJOR(value);
   version.minor = VK_VERSION_MINOR(value);
-  version.patch = VK_VERSION_PATCH(value);
+  version.patch = VK_VERSION_PATCH(VK_HEADER_VERSION);
```
DrChat: This doesn't look quite right - this should still be parsing the input value.

src/xenia/base/clock_posix.cc
dougvj: Good question. So the clock_getres always returns 1ns in Linux even if the clock doesn't actually support that resolution, so I think my thoughts at the time were why bother? But then again why did I bother to change it in the first place? I don't remember... I will revert this.

src/xenia/ui/vulkan/vulkan_util.cc
dougvj: I don't remember why I did this but didn't mean to commit it. Thanks for the catch.

src/xenia/base/clock_posix.cc
dougvj: I noticed that the original frequency calculation was technically wrong even though it produced the right value, so I fixed that.

jcash85:
i have a issue with game not playing it will load up all the way through title screen then when it comes to game play it crashes dragon ball raging blast 1 and 2 and dragon ball ultimate tenkaichi worked two days ago

src/xenia/ui/imgui_drawer.cc
```diff
@@ -117,23 +117,27 @@ void ImGuiDrawer::Initialize() {
...
+      static_cast<int>(xe::ui::KeyEvent::Key::kDown);
+  io.KeyMap[ImGuiKey_Home] = static_cast<int>(xe::ui::KeyEvent::Key::kHome);
+  io.KeyMap[ImGuiKey_End] = static_cast<int>(xe::ui::KeyEvent::Key::kEnd);
+  io.KeyMap[ImGuiKey_Delete] = static_cast<int>(xe::ui::KeyEvent::Key::kDelete);
+  io.KeyMap[ImGuiKey_Backspace] =
+      static_cast<int>(xe::ui::KeyEvent::Key::kBackspace);
+  io.KeyMap[ImGuiKey_Enter] = static_cast<int>(xe::ui::KeyEvent::Key::kEnter);
+  io.KeyMap[ImGuiKey_Escape] = static_cast<int>(xe::ui::KeyEvent::Key::kEsc);
+  io.KeyMap[ImGuiKey_A] = static_cast<int>(xe::ui::KeyEvent::Key::kA);
+  io.KeyMap[ImGuiKey_C] = static_cast<int>(xe::ui::KeyEvent::Key::kV);
```
ananace: I think this one is supposed to be `xe::ui::KeyEvent::Key::kC`

sl1pkn07:
seems needs rebase

greetings

Margen67:
@sl1pkn07 We're aware of this. Can you stop spamming PRs with this?

dougvj:
I will try to find time in the coming month to spend time on a rebase

src/xenia/ui/ui_event.h
```diff
@@ -40,11 +40,107 @@ class FileDropEvent : public UIEvent {
 
 class KeyEvent : public UIEvent {
  public:
-  KeyEvent(Window* target, int key_code, int repeat_count, bool prev_state,
-           bool modifier_shift_pressed, bool modifier_ctrl_pressed,
-           bool modifier_alt_pressed, bool modifier_super_pressed)
+  // TODO(dougvj) The list of keycodes is probably incomplete but may be
+  // sufficient. Maybe there is a better way?
+  enum class Key {
```
Triang3l, line 45: It's preferable that we normalize keys to Windows virtual key codes, as those are sent to the game via XamInputGetKeystroke. This would also allow us to skip most of the translation for Windows host, maybe except for some fixups if needed to match the Xbox 360 controller keyboard better.

Triang3l (changes requested):

Triang3l:
`ui::KeyEvent::Key` superseded by `ui::VirtualKey` in https://github.com/xenia-project/xenia/commit/1cf12ec70b22c4eda28dea7fd8edf44900b098d6 — allows passing all Windows keys to XamInputGetKeystroke and UI subsystems without changes, but needs translation from GDK keys.

dougvj:
Rebased against master, borrowed commits from !1431. 

Builds are fixed and window runs, crashes when trying to run game.  Cpu tests fail. 

I dropped my GTK key abstraction I had in place, I will work on a mapping to the new `ui:VirtualKey`

PR #1106, opened 2018-03-10
[Vulkan] Update the loader to SDK 1.1.70.0 

KDZL:


DrChat:
@KDZL Please don't create so many pull requests. Update the ones you've already created.

PR #1108, opened 2018-03-10
Memory tests

bwrsandman:
Add base tests for memory.
Running memory tests on Linux revealed inconsistent behaviour of `_mm_load_si128` and `_mm_store_si128` where pointers not aligned with 128 bits address space would cause segmentation fault.
A patch for this issue is included.


bwrsandman:
Tests for `copy_and_swap_16_in_32_unaligned` and `copy_and_swap_16_in_32_aligned` were left out since I don't see what they are supposed to do and therefore cannot create good test cases.
I'd be happy to add some in given a bit more context.
I'm particularly confused about the `reinterpret_cast<const uint64_t*>` part. Intuition tells me it should be this instead:
```
  auto dest = reinterpret_cast<uint32_t*>(dest_ptr);
  auto src = reinterpret_cast<const uint16_t*>(src_ptr);
```

src/xenia/base/memory.cc
```diff
@@ -34,15 +34,16 @@ void copy_and_swap_16_aligned(void* dest_ptr, const void* src_ptr,
                    0x04, 0x05, 0x02, 0x03, 0x00, 0x01);
 
   size_t i = 0;
-  size_t unaligned_words = (reinterpret_cast<uintptr_t>(src_ptr) & 0xF) / 2;
+  size_t unaligned_words =
+      (0x10 - (reinterpret_cast<uintptr_t>(src_ptr) & 0xF)) / sizeof(uint16_t);
   for (; unaligned_words > 0 && i < count; unaligned_words--, i++) {
```
DrChat: Nice catch! One thing I wanted to avoid was if `src_ptr` is actually aligned, this will come out with `unaligned_words = 16` and needlessly copy 16 words using the slow route.

src/xenia/base/memory.cc
```diff
@@ -34,15 +34,16 @@ void copy_and_swap_16_aligned(void* dest_ptr, const void* src_ptr,
...
     // Copy up to 16 byte alignment.
     dest[i] = byte_swap(src[i]);
   }
   for (; i + 8 <= count; i += 8) {
     __m128i input = _mm_load_si128(reinterpret_cast<const __m128i*>(&src[i]));
     __m128i output = _mm_shuffle_epi8(input, shufmask);
-    _mm_store_si128(reinterpret_cast<__m128i*>(&dest[i]), output);
+    _mm_storeu_si128(reinterpret_cast<__m128i*>(&dest[i]), output);
```
DrChat: We want to use `_mm_store_si128` on aligned pointers.

DrChat: Yeah - I just realized that as well. If both pointers aren't unaligned in the same manner, the `src` pointer will be aligned and the `dest` pointer just gets unaligned.

Do you have a backtrace for those crashes you were getting on linux? I think it's a good idea to assert alignment on both pointers, since this is an aligned routine after all.

bwrsandman: Adding to that, `src_ptr` and `dst_ptr` should both be aligned (or unaligned by the same amount of bits) or else they only way to avoid a segfault on Linux is to run `_mm_storeu_si128`

bwrsandman: Here's a backtrace: https://gist.github.com/bwrsandman/7d91bb6a1af79366e15161f992858a09
It happens on the call to `_mm_load_si128`

bwrsandman:
More restrictions for `src_ptr` and `dest_ptr` to have similar aligments have been added
I had to change the tests to use these alignments.
The extra bytes slowly copied on aligned data should now be gone.

src/xenia/base/memory.cc
```diff
@@ -27,14 +27,18 @@ void copy_128_aligned(void* dest, const void* src, size_t count) {
 void copy_and_swap_16_aligned(void* dest_ptr, const void* src_ptr,
                               size_t count) {
   assert_zero(reinterpret_cast<uintptr_t>(src_ptr) & 0x1);
+  assert_true((reinterpret_cast<uintptr_t>(src_ptr) & 0xF) ==
```
DrChat: On this - let's just make sure src_ptr and dest_ptr are aligned to a 16-byte boundary (in the aligned routines).
Any cases you run into where this is called with unaligned pointers you can just change to use the `*_unaligned` routines.

It's really confusing to make sure both pointers are unaligned the same way, and _very_ atypical for this to occur :)

src/xenia/base/memory.cc
```diff
@@ -27,14 +27,18 @@ void copy_128_aligned(void* dest, const void* src, size_t count) {
...
   auto dest = reinterpret_cast<uint16_t*>(dest_ptr);
   auto src = reinterpret_cast<const uint16_t*>(src_ptr);
   __m128i shufmask =
       _mm_set_epi8(0x0E, 0x0F, 0x0C, 0x0D, 0x0A, 0x0B, 0x08, 0x09, 0x06, 0x07,
                    0x04, 0x05, 0x02, 0x03, 0x00, 0x01);
 
   size_t i = 0;
-  size_t unaligned_words = (reinterpret_cast<uintptr_t>(src_ptr) & 0xF) / 2;
+  size_t unaligned_words =
+      ((0x10 - (reinterpret_cast<uintptr_t>(src_ptr) & 0xF)) & 0xF) /
```
DrChat: Nice fix! (though won't be needed with the above changes)

src/xenia/base/memory.cc
```diff
@@ -71,14 +75,18 @@ void copy_and_swap_16_unaligned(void* dest_ptr, const void* src_ptr,
 void copy_and_swap_32_aligned(void* dest_ptr, const void* src_ptr,
                               size_t count) {
   assert_zero(reinterpret_cast<uintptr_t>(src_ptr) & 0x3);
+  assert_true((reinterpret_cast<uintptr_t>(src_ptr) & 0xF) ==
```
DrChat: Ditto on the rest of these.

src/xenia/base/testing/memory_test.cc
```diff
@@ -0,0 +1,392 @@
...
+  copy_128_aligned(dest, src, 2);
+  REQUIRE(std::memcmp(dest, src, 256));
+
+  std::memset(dest, 0, sizeof(dest));
+  copy_128_aligned(dest, src + 1, 1);
+  REQUIRE(std::memcmp(dest, src + 1, 128));
+}
+
+TEST_CASE("copy_and_swap_16_aligned", "Copy and Swap") {
+  uint16_t a[8] = {0x1111};
```
DrChat: As for the test cases with src/dest on the stack, I think the `alignas(16)` keyword would work.

bwrsandman:
Now asserting 16bit alignment for each align copy. They all are tested for 16 bit offsets and cause no issues on Linux or Windows.
Removed the unaligned slow copies on aligned copies since the assets no longer allow for this.

DrChat:
Excellent! Have you encountered any cases in Xenia where we're using aligned routines when we shouldn't?

PR #1111, opened 2018-03-10
 Vulkan Atualize o carregador para o SDK 1.1.70.1

KDZL:


gibbed:
@KDZL If you do not stop spamming us with nonsense PR updates, you will be blocked from the project.

gibbed:
I see your thumbs down response.

If these are legitimate PRs then explain what you are doing and why it requires you to spam us with them. Otherwise, you will be blocked.

gibbed:
Closing the PR does not count.

If you submit any further PRs without explaining what you're doing, it will be considered as ignoring my request and you will be blocked.

hlide:
Not only that. Using English in commit message should be mandatory.

PR #1118, opened 2018-03-25
[UI] Set User Language

Parovozik:
Works at English as default, but can choose any language from list (works for multi-lingual games, example Sonic 2006)
xenia_ui_features_snapshot

### DIFF
--- a/src/xenia/app/emulator_window.cc
+++ b/src/xenia/app/emulator_window.cc
@@ -25,6 +25,8 @@
 #include "xenia/ui/imgui_dialog.h"
 #include "xenia/ui/imgui_drawer.h"
 
+uint32_t userlang = 1; // Default as English
+
 namespace xe {
 namespace app {
 
@@ -35,6 +37,7 @@ using xe::ui::UIEvent;
 using xe::ui::FileDropEvent;
 
 const std::wstring kBaseTitle = L"xenia";
+    std::wstring userlangs = L"[UserLang-English]";
 
 EmulatorWindow::EmulatorWindow(Emulator* emulator)
     : emulator_(emulator),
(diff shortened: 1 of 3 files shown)


gibbed:
Nice idea, are games capable of switching languages on the fly through system language change on an actual 360?

Parovozik:
Yes, you can switch even during the game (in the same Sonic 06 it works without any problems)
You can change everything differently, I'm not an expert in C ++ :D
https://www.youtube.com/watch?v=VLFzIYqFN2I (test)

gibbed:
While Sonic 2006 seems to re-detect the language after backing to the title, this isn't what I mean by dynamic language switching. I mean on a 360, are you able to switch the system language on the fly while a game is running, and does it pick it up immediately? (with no game interaction)

LuismaSP89:
Oh man... at last we can select other languages in "multi" games... this is so ***king great!

src/xenia/app/emulator_window.cc
```diff
@@ -25,6 +25,8 @@
 #include "xenia/ui/imgui_dialog.h"
 #include "xenia/ui/imgui_drawer.h"
 
+uint32_t userlang = 1; // Default as English
```
DrChat: Blah - let's try and not use a global variable here.

src/xenia/kernel/xboxkrnl/xboxkrnl_modules.cc
```diff
@@ -58,7 +59,7 @@ X_STATUS xeExGetXConfigSetting(uint16_t category, uint16_t setting,
           break;
         case 0x0009:  // XCONFIG_USER_LANGUAGE
           setting_size = 4;
-          value = 0x00000001;  // English
+          value = userlang;  // 1 - English as default
```
DrChat: We could maybe make this part of the kernel state - is there no config structure present in the kernel state as of now?

src/xenia/app/emulator_window.cc
```diff
@@ -226,6 +229,36 @@ bool EmulatorWindow::Initialize() {
   }
   main_menu->AddChild(std::move(gpu_menu));
 
+  // User Language menu
+  auto ulang_menu = MenuItem::Create(MenuItem::Type::kPopup, L"&User Language");
+  {
+	  ulang_menu->AddChild(
```
DrChat: It'd be cool if we could ping the game to see what languages it supports before constructing this menu. Dunno if that's possible though.

src/xenia/app/emulator_window.h
```diff
@@ -54,6 +55,14 @@ class EmulatorWindow {
   void GpuTraceFrame();
   void GpuClearCaches();
   void ShowHelpWebsite();
+  void UserlangEnglish();
```
DrChat: Don't add separate function calls per language - instead define a function like SetLanguage(Language l) (where Language is an enum of values)

gibbed:
I know you *can* set the language in system settings, but I don't know if you can do it while a game is running and have the game pick up the change automatically, which is what I'm asking.

altiereslima:
Portuguese Brazilian missing

mchtt:
I don't see that option in menu bar with the build I use ?
Nov 3 2018.

Ulquiorr4-Cifer:
@mchtt it's not merged yet

will be added in ui that is currently in the development

gibbed:
I finally got around to checking: you cannot change system language while a game is running on a Xbox 360. Going into system settings requires exiting the game.

As such, changing language at runtime, while it *may* work in some games, is not how we should handle this in Xenia. It should be set at game boot and unchangeable afterwards.

Margen67:
Maybe it should be added as a gflag instead?

gibbed:
A gflag would be possible, I meant for future UI work.

src/xenia/app/emulator_window.cc
altiereslima: value = 0x00000009;  // Portuguese

Margen67:
@Parovozik Any update on this?

BeRzEsKeR:
Is it possible to solve the problem with the merge?

Triang3l:
This specific pull request will not be accepted, maybe it will be rewritten because it's faaaaar from our code style.

BeRzEsKeR:
It does not seem to be much code but surely since it was implemented until now there have been changes, so it will be time for the person to retake it. Is there anyone who knows your style that wants / can assign this task?

gibbed:
Superseded by 2780e7b1a6b411bcba344f206ea916c7918bb364.

PR #1129, opened 2018-05-01
Update building.md

shadowbrony33:
VS2015 comes with the Windows Universal CRT SDK by default however VS2017 needs to have it manually added.
The Windows 8.1 SDK is required when building on both Windows 8.X and 10.
Python 3.4+ is now required to run the setup scripts.


NekokoTime:
Update building.md #1129 #1134

PR #1131, opened 2018-05-03
Memory leak fix

DatBoiJay3:
This is needed for most games.

Without this fix, Xenia will usually crash after a few seconds of rendering 3D scenes, or even lock up your computer.


DrChat:
This isn't a fix - this merely masks the issue. Closed.

PR #1137, opened 2018-05-05
Update README.md link to xenia-project/xenia

NekokoTime:


DrChat:
AppVeyor is still set up with the old link, so we'll just keep it as is for now. Appreciate it.

PR #1140, opened 2018-05-08
Fix the broken build

Jeremy517:
The broken build was caused by the initializations inside of the struct declaration.  This change removes those initializations and then adds RGBA to all the lines that don't currently explicitly specify something.

An alternative approach to fix the build would be to use constructors.


NekokoTime:
Fix the broken build #1141

DrChat:
Holding off on pulling this for now, because what I'm doing should be valid C++11 syntax, and clang appears to agree.

May be possible that AppVeyor's compilers are out-of-date.

gibbed:
Fixed with 29dbc72d48e0256f2f6cc24d6b3d1a3af5e90ad6.

PR #1145, opened 2018-05-11
[Vulkan] Use Format

NekokoTime:
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkFormat.html


gibbed:
Are you guessing at this or have you seen games use these?

NekokoTime:
@gibbed I am ah suppose and following Vulkan's Link

gibbed:
I don't think we need to define mappings for texture formats unless we know of a game that wants to use it. @DrChat?

gibbed:
Actually, the mappings you've defined are incorrect to begin with.

PR #1146, opened 2018-05-12
Fix texture related memory leak

shadowbrony33:
Actually have textures pending for deletion deleted and not make a huge-a** memory leak


shadowbrony33:
Fixed in f591f2d

PR #1150, opened 2018-05-13
Aegis Wings Fixes - XGI session implementations, Xuid Update, XMsgStartIORequest Thread error for SessionJoin

PermaNulled:
This should allow Aegis Wings to go fully in-game at this point, it may potentially fix other games, but the fixes performed were specifically due to issues Aegis wings was having.

It now goes fully in-game but has some graphic issues on the latest commit.

### DIFF
--- a/src/xenia/kernel/xam/user_profile.cc
+++ b/src/xenia/kernel/xam/user_profile.cc
@@ -14,7 +14,9 @@ namespace kernel {
 namespace xam {
 
 UserProfile::UserProfile() {
-  xuid_ = 0xBABEBABEBABEBABE;
+  // Write XUID with 0xE as it indicates the profile is offline.
+  // https://github.com/hetelek/Velocity/blob/master/XboxInternals/Account/Account.cpp#L77
+  xuid_ = 0xE000000004CE0DB0;
   name_ = "User";
 
   // http://cs.rin.ru/forum/viewtopic.php?f=38&t=60668&hilit=gfwl+live&start=195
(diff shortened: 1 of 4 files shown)


zemzema:
I tested all games that I have already been tested (compatibility list), no changes for them with this patch.

src/xenia/kernel/xam/user_profile.cc
```diff
@@ -14,7 +14,9 @@ namespace kernel {
 namespace xam {
 
 UserProfile::UserProfile() {
-  xuid_ = 0xBABEBABEBABEBABE;
+  // Write XUID with 0xE as it indicates the profile is offline.
+  // https://github.com/hetelek/Velocity/blob/master/XboxInternals/Account/Account.cpp#L77
+  xuid_ = 0xE000000004CE0DB0;
```
DrChat: I can understand the upper `0xE`, but what about the `0x4CE0DB0`?

PermaNulled: The 0x4CE0DB0 was just ripped from an actual XUID off a console as filler data, it's not relevant I just didn't want to write complete 0'd out data as I wasn't sure if a game would be checking for this or not.

Triang3l: Is it some special kind of a user ID consistent across different consoles, or just your personal ID? This should be explained in comments preferably.

pnill: It's not specific or special in any way outside of the upper 0xE a real XUID was simply used incase any game actually had decided to check against it, or instead of leaving those bytes as 0s.

randprint: The 0xE0 is checked against, we used the XUID 0xE000BABEBABEBABE in canary with no adverse effects

halotroop2288:
This PR really got hung up on one issue for almost 3 years?

gibbed:
Superseded by #1444. Will leave open until I've verified things made it over.

PR #1164, opened 2018-05-29
Support for Discord rich presence

jamsterwes:
Added `discord-rpc` library and its dependency `rapidjson` to add full support for:
+ Showing that Xenia is being played
+ Showing what game is being played or "Idle" if none


janisozaur:
Please mind https://github.com/Tencent/rapidjson/issues/1205

benvanik:
Would be nice to have this under xenia/app/discord instead of xenia/discord, as it's only used by app and shouldn't be pulled in elsewhere. Helps keep it clearer what is a core part of the emulator and what is runner specific.

jamsterwes:
Ok, I moved it.

gibbed:
I think until we have an actual UI with configuration options, this would need to be disabled by default, enabled with a command-line flag.

Margen67:
@jamsterwes ?

brodycas3:
You might want to update Xenia's icon since it was recently updated.

JeBobs:
> 
> 
> You might want to update Xenia's icon since it was recently updated.
> 
> image

There, it should be fixed now.

src/xenia/app/discord/discord_presence.cc
```diff
@@ -0,0 +1,63 @@
...
+
+  Discord_UpdatePresence(&discordPresence);
+}
+
+void DiscordPresence::PlayingTitle(std::wstring game_title) {
+  DiscordRichPresence discordPresence;
+  memset(&discordPresence, 0, sizeof(discordPresence));
+  discordPresence.state = "In Game";
+  char mb_game_title[128];  // buffer to hold game title (max 128 bytes)
+  std::wcstombs(mb_game_title, game_title.c_str(),
```
gibbed: This is bad and will convert based on current based on current C locale. Strings provided to Discord rich presence API should be in UTF8.

Should probably use WideCharToMultiByte on Windows because we can explicitly specify the code page there. Not sure about options on posix that don't involve changing C locale.

Issue #1163, opened 2018-05-28
vpkd3d128/vupkd3d128 implementation is incomplete


Triang3l:
While launching a level in [Call of Duty 4: Modern Warfare](https://github.com/xenia-project/game-compatibility/issues/152) pre-alpha build 253, I got a failed assertion during an attempt to execute [vupkd3d128 for 10.10.10.2 vertex data](https://github.com/xenia-project/xenia/blob/master/src/xenia/cpu/backend/x64/x64_sequences.cc#L7485) (executed many times every frame) due to missing implementation.

It looks like along with unpack code for the 10.10.10.2 format being missing, there possibly are multiple other flaws in the implementation of these two instructions.

The 10.10.10.2 vpkd3d implementation in Xenia produces 10-bit floating-point numbers, however, there's no 10.10.10.2 vertex format in DXGI, Direct3D 9 or OpenGL — there's only an 11.11.10 floating-point format, while DXGI only features R10G10B10A2_UNORM and R10G10B10A2_UINT. Call of Duty 4 does unpack, in particular, for values such as 0x00000201, 0x00080400, 0x1FF00000, which are 0.5f±1u for each component if interpreted as UNORM (alpha is stored in the most significant bits).

Also, based on usage examples found by googling `__vpkd3d` and `__vupkd3d`, it appears that vupkd3d "unpacks" D3DCOLORs and short numbers simply by ORing the component integer values with 1.0f, or 0x3F800000, resulting in weird values like 0x3F8000FF. Among these usage examples are:

- [CPixelWriter::WritePixelNoAdvance_BGRA8888](https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/public/pixelwriter.h#L662) of the Source Engine, which packs floating-point values between 0.0f and 255.0f as BGRA8888. The weird 3 + i * 2^-22 formula simply puts the integer part of the low bits of a floating-pointer number, making asfloat(0x404000FE) from 254.0f.
- Some Unreal Engine 3 math code cached in Google search results, which produces a vector of four 1.0f components as `__vspltw( __vupkd3d(__vspltisw( 0 ), VPACK_NORMSHORT2), 3 )`.

[...]

PR #1165, opened 2018-05-30
[CPU] Only truncate integers in vpkd3d and OR with 1.0 in vupkd3d

Triang3l:
Based on the analysis provided in #1163.

This improves rendering in Call of Duty 4 alpha build 253 (which also previously crashed in debug builds because vupkd3d wasn't implemented for uint2.10.10.10):

- Removes random colors on models: [before](https://media.discordapp.net/attachments/309008096800276490/451346142563663902/bog_old.png), [after](https://media.discordapp.net/attachments/309008096800276490/451346187878924288/bog.png).
- Makes certain broken world polygons more visible — the legs of the tables on the firing range can be [somewhat visible](https://i.imgur.com/ehFz9qq.jpg) now (vupkd3d for uint2.10.10.10 is used to unpack them).

Implements vupkd3d for uint2.10.10.10, removes saturation to values near 3 as it's just a magic number that lets you pack whole floating-point numbers in the lowest bits using only the FPU as 3+i*2^-22, removes sign extension for short unpacking (as it results in NaN for negative shorts), and puts 1.0 rather than 3.0 in the higher bits when unpacking because Unreal Engine 3 expects that when creating the 1.0x4 constant.

I have NOT updated the tests because I don't know how they were originally generated, but will do if it's fine to do so.


Triang3l:
This is fundamentally wrong, will send a PR with a different commit later.

PR #1166, opened 2018-05-31
[CPU] v(u)pkd3d128: Support UINT_2101010 and don't saturate D3DCOLOR

### LINKED ISSUE #1163: vpkd3d128/vupkd3d128 implementation is incomplete

Triang3l:
This pretty much fixes #1163, though there's one question left unanswered — is the W component in UINT_2101010 really unsigned? According to the way OpenGL handles it, W is signed, however there's no signed 2.10.10.10 format in D3D9 or DXGI, so the Xbox 360 is free to set its own rules. Since it was handled as unsigned before this pull request, I made the same assumption here.

This fixes broken world polygons (the thin lines coming from a corner of the sky) in the Call of Duty series.

Also providing better explanations of what happens in SHORT2/4 unpacking, and clamping the magnitude when unpacking it, so there can't be -32768.

The saturation was also removed from D3DCOLOR packing because packing can be done with 3+i*2^-22, however, the unpack instruction gives 1|value — in this case, if you pack after unpacking, you'll get four zeros. The Source engine usage example also explicitly clamps the color to 255.


gibbed:
Looks good!

PR #1176, opened 2018-06-09
[Core] Treat .exe files as XEX

Triang3l:
Call of Duty 4 and World at War alphas have .exe files like CoD3SP.exe and CoD3SP_profile.exe alongside default.xex that are actually XEX files.

### DIFF
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -228,7 +228,7 @@ X_STATUS Emulator::LaunchPath(std::wstring path) {
   auto extension = path.substr(last_dot);
   std::transform(extension.begin(), extension.end(), extension.begin(),
                  tolower);
-  if (extension == L".xex" || extension == L".elf") {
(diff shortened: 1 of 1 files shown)


Triang3l:
Should .exe and .elf be added to the filter in the file open menu, or would that lead to false assumptions about the capabilities of the emulator?

DrChat:
Hmm... Odd that CoD has exe filenames, but otherwise I don't see much of an issue with this.
If users try to load PE exes, they'll obviously fail to verify.

Let's not add them to any menus, so users don't start thinking we can run Windows programs.

Otherwise, LGTM!

Triang3l:
Quake 2 on the Quake 4 bonus disc also uses quake2-xenon.exe.

x1nixmzeng:
I did some work a while ago to warn for XNA-based EXE files, but that is a case where EXE files are valid modules to run - eventually.

PR #1177, opened 2018-06-09
[GPU] Depth bias support

Triang3l:
Fixes Z fighting of decals in various games including the Call of Duty series, and possibly shadow acne in some games also.

Since the registers contain offset values for both polygon sides, the actual value is chosen based on the culled side (if none are culled, then choosing the side that has any offset, if both have, using the front). However, I doubt different values for different sides are even exposed via D3D at all, but who knows what games do.

I'm not sure what we should prefer — enable depth offset in all Vulkan pipelines, or toggle based on the actual state. However, it's usually enabled for a small portion of drawn surfaces, and those surfaces already have their own pipelines generally, so I think the second option, implemented here, is better.

### DIFF
--- a/src/xenia/gpu/vulkan/pipeline_cache.cc
+++ b/src/xenia/gpu/vulkan/pipeline_cache.cc
@@ -714,6 +714,66 @@ bool PipelineCache::SetDynamicState(VkCommandBuffer command_buffer,
     vkCmdSetViewport(command_buffer, 0, 1, &viewport_rect);
   }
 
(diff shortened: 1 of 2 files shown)


Triang3l:
Amended to remove semantically incorrect SetShadowRegister usage, hopefully will pass the lint now as well.

Triang3l:
Amended again. 65536 looks less arbitrary than 32768 — and for shadow maps, `sm_polygonOffsetBias` is 0.5, and the value in the register is −0.5/65536, so I'd assume 2 is some decal-specific scale.

PR #1178, opened 2018-06-09
Persist title-specific profile content

Jeremy517:
Save and load title-specific profile content to disk in the content folder for that title.  Allows saving settings for many games and saving games for games that store saved games there, such as BattleBlock Theater.

### DIFF
--- a/src/xenia/kernel/xam/content_manager.cc
+++ b/src/xenia/kernel/xam/content_manager.cc
@@ -23,6 +23,8 @@ namespace xam {
 
 static const wchar_t* kThumbnailFileName = L"__thumbnail.png";
 
+static const wchar_t* kGameUserContentDirName = L"profile";
+
 static int content_device_id_ = 0;
 
 ContentPackage::ContentPackage(KernelState* kernel_state, std::string root_name,
@@ -250,6 +252,20 @@ X_RESULT ContentManager::DeleteContent(const XCONTENT_DATA& data) {
   } else {
     return X_ERROR_FILE_NOT_FOUND;
   }
+}
+
+std::wstring ContentManager::ResolveGameUserContentPath() {
+  wchar_t title_id[9] = L"00000000";
+  std::swprintf(title_id, 9, L"%.8X", kernel_state_->title_id());
+  auto user_name = xe::to_wstring(kernel_state_->user_profile()->name());
+
+  // Per-game per-profile data location:
+  // content_root/title_id/profile/user_name
+  auto package_root = xe::join_paths(
+      root_path_,
+      xe::join_paths(title_id,
+                     xe::join_paths(kGameUserContentDirName, user_name)));
+  return package_root + xe::kWPathSeparator;
 }
 
 }  // namespace xam
--- a/src/xenia/kernel/xam/user_profile.cc
+++ b/src/xenia/kernel/xam/user_profile.cc
@@ -7,6 +7,10 @@
  ******************************************************************************
  */
 
+#include <sstream>
+
+#include "xenia/kernel/kernel_state.h"
+#include "xenia/kernel/util/shim_utils.h"
 #include "xenia/kernel/xam/user_profile.h"
 
 namespace xe {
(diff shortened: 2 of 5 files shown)


src/xenia/kernel/xam/content_manager.cc
```diff
@@ -252,6 +254,25 @@ X_RESULT ContentManager::DeleteContent(const XCONTENT_DATA& data) {
...
 
+std::wstring ContentManager::ResolveGameUserContentPath() {
+  wchar_t title_id[9] = L"00000000";
+  std::swprintf(title_id, 9, L"%.8X", kernel_state_->title_id());
+
+  std::string user_name = kernel_state_->user_profile()->name();
+  size_t len = user_name.length();
+  auto user_name_w =
+      reinterpret_cast<wchar_t*>(alloca(sizeof(wchar_t) * (len + 1)));
+  std::mbstowcs(user_name_w, user_name.c_str(), len + 1);
```
DrChat: Use `xe::to_wstring`

Jeremy517: Thanks.  I've made the change and squashed the changes into one commit.

src/xenia/kernel/xam/user_profile.cc
```diff
@@ -107,7 +115,56 @@ UserProfile::Setting* UserProfile::GetSetting(uint32_t setting_id) {
...
+  }
+  return setting;
+}
+
+void UserProfile::LoadSetting(UserProfile::Setting* setting) {
+  if (setting->is_title_specific()) {
+    auto content_dir =
+        kernel_state()->content_manager()->ResolveGameUserContentPath();
+    std::wstringstream id_stream;
+    id_stream << std::hex << setting->setting_id;
```
DrChat: Could use `xe::format_string("%.8X", setting->setting_id)` instead

src/xenia/kernel/xam/user_profile.cc
```diff
@@ -107,7 +115,56 @@ UserProfile::Setting* UserProfile::GetSetting(uint32_t setting_id) {
...
+}
+
+void UserProfile::SaveSetting(UserProfile::Setting* setting) {
+  if (setting->is_title_specific()) {
+    auto serialized_setting = setting->Serialize();
+    auto content_dir =
+        kernel_state()->content_manager()->ResolveGameUserContentPath();
+    xe::filesystem::CreateFolder(content_dir);
+    std::wstringstream id_stream;
+    id_stream << std::hex << setting->setting_id;
```
DrChat: Ditto.

src/xenia/kernel/xam/user_profile.cc
```diff
@@ -107,7 +115,56 @@ UserProfile::Setting* UserProfile::GetSetting(uint32_t setting_id) {
...
+    auto file_path = xe::join_paths(content_dir, id_stream.str());
+    auto file = xe::filesystem::OpenFile(file_path, "wb");
+    fwrite(serialized_setting.data(), 1, serialized_setting.size(), file);
+    fclose(file);
+  } else {
+    // Unsupported for now.  Other settings aren't per-game and need to be
```
DrChat: Perhaps log this with `XELOGW` so we can pin this down in the future

src/xenia/kernel/xam/user_profile.cc
```diff
@@ -107,7 +115,56 @@ UserProfile::Setting* UserProfile::GetSetting(uint32_t setting_id) {
   if (it == settings_.end()) {
     return nullptr;
   }
-  return it->second;
+  UserProfile::Setting* setting = it->second;
+  if (setting->is_title_specific()) {
+    if (kernel_state()->title_id() != setting->loaded_title_id) {
```
DrChat: I'm confused - why is this a `!=` comparison?

Jeremy517: The idea is that we only load from disk once, and after that it is already in memory.  Once it is in memory, we only need to load from disk if we're now running a different title.  Thus the check here is to see if what we've got in memory is for the title that is currently running.  I'll add a comment explaining.

Jeremy517:
Requested changes as of now has been addressed and squashed into one commit.

src/xenia/kernel/xam/user_profile.cc
Jeremy517: We could just always load it from disk if we're not worried about games spamming the load though.  None of the games I've tried did it more than once.

Jeremy517:
@DrChat, is there anything else you need done in this pull request?

Issue #1187, opened 2018-06-13
Vulkan BufferCache capacity too small for some games


Parovozik:
Tested on (https://github.com/xenia-project/xenia/commit/2314f25fb7252bb9054fd7c863e1a57bee90282d)
Missing graphics for some games will repaired if increase buffer from default '256' to '512'
Source: xenia-gpu-vulkan vulkan_command_processor.cc
`33 constexpr size_t kDefaultBufferCacheCapacity = 256 * 1024 * 1024;`
`33 constexpr size_t kDefaultBufferCacheCapacity = 512 * 1024 * 1024;`
Maybe the root of the problem is elsewhere, but it gives an effect (as an option it can be added as an option in the emulator settings.) :yum: 
# Screenshots:
![bandicam 2018-06-13 15-53-10-046](https://user-images.githubusercontent.com/19775644/41354823-a0d35bc2-6f28-11e8-9b6e-24ea3100594f.jpg)
![bandicam 2018-06-13 15-57-57-675](https://user-images.githubusercontent.com/19775644/41354826-a3b46d18-6f28-11e8-815f-1dda22ecf642.jpg)
![bandicam 2018-06-13 15-53-16-866](https://user-images.githubusercontent.com/19775644/41354856-b8789b2a-6f28-11e8-8916-5659c52def5c.jpg)
![bandicam 2018-06-13 15-58-05-430](https://user-images.githubusercontent.com/19775644/41354836-a9df8a7e-6f28-11e8-96ea-b9e4a9b505d1.jpg)
![bandicam 2018-06-13 15-54-44-572](https://user-images.githubusercontent.com/19775644/41354878-c94d062a-6f28-11e8-801c-287a30cbbab1.jpg)
![bandicam 2018-06-13 15-59-44-143](https://user-images.githubusercontent.com/19775644/41354883-cc1efe4e-6f28-11e8-80a6-eced7355cc61.jpg)

zemzema:
@Parovozik, which software did you use for statistic in top left corner?

Razzile:
@zemzema I believe that is the afterburner overlay. not 100% sure though

Parovozik:
@zemzema  MSI Afterburner + RTSS (all included to AB installer)

Parovozik:
[...]

Issue #745, opened 2017-07-02
Cant launch program


pkmnlovr74:
it says I am missing VCRUNTIME140.dll and MSVCP140.dll and the program cant launch, I forget how to get those, can you tell me?

pkmnlovr74:
I have solved that problem, but not I cant "create real GL context" help

Margen67:
What are the specs of your PC? (CPU, GPU, etc.)

PR #1192, opened 2018-06-20
Update Vulkan BufferCache Capacity

allens39:
Should fix missing graphics in issues #1187, #123, and #745

### DIFF
--- a/src/xenia/gpu/vulkan/vulkan_command_processor.cc
+++ b/src/xenia/gpu/vulkan/vulkan_command_processor.cc
@@ -30,7 +30,7 @@ namespace vulkan {
 using namespace xe::gpu::xenos;
 using xe::ui::vulkan::CheckResult;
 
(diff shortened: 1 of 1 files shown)


gibbed:
This will only hide whatever the issue actually is. The Xbox 360 only has 512MB of RAM, so having a buffer be 512MB is overkill.

allens39:
@gibbed where would the issue lie then? I also dont understand how having a buffer be the size of the xbox 360 ram would be an issue for an emulator trying to get it to work on PCs.

Irixion:
Wouldn't having a large buffer size improve performance anyhow? What of having it as a temporary workaround?

gibbed:
@allens39 The issue lies in that a real 360 wouldn't be able to exhibit the behavior games are doing currently if increasing the buffer size is improving it, so the underlying issue needs to be researched and understood instead of mitigating the issue without understanding it.

PR #1194, opened 2018-06-22
Shim Conversion for xam_ui.cc

allens39:


x1nixmzeng:
You've included another fix in the PR - was that intentional? If f2f0aa0 should be separate it can be reviewed properly

Jeremy517:
He reverted that change, but he needs to squash the three commits into one so the history doesn't get polluted

src/xenia/kernel/xam/xam_ui.cc
```diff
@@ -23,11 +23,11 @@ namespace xam {
 
 std::atomic<int> xam_dialogs_shown_ = {0};
 
-SHIM_CALL XamIsUIActive_shim(PPCContext* ppc_context,
-                             KernelState* kernel_state) {
+dword_result_t XamIsUIActive() {
   XELOGD("XamIsUIActive()");
```
DrChat: This logging line can die.

src/xenia/kernel/xam/xam_ui.cc
```diff
@@ -318,19 +307,16 @@ dword_result_t XamShowDeviceSelectorUI(dword_t user_index, dword_t content_type,
 }
 DECLARE_XAM_EXPORT(XamShowDeviceSelectorUI, ExportTag::kImplemented);
 
-SHIM_CALL XamShowDirtyDiscErrorUI_shim(PPCContext* ppc_context,
-                                       KernelState* kernel_state) {
-  uint32_t user_index = SHIM_GET_ARG_32(0);
-
+dword_result_t XamShowDirtyDiscErrorUI(dword_t user_index) {
   XELOGD("XamShowDirtyDiscErrorUI(%d)", user_index);
```
DrChat: Ditto.

src/xenia/kernel/xam/xam_ui.cc
```diff
@@ -121,16 +108,16 @@ SHIM_CALL XamShowMessageBoxUI_shim(PPCContext* ppc_context,
   XELOGD(
```
DrChat: Ditto.

PR #1226, opened 2018-10-11
[JIT] stmw/lmw

chris-hawley:
This fixes sonic adventures so it is now playable

### DIFF
--- a/src/xenia/cpu/ppc/ppc_emit_memory.cc
+++ b/src/xenia/cpu/ppc/ppc_emit_memory.cc
@@ -9,11 +9,10 @@
 
 #include "xenia/cpu/ppc/ppc_emit-private.h"
 
+#include <stddef.h>
 #include "xenia/base/assert.h"
 #include "xenia/cpu/ppc/ppc_context.h"
 #include "xenia/cpu/ppc/ppc_hir_builder.h"
-
-#include <stddef.h>
 
 namespace xe {
 namespace cpu {
@@ -504,10 +503,26 @@ int InstrEmit_stw(PPCHIRBuilder& f, const InstrData& i) {
   } else {
     b = f.LoadGPR(i.D.RA);
   }
-
   Value* offset = f.LoadConstantInt64(XEEXTS16(i.D.DS));
   f.StoreOffset(b, offset,
                 f.ByteSwap(f.Truncate(f.LoadGPR(i.D.RT), INT32_TYPE)));
+
+  return 0;
+}
+
+int InstrEmit_stmw(PPCHIRBuilder& f, const InstrData& i) {
+  Value* b;
+  if (i.D.RA == 0) {
+    b = f.LoadZeroInt64();
+  } else {
+    b = f.LoadGPR(i.D.RA);
+  }
+
+  for (uint32_t j = 0; j < 32 - i.D.RT; ++j) {
+    Value* offset = f.LoadConstantInt64(XEEXTS16(i.D.DS) + j * 4);
+    f.StoreOffset(b, offset,
+                  f.ByteSwap(f.Truncate(f.LoadGPR(i.D.RT + j), INT32_TYPE)));
+  }
   return 0;
 }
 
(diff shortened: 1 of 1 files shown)


src/xenia/cpu/ppc/ppc_emit_memory.cc
```diff
@@ -511,6 +511,23 @@ int InstrEmit_stw(PPCHIRBuilder& f, const InstrData& i) {
...
+
+  for (uint32_t j = 0; j < 32 - i.D.DS; ++j) {
+    if (i.D.RA == 0) {
+      b = f.LoadZeroInt64();
+    } else {
+      b = f.LoadGPR(i.D.RA + j);
+    }
+    Value* offset = f.LoadConstantInt64(XEEXTS16(i.D.DS + 4 * j));
+    f.StoreOffset(b, offset,
+                  f.ByteSwap(f.Truncate(f.LoadGPR(i.D.RA + j), INT32_TYPE)));
```
Triang3l: It's RS (Register Source), not RA (Register Address). RA contains the address where to save the register to.

src/xenia/cpu/ppc/ppc_emit_memory.cc
```diff
@@ -674,13 +691,19 @@ int InstrEmit_stdbrx(PPCHIRBuilder& f, const InstrData& i) {
...
 int InstrEmit_lmw(PPCHIRBuilder& f, const InstrData& i) {
-  XEINSTRNOTIMPLEMENTED();
-  return 1;
-}
+  Value* b;
 
-int InstrEmit_stmw(PPCHIRBuilder& f, const InstrData& i) {
-  XEINSTRNOTIMPLEMENTED();
-  return 1;
+  for (uint32_t j = 0; j < 32 - i.D.RT; ++j) {
```
Triang3l: `if (i.DS.RT + j == i.DS.RA) { continue; }`, according to the example with -1 and the comment about `lm` on POWER in the documentation? But I'm not sure about this.

src/xenia/cpu/ppc/ppc_emit_memory.cc
Triang3l: It's weird that in the syntax in the documentation, RT (Register Target) is listed, but later, it's RS everywhere. But semantically RS is more correct here.

Triang3l:
By the way, while not really necessary maybe, commit messages are prefixed with the name of the subsystem, such as [CPU] in this case.

src/xenia/cpu/ppc/ppc_emit_memory.cc
```diff
@@ -511,6 +511,23 @@ int InstrEmit_stw(PPCHIRBuilder& f, const InstrData& i) {
   return 0;
 }
 
+int InstrEmit_stmw(PPCHIRBuilder& f, const InstrData& i) {
+  Value* b;
+
+  for (uint32_t j = 0; j < 32 - i.D.DS; ++j) {
```
Triang3l: It's RS here too, not DS, DS is the offset to apply to the address.

src/xenia/cpu/ppc/ppc_emit_memory.cc
chris-hawley: The second part of the if statement satisfies the condition of that example. if we try to load something from GPR 30 into GPR 30, it will not actually have an effect. However, your suggested change is an optimization to this process.

src/xenia/cpu/ppc/ppc_emit_memory.cc
chris-hawley: it appears as though RS is being put in RA for these instructions.

chris-hawley: this appears to be correct. To verify that just the presence of these functions (ie not forcing xenia to exit immediately) was not the reason for the game getting past frame 1, I just had the functions do nothing. it crashed same as before. It appears as if the functions are actually written correctly otherwise crashing would occur.

Triang3l: Maybe it's in RT actually, but will check by disassembling and matching with the values in the instruction.

Issue #1172, opened 2018-06-07
Support Title Updates


DrChat:
Let's see if we can't add support for *.xexp files.

We may have to do some trickery to allow runtime patching (maybe scan for *.xexp in the same directory?)

maxton:
Title updates are officially installed via STFS archives to the hard drive, right? It would be nice to support loading them from the virtual filesystem in addition to loose .xexp files.

x1nixmzeng:
I'm interested in working on this - leave a comment if you have made a start please!

DrChat:
Go for it! I haven't begun researching it. Good starting point is in `cpu/xex_module.cc`.

x1nixmzeng:
I have got a bit stuck reversing the `XEX_HEADER_DELTA_PATCH_DESCRIPTION` structure - which actually contains the patch instructions. Annoyingly `xextool` supports generating patched xex files but decided to do some packing tricks.

emoose:
Had a shot at it and managed to get .xexp's working here: https://github.com/emoose/xenia/commits/xexp

Example log output: https://pastebin.com/gupjZXgB

Had to change a lot to make it work though, first moved all the XEX2 code into XexModule, then had to change how XEX loading works: it now returns X_STATUS_PENDING after loading in XEX headers & data, then if a patch is found it can be applied, and after that UserModule::LoadXexContinue() is called to finish up the XEX load (ie looking for imports/functions/symbols/etc, setting up loader table entry...).

The commit messages go into detail about it a little more, it's mostly done this way so we don't have to bother resetting all the import/function/symbol stuff after patching.

If someone could look it over that'd be great, if you think it's good enough I'll make a PR for it, otherwise I'm happy to leave it as a reference for someone who wants to work on XEXP stuff in future.

DrChat:
@emoose This is actually really wonderful! Go ahead and submit a pull request so I can do a full review - but the code looks great at a glance.

emoose:
@DrChat Will do, since this is such a big change I expect there'll be a lot of things that can be improved, my C++ isn't exactly the best neither (always been more of a C person really :P), but I'll be happy to fix up anything that gets flagged.

PR #1238, opened 2018-10-14
[CPU] Improve XexModule, add support for *.xexp patches.

emoose:
See https://github.com/xenia-project/xenia/issues/1172#issuecomment-429644327 for more info about what's been changed, the commit messages go into it some more too.

If you have any suggestions on things to change please let me know, like I said in that issue thread my C++ isn't really the best, but I'll be happy to try and fix anything that gets brought up.


src/xenia/cpu/xex_module.h
```diff
@@ -30,61 +29,158 @@ class Runtime;
...
+    if (base_address_)
+      return memory()->TranslateVirtual(base_address_);
+    else
+      return nullptr;
+  }
+
  protected:
   std::unique_ptr<Function> CreateFunction(uint32_t address) override;
 
+  void DecryptSessionKey(const uint8_t* key);
```
DrChat: Instead of having a function dedicated to doing this - we can do DecryptBuffer instead.
i.e:
```
      DecryptBuffer(xe_xex2_retail_key,
                    reinterpret_cast<const uint8_t*>(sec_header->aes_key), 16,
                    session_key_, 16);
```

src/xenia/cpu/xex_module.h
```diff
@@ -30,61 +29,158 @@ class Runtime;
...
+
   uint32_t GetProcAddress(uint16_t ordinal) const;
   uint32_t GetProcAddress(const char* name) const;
 
-  bool ApplyPatch(XexModule* module);
   bool Load(const std::string& name, const std::string& path,
-            const void* xex_addr, size_t xex_length);
-  bool Load(const std::string& name, const std::string& path, xe_xex2_ref xex);
+            const void* xex_addr, size_t xex_length,
+            XexModule* base_xex = nullptr);
```
DrChat: What made you decide to load patches on top of their base XEX files?
It looks like you're doing this to reuse a lot of the loader code, and I'd like to hear your thoughts. I originally wanted to have an ApplyPatch function.

src/xenia/cpu/xex_module.cc
```diff
@@ -82,6 +95,101 @@ const xex2_security_info* XexModule::GetSecurityInfo(
...
+  for (std::vector<PESection>::iterator it = pe_sections_.begin();
+       it != pe_sections_.end(); ++it) {
+    if (!strcmp(it->name, name)) {
+      return &(*it);
+    }
+  }
+  return nullptr;
+}
+
+typedef struct mspack_memory_file_t {
```
DrChat: Could possibly move this MSPack handling code into a util library, but not a biggie at this point.

src/xenia/cpu/xex_module.cc
```diff
@@ -178,57 +707,387 @@ bool XexModule::ApplyPatch(XexModule* module) {
...
   // Setup debug info.
   name_ = std::string(name);
   path_ = std::string(path);
-  // TODO(benvanik): debug info
+
+  uint8_t* data = memory()->TranslateVirtual(base_address_);
+
+  // Load in the XEX basefile
+  // We'll try using both XEX2 keys to see if any give a valid PE
+  while (true) {
```
DrChat: Instead of doing a loop here to try both decryption keys, it'd be better to just try loading the xex with the retail key first, and if that fails, try with the devkit key.
If they both fail, error out.

Razzile: Isn't --xex_dev_key used to tell xenia to use the dev key?

src/xenia/cpu/xex_module.h
emoose: Good idea, I think there's another part in ReadImageDeltaCompressed that calls rijndael stuff directly which can use that too.

src/xenia/cpu/xex_module.h
emoose: Well I did start by adding to ApplyPatch, but when I noticed it was doing most of what Load/ReadImage already did I decided an overlay approach would be better (also at that time though I thought the ReadImageCompressed function could be re-used with some small tweaks, which turned out not to be the case...)

Changing it to back to be like ApplyPatch might actually be better, wouldn't have to keep hold of base_xex_ or anything. I'll try looking into it now.

src/xenia/cpu/xex_module.cc
emoose: Yeah would be nice, this is just copied from the top of xex2.cc, I'm not really sure what the best way to move it is though..

src/xenia/cpu/xex_module.cc
emoose: As in passing something like "bool use_dev_key" to XexModule? That probably would be better.

@Razzile this PR tries both keys right now, so the xex_dev_key option is removed

emoose:
Added most of the suggestions (will commit once I know I didn't break anything...), will look into changing it to use ApplyPatch() instead too. (also will try squashing these commits before it gets merged)

BTW does anyone know I can get it reformatted properly?
I tried "xb format" / "xb format --origin" but looks like lint is still failing, I guess because the first move-code-to-XexModule commit is already pushed at my origin "xb format --origin" doesn't work?

I'd really rather not rebase and have to redo those commits again, would prefer keeping the move-to-XexModule commit as-is and squashing all the reformat/code-review stuff into the xexp commit.

emoose:
Just noticed debug build gives an assert on "assert_zero(pointer_ref_count_);" in kernel/xobject.cc when deleting patch_module with this, I'm sure the earlier commits didn't trigger that, but I'm not sure how those changes would neither...

Hopefully ApplyPatch() rewrite might get rid of it, otherwise I'm not really sure what could be causing it.

emoose:
Done the ApplyPatch rewrite, luckily was pretty simple to change ReadImageDeltaCompressed into it.

The .xexp XexModule has to keep a copy of the xexp data now though (can't store it at xexp's base address, since xexp base addr == xex base addr...), but I guess that's better than needing to keep hold of a base_xex pointer.

Also object_ref looks like it fixed the assert too, thanks for the info 👍

x1nixmzeng:
The linter is stilll failing.

If you stage your local changes (add them all with `git add .`) then run `./xb.bat format`, then your changes will be formatted and unstaged automatically.

Check the log from the Travis link - it shows you what is expected

emoose:
Redid the commits with "xb format" this time, there were some small things xb format seemed to miss which the linter still complained about though, maybe a bug? (linter log here: https://travis-ci.org/xenia-project/xenia/jobs/443963977)

Anyway I merged all the changes into the XEXP commit, also reversed ApplyPatch so it's called on the patch module XexModule instance instead (ie: patch_module->ApplyPatch(base_module) ), seems that was how ApplyPatch was going to be used originally, not sure why I had it flipped around...

Now I'm just crossing my fingers hoping the linter won't fail on me again :P

emoose:
And now it's finally passed the linter :)

It might be worth testing more XEXP files against it before merging though... I've only tested Halo 3 and XBLA Perfect Dark so far, seems to work fine for those, but there's still a ton more title updates out there... 

(also if any of them have non-zero delta_*_offset fields in the patch descriptor that might cause some problems, not really sure how they're meant to be handled since I haven't seen any patches with them set...)

Anyway if someone does find any patches that don't work please let me know!

gibbed:
@emoose Are you asserting offsets are zero then? (if not, I would suggest adding asserts for those cases)

emoose:
@gibbed I do have two asserts for some of those offset fields, but now I just looked and I might be missing them for delta_image_source_offset / delta_image_target_offset, will add those in now.

emoose:
Added those asserts, also logs a warning for people not running it with a debugger, so if those fields do cause a crash at least the user can know why.

Anyway I'll see about downloading a ton of TUs later and check if any have non-zero fields, hopefully those can help show us how to handle them (or at least tell us whether those fields ever actually get set)

emoose:
Think I've figured out how those offsets are used, the PE data at **[\*_source_offset:\*_source_size]** is just copied to **\*_target_offset** right before patching (not really sure why this is done though, guess it must help with compression in some way or something)

Also looks like when the patched PE is smaller than the original it nulls out any extra bytes from the original data first, I added some code to try and release the extra/unused PE data from memory after patching too, but not really sure if I'm doing that properly (the heap->Release() function looks like it only works on entire regions so I had to use heap->Decommit() instead)

Checked the patched image Xenia creates against one made with xextool and it seems to match fine, but I only have 1 xex/xexp pair that uses those fields though, so haven't really been able to test it out all that much... will try getting hold of some more games soon to test it out even more.

(Speaking of testing, maybe it'd be worth releasing a build on Discord with this + d3d12 branch merged for people to test things out against it? I'm sure people on there probably have a lot more games on-hand than I do :P )

Anyway with this the XEXP part of title updates should be pretty much done, now we just need to add in support for loading whole TU packages...
I did try hacking in something to load from a package just to test that it actually works, and seems like it does fine, will start looking into a proper way to do it soon.

PR #1240, opened 2018-10-20
Define ALL imports for xboxkrnl, xam, and XBDM

0x8080:


0x8080:
I've been using tabs, even did **xb lint** and **xb format** before pushing

emoose:
@0x8080 in earlier Xam "XamBuildLegacySystemResourceLocator" and "XamBuildSharedSystemResourceLocator" both pointed to the same function, I figured that's why Xenia has it down as XamBuildSharedSystemResourceLocator_

0x8080:
Ah, I think I've also see something similar to that with some functions that have the **Ex** suffix, like **XamShowMessageBoxUIEx**

emoose:
Some *Internal funcs seem to be like that too, like XamNotifyCreateListenerInternal, makes them pretty easy to implement heh.

Pretty interesting that XamBuildLegacy... has it's own func in later xam's though, I reimplemented that to get 1888 dash to run (just calls the XamBuildSystem... func instead right now), but maybe I'll have to look into the later xam's for it too.

Really hope they don't go around changing how exports work every other version though, would make supporting other dash versions a real chore... but I guess they probably wouldn't change many game-related exports that much (I think I've heard of games that don't run properly on newer versions though... but I might be thinking of a different console)

0x8080:
@emoose They don't change export ordinals after every new dash, they don't even change between retail and debug kernels

emoose:
Oh yeah I know, I just mean the workings of the exports themselves, like with XamBuildLegacy.. in earlier Xam it just returns the same as XamBuildSystem.., but it seems in later Xam it must be doing something else if it has it's own function now.

Really hope they don't change these internal-not-used-by-games functions that much, but I guess it probably shouldn't matter much for the main point of Xenia though.

0x8080:
looks like formatting fucking broke again

PR #1243, opened 2018-10-22
Various additions to support dash.xex

emoose:
Here's some changes need to make 1888 dash.xex run, the commit messages go into what was changed & why, but here's a summary:

- Added required exports (mostly completely implemented afaik, except for one)
- Minor bug fix for _snwprintf
- Fixes for Xam's XMP app
- Added more mount points for the local folder

With these the dash.xex will start running... but quickly crash due to something I haven't figured out yet. 
You can patch the dash.xex to stop the crash from happening though (the subs that crash have code paths that can return an error code, just patch them so they always use the error code path)
It's kinda cheating though so I'm still trying to find the real cause of it (also I think that patch stops most of the blades being drawn, since only the settings blade actually has any content right now)

dash.xex also requires some resources inside xam.xex, luckily it's easy to make it load these resources from a local file instead (which doesn't even require a code patch neither!), but it means you'll have to extract the resources from a 1888 xam.xex into dash.xex's folder (and add a .xzp extension), the two resources you need are called gamercrd.xzp and shrdres.xzp.

Normally dash.xex would load into the initial setup/OOBE, which is useless right now and can't progress due to missing exports, but by default Xenia will set an XConfig flag that skips the OOBE and lets it load straight into the dashboard, this flag can be unset if needed by using the **--xconfig_initial_setup** parameter.

This PR also has some fixes to help get xshell.xex running too (PVR register & /Device/LauncherData mount path), but xshell won't run with this commit since it needs some import libraries to be loaded on launch, which Xenia doesn't seem to do yet...
I did find a fix for that which was pretty simple, and let XShell load up fine, but I'm not sure if that was a proper way to do it, might have some side-effects so I haven't included that in the PR. I'll post about that fix in the issue section soon.


gibbed:
There's only four commits here and you're making a lot of changes.

IMO, all of the bugfixes should be made their own PRs as it's not directly related to the dash stuff. Too much of this is bundled together under generic commit messages.

emoose:
I guess maybe the export commit could be split up, I figured most of it was tied together and didn't really deserve their own commits... but looking at it now it does seem pretty huge though, I'll remove that from this PR and redo it in a different one.

Anyway sorry for the huge commits, I usually code & test things out of tree and then copy out/commit the things that work, will try keeping them smaller next time.

gibbed:
No worries at all, that's what reviews are good for.

emoose:
Alright, I've split the bugfixes and the exports into seperate PRs: https://github.com/xenia-project/xenia/pull/1245 and https://github.com/xenia-project/xenia/pull/1244

I've left the miscellaneous stuff needed for dash.xex in this PR, I suppose at least now even if this doesn't get merged the other stuff could be :) 

Hopefully the new commits are a lot better than before too :P

emoose:
Hmm, seems I was wrong about these PRs not working anymore, just tried merging them again with latest master and 1888 dash seems to load fine now, guess I was doing something wrong before.

Could someone else try it out so I can be sure? You'll have to:

- clone latest master
- merge in changes from this PR and #1245 
- search google for "1888 dashboard update", you should find a download for the 1888 NAND files (xam.xex, dash.xex etc...)
- extract all those files somewhere
- download xextool, put xextool.exe in same folder as 1888's xam.xex
- from command prompt, go to your extracted 1888 folder and run "xextool.exe -d . xam.xex", this should extract 4 new files from xam: gamercrd, mplayer, shrdres and xam
- rename those 4 files so that they have a .xzp extension (gamercrd.xzp, shrdres.xzp, xam.xzp...)
- now run dash.xex from your merged xenia build

With any luck hopefully the dashboard will run fine, if someone can try this and report back it'd really be appreciated!

TGLTClutch:
I get error when running xextool.exe -d . xam.xex when running powershelll from the same folder as xam.xex. So I can't test this out

randprint:
I am using this merged with latest master and can confirm it still works

TGLTClutch:
So it's safe to say this PR is working.

emoose:
Dash itself still works but looks like initial setup/OOBE is now crashing after selecting a language, log shows it calling _vscwprintf & vswprintf tons of times before it crashes, and running with debugger attached shows it crashing inside ntdll with a messed up stack, so I guess some loop is broken somewhere...

It did work fine before though, previously would at least get up to a screen asking you to input a new gamertag (which I was hoping to try out with the new UI notification fixes), not really sure what could have broken it, will have to try a bisect soon.

jgoyvaerts:
Could any of these be the issue?
https://github.com/xenia-project/xenia/commit/8b1f728d85f5521a1212c51aab16657050c4aa96
https://github.com/xenia-project/xenia/commit/40cc8c52d78a53186a2a6abaa7a7627a09b79f36

gibbed:
I'm sure emoose meant the guest imports of those functions, not host usage.

emoose:
I was wondering that too, but it looks like those changes should only really affect Xenia's logging stuff? Not really sure...

Still, pretty funny that just after some printf changes were made, we find out dash.xex is having printf problems - but then again I don't think dash's OOBE has been tried out in a long time, there's been a lot changed since the last time I posted about it anyway 😛

SmilerRyan (approved):

PR #1245, opened 2018-10-22
[Kernel] Add 12 kernel/xam exports used by dash.xex

emoose:
This adds the following exports needed for dash.xex & xshell.xex:

- xboxkrnl::ExConsoleGameRegion (set to 0xFFFFFFFF)
- xboxkrnl::_vsnwprintf
- XamContentCreateDeviceEnumerator
- XamFormatDateString
- XamFormatTimeString
- XamBuildResourceLocator
- XamBuildGamercardResourceLocator
- XamBuildSharedSystemResourceLocator
- XamBuildLegacySystemResourceLocator
- XamBuildXamResourceLocator
- XamNotifyCreateListenerInternal (1 unknown param)
- XamGetOnlineSchema (returns dummy data)

The commit messages mostly go into what the exports are for, besides XamNotifyCreateListenerInternal these are pretty much completely implemented, the XamFormat* exports currently don't localize for different system regions though, but since it's only really used in dash.xex AFAIK it's probably not a huge priority right now :P 

XamBuildResourceLocator returns a string that can refer to a resource either from a file path or an XEX resource.
On 360 the other *ResourceLocator exports call that and pass a handle to xam.xex's module to retrieve resources from it, our xam.xex doesn't have any kind of resources though, but luckily we can just make XamBuildResourceLocator return a file path for those instead and the requesting app will happily read them from your local directory (as long as media:\ is symlinked anyway)

With dash.xex it requires gamercrd & shrdres to be extracted from xam.xex (with a .xzp extension), you can use "xextool -d . xam.xex" to extract them.


BearGamesStudio:
how do i get nxe2 dashes to run? thx.

Issue #1247, opened 2018-10-25
XamContentOpenFile, xam::ContentManager


gibbed:
This appears to open a PIRS-format archive to access nested files (possibly mounted automatically?).

Used by NUI to mount a few specific files to access files within:

```c
XamContentOpenFile(0xFF, "stexemplar", "update:\\Database.xmplr", 0x43, nullptr, nullptr, nullptr);
XamContentOpenFile(0xFF, "stexemplar", "game:\\Database.xmplr", 0x43, nullptr, nullptr, nullptr);
XamContentOpenFile(0xFF, "nuiidentity", "update:\\NuiIdentity.bin.be", 0x43, nullptr, nullptr, nullptr);
XamContentOpenFile(0xFF, "nuiidentity", "game:\\NuiIdentity.bin.be", 0x43, nullptr, nullptr, nullptr);
```

Having this implemented properly is probably required for NUI (Kinect) to initialize properly in games.

Will require some reworking of how `xam::ContentManager` to enable different kinds of content sources to make this work. Currently everything is a flat directory structure.

PR #1250, opened 2018-11-01
[Kernel] Support loading XEXP from STFS package/extracted TU

emoose:
UserModule will now try using ContentManager to find & mount a TU to the update:\ root, ContentManager is also updated so that it'll support loading STFS packages too.
(this doesn't affect #1247 though, since that requires loading the package from a specific path)

I also had to add a "title id override" field to ContentManager, as the kernel_state_->title_id() that it used to call won't work as the XEX isn't fully loaded in yet...
This is probably a good thing anyway, in future we can probably use it to support games that load packages from other title IDs (eg. games that search for the previous games save data to unlock a special bonus)

TUs are loaded from Content\[TitleID]\000B0000\, the TU can either be an STFS package, or a folder with the extracted TU files inside.
On X360 updates can also be inside Hdd1:\Cache\ but I haven't added support for that yet, as it'd require checking title ID against STFS headers etc... that should probably be added eventually though, so people can just dump their TUs into a single folder instead of needing to create a bunch of them.

### DIFF
--- a/src/xenia/kernel/user_module.cc
+++ b/src/xenia/kernel/user_module.cc
@@ -19,6 +19,7 @@
 #include "xenia/emulator.h"
 #include "xenia/kernel/xfile.h"
 #include "xenia/kernel/xthread.h"
+#include "xenia/vfs/devices/stfs_container_device.h"
 
 namespace xe {
 namespace kernel {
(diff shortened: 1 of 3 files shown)


altiereslima:
Content folder in C:\Users\UserName\Documents\Xenia\content?

emoose:
Sorry for the long delay in updating it, used githubs conflict resolver to make it work with the latest master, so hopefully this should be good to go now.

It'd be appreciated if someone could test and make sure it works properly, IIRC you should just be able to put a TU package inside /Content/[title id]/000B0000/, then load the XEX and hopefully it should try patching it with the TU (not sure if the patch will work successfully though, I did hear xexp might be broken? Anyway as long as it does attempt to patch it then this P/R is working fine)

PR #1254, opened 2018-11-07
Games on Demand / Multi-file STFS (SVOD) Support

Wildenhaus:
Tested on several GoD container and XBLA Games (Burnout Revenge, Halo 3, Halo 3 ODST, JUJU, Duke Nukem 3D). Diff'ed various container files from the original source with no discrepancies. Handles both GDF and non-GDF formats with the same functions.

### DIFF
--- a/src/xenia/vfs/devices/stfs_container_device.cc
+++ b/src/xenia/vfs/devices/stfs_container_device.cc
@@ -288,10 +288,16 @@ StfsContainerDevice::Error StfsContainerDevice::ReadSVOD() {
 
 StfsContainerDevice::Error StfsContainerDevice::ReadEntrySVOD(
     uint32_t block, uint32_t ordinal, StfsContainerEntry* parent) {
+  // For games with a large amount of files, the ordinal offset can overrun
+  // the current block and potentially hit a hash block.
+  size_t ordinal_offset = ordinal * 0x4;
+  size_t block_offset = ordinal_offset / 0x800;
+  size_t true_ordinal_offset = ordinal_offset % 0x800;
+
   // Calculate the file & address of the block
   size_t entry_address, entry_file;
-  BlockToOffsetSVOD(block, &entry_address, &entry_file);
-  entry_address += ordinal * 0x04;
+  BlockToOffsetSVOD(block + block_offset, &entry_address, &entry_file);
+  entry_address += true_ordinal_offset;
 
   // Read block's descriptor
   auto data = mmap_.at(entry_file)->data() + entry_address;
--- a/src/xenia/vfs/devices/stfs_container_device.cc
+++ b/src/xenia/vfs/devices/stfs_container_device.cc
@@ -10,6 +10,7 @@
 #include "xenia/vfs/devices/stfs_container_device.h"
 
 #include <algorithm>
+#include <queue>
 #include <vector>
 
 #include "xenia/base/logging.h"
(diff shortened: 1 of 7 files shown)


Wildenhaus:
I just realized that this probably wont handle large files split up across multiple datafiles. Working on a fix, don't merge before it's appended.

Wildenhaus:
Fixed device entries spanning multiple datafiles. Datafiles are split into ~162MB chunks. Tested with a few Halo 3 maps as well as Dead Space 2's bigfile0.viv (3GB). All files were diff'ed from the originals and no differences were found.

Aside from code review, this pull is in working state and ready for merge.

x1nixmzeng:
Great work!

Could you double check the VFS dump tool works with your change?

https://github.com/xenia-project/xenia/tree/master/src/xenia/vfs

Wildenhaus:
> Could you double check the VFS dump tool works with your change?

VFS Dump Tool is confirmed working. Top screenshot is dumping an nxeart STFS container, and bottom screenshot is a dump of a GoD Container, plus xenia running that dump.

vfsdump

If any issues arise, be sure to do a clean-rebuild. It was acting weird for me until I realized it was using an incomplete version of my changes to `stfs_container_device.cc`.

gibbed:
1. Why were two ancient PRs pulled into this?
2. Please clean up your commit messages and squash unnecessary ones.

Wildenhaus:
There are a few edge cases with offset calculation that are causing issues. Working on a more robust way to calculate offsets based on file layout.

Wildenhaus:
- Rewrote SVOD offset calculation to no longer rely on known hard-coded addresses. Should now work with all cases. Added documentation for reference.
 - Added support for STFS Metadata v2

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -617,6 +612,20 @@ bool StfsHeader::Read(const uint8_t* p) {
   title_thumbnail_image_size = xe::load_and_swap<uint32_t>(p + 0x1716);
   std::memcpy(thumbnail_image, p + 0x171A, 0x4000);
   std::memcpy(title_thumbnail_image, p + 0x571A, 0x4000);
+
+  // Metadata v2 Fields
+  std::memcpy(series_id, p + 0x3B1, 0x10);
```
x1nixmzeng: Should this be checking the metadata version is 2 (or greater) before reading these fields?

Wildenhaus: In the case of a game being Metadata v1, these offsets are unused, so they contain null data. Considering all of the fields added in Metadata v2 are irrelevant to emulation and will default to the same null data in `struct StfsHeader` if not read, I don't think it'll be an issue, but can be corrected if it's considered a code quality issue.

Wildenhaus:
- Now supports all known layouts (EGDF, XSF, Single File)
- Major refactoring, code cleanup, and extensive documentation.
- Fixed code quality issue pointed out by @x1nixmzeng 

Tested both in-game and binary diff'ing VFS dump files. Works as it should for Digital Download, Installed, Arcade, and converted games (done by third-party tools, XSF).

src/xenia/vfs/devices/stfs_container_device.cc
Wildenhaus: Conditionally parsing these fields as of the latest commit.

Wildenhaus:
Corrected an issue where file entry ordinal calculation in games with a large amount of files (e.g. Mass Effect with 4508 files) would run into hash blocks.

Ulquiorr4-Cifer:
Was this merged into DX12 or just master? 🤔

PR #1263, opened 2018-11-17
Fixes  #784

### LINKED ISSUE #784: Save System 'Not enough space' (solved)
I tried many games and sometimes saw message 'Not enough free space...'.
Example: Brave, Bully, Phineas and Ferb Quest for Cool Stuff, Red Dead Redemption, Spongebob Plankton Robotic Revenge, Wipeout Create and Crush and other.
This specific error and I solved it as follows:
**Path to file [...xenia-master\src\xenia\kernel\xam\xam_content.cc]**

> 28      static const DeviceInfo dummy_device_info_ = {
> 29      0xF00D0000,
> 30      1,
> 31      60ull * 1024ull * 1024ull * 1024ull,  // 60GB
> 32      27ull * 1024ull * 1024ull * 1024ull,  // 27GB, so it looks a little used.
> 33      L"Dummy HDD",
> 34      };

In original there 120GB/100GB
I think this is absolutely safe and should be added to the master build :)

# Screenshots from xenia-master without it fix
bandicam 2017-08-20 18-51-30-858
bandicam 2017-08-20 18-40-55-052
bandicam 2017-08-20 18-40-05-973
bandicam 2017-08-20 18-39-14-984
bandicam 2017-08-20 18-38-07-627

goedzo:
Since this fixes  #784 and is confirmed, just created the pull request myself:
https://github.com/xenia-project/xenia/issues/784

Please merge.

### DIFF
--- a/src/xenia/kernel/xam/xam_content.cc
+++ b/src/xenia/kernel/xam/xam_content.cc
@@ -28,8 +28,8 @@ struct DeviceInfo {
 static const DeviceInfo dummy_device_info_ = {
     0xF00D0000,
     1,
-    120ull * 1024ull * 1024ull * 1024ull,  // 120GB
-    100ull * 1024ull * 1024ull * 1024ull,  // 100GB, so it looks a little used.
+    60ull * 1024ull * 1024ull * 1024ull,  // 60GB
+    27ull * 1024ull * 1024ull * 1024ull,  // 27GB, so it looks a little used.
     L"Dummy HDD",
 };
 


gibbed:
The reason such a change has not been merged is because this is a blind change that may or may not work for some games without any explanation of *why* some games have issues with these values. It's better left as an open issue until it's properly researched.

goedzo:
Sorry but it's not a blind guess actually. It's caused by an overflow and incorrect HD detection in these games. When trying to detect the free space of the storage based on the hd inside the 360. There are various hd sizes in the original 360 and not all the games detect the disk size properly. In some cases the game thinks there is a 60gb disk inside that is 100gb full. That causes it to think it's full. In other cases the game just writes without checking and checks for an exception. This pr makes sure to set the hd size to 60gb to match the size inside the emulated xbox device. 

Hope this explains the cause this issue for you.

gibbed:
Okay, how are games calculating this in their code? Which games did you test?

goedzo:
That doesn't work with all the games but for some of them that is rounding issue based on exactly being full 100gb. So please note that setting them to 99gb is not the solution for this issue in all cases, although it helps in more off them.

goedzo:
@gibbed are you able to re-open this one and merge, or do I need to re-create the PR?

gibbed:
It won't be merged. You haven't properly researched it via debugging game code.

goedzo:
Based on what do you make that assumption? But hey, since you have clearly made your mind already, feel free to check the code yourself. Bladestorm is easier to check this with.

Parovozik:
@goedzo This error occurs for all these games in the same way. And this is not a 60GB problem, because the error disappears, for example, at 120 \ 101 or 120 \ 99, but it is trouble at 120 \ 100

TheCubeMaster019:
@goedzo what exact function or instruction in the game code is causing the overflow? please provide an example so we can test edge cases.

PR #1264, opened 2018-11-17
Fix Linux Build Errors By Adding Virtual Destructors to Abstract Classes

bmzhao:
Made xe::kernel::App and xe::cpu::StackWalker destructors virtual, since they are abstract classes.
This fixes build the following build errors on Linux:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0/../../../../include/c++/7.3.0/bits/unique_ptr.h:78:2: error: delete called on
      'xe::kernel::xam::App' that is abstract but has non-virtual destructor [-Werror,-Wdelete-non-virtual-dtor]
        delete __ptr;
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0/../../../../include/c++/7.3.0/bits/unique_ptr.h:78:2: error: delete called on
      'xe::cpu::StackWalker' that is abstract but has non-virtual destructor [-Werror,-Wdelete-non-virtual-dtor]
        delete __ptr;
        ^ 
```

### DIFF
--- a/src/xenia/cpu/stack_walker.h
+++ b/src/xenia/cpu/stack_walker.h
@@ -92,6 +92,8 @@ class StackWalker {
   // populated.
   virtual bool ResolveStack(uint64_t* frame_host_pcs, StackFrame* frames,
                             size_t frame_count) = 0;
+
+  virtual ~StackWalker() = default;
 };
 
 }  // namespace cpu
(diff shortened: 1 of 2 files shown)


gibbed:
Please split this into two commits, one per file, with individual messages that follow the formatting of existing commits.

IE:
`[Kernel] Made xe::kernel::App destructor virtual since it is an abstract class.`
`[CPU] Made xe::cpu::StackWalker destructor virtual since it is an abstract class.`

bmzhao:
No probs! Split the PR into 2 commits.

PR #1273, opened 2018-11-20
Allow depth buffers to have mips

goedzo:
While debugging Bully, i ran into an assertion which thought that Depth buffers could not have mips. Well it seems that in some games they do, so fixed this.

### DIFF
--- a/src/xenia/cpu/ppc/ppc_emit_memory.cc
+++ b/src/xenia/cpu/ppc/ppc_emit_memory.cc
@@ -9,11 +9,10 @@
 
 #include "xenia/cpu/ppc/ppc_emit-private.h"
 
+#include <stddef.h>
 #include "xenia/base/assert.h"
 #include "xenia/cpu/ppc/ppc_context.h"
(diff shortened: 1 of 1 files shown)


jgoyvaerts:
Cleanup your commit history, your PR contains 5 commits

goedzo:
That's what i tried to do by removing changes i made earlier. I don't know how to exclude other commits from pr's (or to delete them) so this is how i solved it. As you can see, only one file is effectively changed in master. If you can explain how to exclude files from a pr, or commits that would be great. I was not able to find a way to do this.

jgoyvaerts:
https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit
https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

jgoyvaerts:
Also you should probably prefix your commit with [Vulkan]

goedzo:
@jgoyvaerts thanks for the help! Took me a while to get it right, but now a new commit and PR has been created #1274

PR #1274, opened 2018-11-20
[Vulkan] Allow depth buffers to have mips

goedzo:
Single commit only.

### DIFF
--- a/src/xenia/gpu/vulkan/texture_cache.cc
+++ b/src/xenia/gpu/vulkan/texture_cache.cc
@@ -1078,10 +1078,11 @@ bool TextureCache::UploadTexture(VkCommandBuffer command_buffer,
   if (dest->format == VK_FORMAT_D16_UNORM_S8_UINT ||
       dest->format == VK_FORMAT_D24_UNORM_S8_UINT ||
       dest->format == VK_FORMAT_D32_SFLOAT_S8_UINT) {
-    // Do just a depth upload (for now).
-    // This assumes depth buffers don't have mips (hopefully they don't)
-    assert_true(src.mip_levels() == 1);
-    copy_regions[0].imageSubresource.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
+    // Depth buffers can have mips too
+    int mip_levels = src.mip_levels();
+    for (int i = 0; i < mip_levels; i++) {
+        copy_regions[i].imageSubresource.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
+    }
   }
 
   vkCmdCopyBufferToImage(command_buffer, staging_buffer_.gpu_buffer(),


src/xenia/gpu/vulkan/texture_cache.cc
```diff
@@ -1078,10 +1078,17 @@ bool TextureCache::UploadTexture(VkCommandBuffer command_buffer,
   if (dest->format == VK_FORMAT_D16_UNORM_S8_UINT ||
       dest->format == VK_FORMAT_D24_UNORM_S8_UINT ||
       dest->format == VK_FORMAT_D32_SFLOAT_S8_UINT) {
-    // Do just a depth upload (for now).
-    // This assumes depth buffers don't have mips (hopefully they don't)
-    assert_true(src.mip_levels() == 1);
-    copy_regions[0].imageSubresource.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
+    // Depth buffers can have mips, so dont's assert this:
+    // assert_true(src.mip_levels() == 1);
+    if (src.mip_levels() == 1) {
```
jgoyvaerts: The if is not required, just the for loop gives the exact same result

src/xenia/gpu/vulkan/texture_cache.cc
```diff
@@ -1078,10 +1078,17 @@ bool TextureCache::UploadTexture(VkCommandBuffer command_buffer,
   if (dest->format == VK_FORMAT_D16_UNORM_S8_UINT ||
       dest->format == VK_FORMAT_D24_UNORM_S8_UINT ||
       dest->format == VK_FORMAT_D32_SFLOAT_S8_UINT) {
-    // Do just a depth upload (for now).
-    // This assumes depth buffers don't have mips (hopefully they don't)
-    assert_true(src.mip_levels() == 1);
-    copy_regions[0].imageSubresource.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
+    // Depth buffers can have mips, so dont's assert this:
```
jgoyvaerts: If Depth buffers can have mips (hopefully some other dev can validate this), change the line to 
```suggestion
    // Depth buffers can have mips too
```
and remove the commented assert line

src/xenia/gpu/vulkan/texture_cache.cc
goedzo: I know, but i kept the original code in there for reference and in case you want to do something special in the case there is only a single level in there. I can imagine that in some cases you want to do something different, so this allows that.

src/xenia/gpu/vulkan/texture_cache.cc
goedzo: As explained, kept the original assertion for reference. Not sure how you usually deal with these kind of things for this project.

src/xenia/gpu/vulkan/texture_cache.cc
jgoyvaerts: "Keeping the code for reference" is what git does for you, no need to keep code around which doesn't do anything.
If you want to clarify the for loop, add a comment stating that i = 0 is different from the rest, and/or why.
If someone wants to implement something different for i=0, he can change the code again later

goedzo: Ok clear. I'll just remove it for performance reasons so that another developer can always try to do something different :). So will commit your requests here.

goedzo: This line will fail:

 if (src.mip_levels() == 1) 

src.mip_levels() must be stored in a variable first.

goedzo: Ok fixed with my latest commit.

goedzo:
@jgoyvaerts  should be clean now :). Thanks for the review! 👍

jgoyvaerts:
You have 4 commits again...
https://github.com/xenia-project/xenia/pull/1273#issuecomment-440278429

And NO, you don't have to close this PR and make a new one, just do what is said in the link

goedzo:
Well I will try, however when I did the above, my code changes were gone completely :(

Triang3l:
Have you even tested this, and how do you expect the mipmaps to be created? Resolving 4 samples into 1 texel, then reuploading to EDRAM, then resolving 4x to 1x again, and again, would be pretty weird (and also not really possible with the current Vulkan renderer likely since that would involve reinterpreting color as depth). But well, if there's a valid mip address provided and the smallest mip is not 0, something has to be done of course.

goedzo:
@jgoyvaerts  pfew, one commit left :). Not an easy task haha.

@Triang3l yes I tested it, and gives no issues as we just apply the existing logic that was there to the other regions aswell.

goedzo:
@jgoyvaerts I hope you can help me, because everytime I want to update the my master version with the base master version, all kind of random commits appear in this one. I'd had to rebase my project already a couple of times to make only 1 commit in this pull. How can I avoid this PR to change whenever I want to get the latest commits from the base?

goedzo:
@jgoyvaerts ah so I need to create a PR from a branch, not merge them back first! Ok I see a lot of things go wrong by doing it this way, so I will save us all the trouble and close this one and do it again :). Thanks for all the help and sorry to cause all these issues!

PR #1276, opened 2018-11-21
[Vulkan] Allow depth buffers to have mips

goedzo:
Fixed in one single commit, tested and works.


d4rkne55:
I guess you should be squashing these three commits also into one; the last two are just polluting the commit history and not doing much by their own.
At least if you know to do it cleanly now.^^

goedzo:
Actually i don't know how to do this and in the end i needed to completely delete my own fork, recreate the file and the PR and then found out it failed because of two spaces, even though the xb format and xb lint gave the full ok on the code. It cost me 4 hours just to do this simple pr so i am done with it to be honest. The code is now here and works great so do whatever you want to do with it. I'm giving up on this worry about a couple of extra commits.

goedzo:
I created #1277 to check the space issues with xb format not being found. So the extra commits are also a way to test this properly.

goedzo:
Since I understand the cause, let me try to rebase this one, now I have a new branch.

goedzo:
Well I think this is on now, so what about merging it?

PR #1280, opened 2018-11-23
[third_party] WIP: Add pevents to third_party folder

chris-hawley:
This will be used for threading.


davidgumberg:
Premature?

chris-hawley:
This was really more to make sure nothing was broke before getting started. Both with my premake files and anything travis or appveyor might not like.

chris-hawley:
You're right. Per the discord discussion, pevents doesn't allow for waiting on other objects. Unless I wrap around pevents, then it won't suit our needs.

PR #1284, opened 2018-11-25
 Proper MMIO Handlers, Texture Utils, AluVectors and Shader Translators 

goedzo:
Allow implementation of MMIO Handlers, Texture Utils, AluVectors and Shader Translators.

### DIFF
--- a/src/xenia/cpu/mmio_handler.cc
+++ b/src/xenia/cpu/mmio_handler.cc
@@ -9,6 +9,9 @@
 
 #include "xenia/cpu/mmio_handler.h"
 
+#include <algorithm>
+#include <cstring>
+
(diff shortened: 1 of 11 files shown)


jgoyvaerts:
What are you trying to achieve by opening a pull request which merges changes from d3d12 to master?

goedzo:
Implementing vulcan improvements, or is vulcan no longer maintained in favor of d3d12?

Triang3l:
That will be merged when it's ready to be merged. Those implementations exist separately from the Vulkan version, it doesn't have the code that uses the new watches yet, this is why we still have two separate implementations (and precisely why I'm NOT merging d3d12 into master). We know our architecture better than you do, believe me.

PR #1286, opened 2018-11-25
Add config to Xenia [WIP]

jgoyvaerts:
This PR adds a config file (config.toml) to Xenia. It replaces the existing GFLAGS way of config loading.

This PR is still in progress, but is open for feedback as to the current way of working.

Topics of discussion:
- Location of config file (right now it's next to the exe, Microsoft suggest AppData somewhere, but I'm not a fan of that. Personal preference is next to the exe). We can't actually use the content_root specified by the user because that's specified in the config, and that leads to a chicken and egg problem
- GFLAGS: remove completely? Implement away to make them work together? Some GFLAGS (like "target" are not in the config (doesn't make sense to have that in the config) so if we want to get rid of GFLAGS we have to find an alternate way to implement flags such as target.
- Anything else


src/xenia/config.h
```diff
@@ -0,0 +1,42 @@
...
+class Config {
+ public:
+  static Config& GetInstance() { return (*instance_); }
+  static void Init(const std::wstring& content_path);
+  static void Shutdown();
+  void SaveConfig();
+  void RestoreDefaults();
+
+  // config variables
+  std::string gpu_;
```
Triang3l: Not really sure about having things like this public, usually getters are created for this. Still, why a class rather than a namespace?

src/xenia/config.cc
```diff
@@ -0,0 +1,101 @@
...
+  auto file_path = xe::join_paths(content_path, L"config.toml");
+  instance_ = new Config(file_path);
+}
+void Config::Shutdown() {
+  delete instance_;
+  instance_ = nullptr;
+}
+
+void Config::LoadDefaults() {
+  gpu_ = "any";
```
DrChat: This won't be sustainable with the amount of options Xenia will eventually have.
Plan to create a static registration system in the future, ala `ConVar` in Source Engine.

jgoyvaerts:
Also thoughts on using config friendly names (like InputSystem, GraphicsSystem) or just dropping hid, gpu in there (since there's a description anyway)

jgoyvaerts:
Pushed new changes, categories are back and now also supports game specific configs (even though Xenia should load components in a different order for that to be more effective).

Please leave feedback, especially on c++ stuff (any possibilities for macros to make this prettier).

Sample config:
```
#This is the Xenia config file, in toml format

[General]
AudioSystem = "any" 		#Audio system. Use: [any, nop, xaudio2]
ContentRoot = "" 		#Root path for content (save/etc) storage.
Fullscreen = "false" 		#Toggles fullscreen
GraphicsSystem = "any" 		#Graphics system. Use: [any, vulkan, null]
InputSystem = "any" 		#Input system. Use: [any, nop, winkey, xinput]
MountCache = "false" 		#Enable cache mount
MountScratch = "false" 		#Enable scratch mount
TimeScale = "1.0" 		#Scalar used to speed or slow time (1x, 2x, 1/2x, etc).
```

Sample game specific config: 4E4D07FC.cfg
`TimeScale = "3.0"`

goedzo:
Please don't remove GFLAGS as it is used by frontends like retrofe or hyperspin to control how the emulator behaves. It is harder to generate specific config files if you want to use different options per game (like gui renderer, or audio devie). Command line parameters are the most convenient method to use. If you are replacing GFLAGS, then please passthrough some command params to override things in the config file.

jgoyvaerts:
The goal isn't to get rid of commandline arguments, just replace GFLAGS with something better which ties into the config system (notice the config already has commandline arguments built in, they just need to be loaded). Potential alternative for GFLAGS is https://github.com/jarro2783/cxxopts or so

src/xenia/app/xenia_main.cc
```diff
@@ -39,26 +40,31 @@
 
 #include "third_party/xbyak/xbyak/xbyak_util.h"
 
-DEFINE_string(apu, "any", "Audio system. Use: [any, nop, xaudio2]");
-DEFINE_string(gpu, "any", "Graphics system. Use: [any, vulkan, null]");
-DEFINE_string(hid, "any", "Input system. Use: [any, nop, winkey, xinput]");
+CVar apu_("AudioSystem", "any", "Audio system. Use: [any, nop, xaudio2]", "General");
```
Wildenhaus: I suggest maybe [using a singleton](https://github.com/Wildenhaus/xenia/blob/qt-experimental/src/xenia/app/library/game_library.cc#L8) or some other global system for managing configuration variables and and registering callbacks (like Dr. Chat suggested earlier and you have already started towards implementing). That way it's self contained, can be referenced from anywhere, and you don't need to do checks/handling in `xenia_main.cc`. Those can go in their component-specific files.

Triang3l: Ugh, looks like we're going to confuse jgoyvaerts even more with contradicting suggestions :D But the choice depends primarily on whether some initialization code is required for the whole system, I think. Not very important probably though.

Wildenhaus: I overlooked the part where you're using a global map to store CVars, which is essentially what I was suggesting. Carry on 😛

jgoyvaerts:
Need another round of feedback, before I finish replacing every use of DEFINE_bool and other gflag initializations.

Some implemented features:
- CmdVar (command line variable) and CVar (config variable), every config value can be set through command line but command line can only be set through command line.
- Game specific config files. It's possible to create a config file in the form of <gameID>.cfg next to your xenia.cfg and it will load game specific settings
- variable names have been left to original code variable names (so apu instead of GraphicsSystem) as not to cause any breaking changes to current command line options.

Should I be worried about the performance of CVar->GetBool, GetFloat and GetInt in case someone calls these in a tight loop? Or is that the responsibility of the caller?

src/xenia/app/xenia_main.cc
```diff
@@ -37,26 +38,32 @@
...
 #endif  // XE_PLATFORM_WIN32
 
-DEFINE_string(apu, "any", "Audio system. Use: [any, nop, xaudio2]");
-DEFINE_string(gpu, "any", "Graphics system. Use: [any, vulkan, null]");
-DEFINE_string(hid, "any", "Input system. Use: [any, nop, winkey, xinput]");
+#include "third_party/xbyak/xbyak/xbyak_util.h"
 
-DEFINE_string(target, "", "Specifies the target .xex or .iso to execute.");
-DEFINE_bool(fullscreen, false, "Toggles fullscreen");
+CVar apu_("apu", "any", "Audio system. Use: [any, nop, xaudio2]", "General");
```
DrChat: Perhaps use a different variable naming convention - maybe `var_apu`?

src/xenia/config.h
```diff
@@ -0,0 +1,52 @@
+#ifndef XENIA_CONFIG_H_
+#define XENIA_CONFIG_H_
+#include <cpptoml/include/cpptoml.h>
+#include <string>
+
+class CmdVar {
+ public:
+  CmdVar(const char* name, const char* defaultValue, const char* description);
```
DrChat: Add a `flags` parameter to this constructor (default 0). We can define CVar flags such as one that says this CVar cannot be modified at runtime.
We can also add a CVar changed callback function parameter (once again, default NULL).

DrChat:
Nah - I don't imagine anyone will call getter functions in a tight loop. If they do, we can come back to it later.

I'm actually thinking this would work nicely with an internal console as well. Users could change CmdVars at runtime (if those particular vars support it).

PR #1289, opened 2018-11-26
Updated copyright dates to 2018

halotroop2288:
Changes nothing but comments.


Triang3l:
Don't need to replace almost all the relevant last change messages and dates in directories with one single commit IMO.

gibbed:
Yeah that's not happening.

PR #1298, opened 2018-12-02
Make sure the audio system is initialized when starting a file from the command line

goedzo:
When calling CreateMasteringVoice in xaudio2, this can crash when the windows comsystem is not initialized.

This will always happen when a game file is passed as an command parameter to the emulator.

Calling CoInitialize fixes this behaviour.


gibbed:
https://github.com/xenia-project/xenia/commit/18d8a59cfab41d5684f99ce62a644648e9ad3d59

PR #1301, opened 2018-12-05
Update README and issue template

Margen67:


README.md
```diff
@@ -2,17 +2,17 @@ Xenia - Xbox 360 Emulator Research Project
...
 
 Xenia is an experimental emulator for the Xbox 360. For more information see the
-[main xenia website](https://xenia.jp/).
+[main Xenia website](https://xenia.jp/).
 
 **Interested in supporting the core contributors?
 [Xenia Project on Patreon](https://www.patreon.com/xenia_project).**
 
-Come chat with us about **emulator-related topics** on [Discord](https://discord.gg/Q9mxZf9).
+Come chat with us about **emulator-related topics** on [Discord](https://discord.me/xenia-emulator).
```
DrChat: Why make this change?

README.md
```diff
@@ -80,17 +80,18 @@ Fixes and optimizations are always welcome (please!), but in addition to
 that there are some major work areas still untouched:
 
 * Help work through missing functionality/bugs in game [compat](https://github.com/xenia-project/xenia/issues?labels=compat)
-* Add input drivers for [PS4 controllers](https://github.com/xenia-project/xenia/issues/60) (or anything else)
+* Add input drivers for [DS4 controllers](https://github.com/xenia-project/xenia/issues/60) (or anything else)
```
DrChat: Although it's understandable to call these DualShock 4 controllers, less people are going to understand that these are the controllers for the PS4.

README.md
```diff
@@ -80,17 +80,18 @@ Fixes and optimizations are always welcome (please!), but in addition to
...
 * Skilled with Linux? A strong contributor is needed to [help with porting](https://github.com/xenia-project/xenia/labels/cross%20platform)
 
-See more projects [good for contributors](https://github.com/xenia-project/xenia/labels/good%20first%20issue). It's a good idea to ask on Discord/the bugs before beginning work
-on something.
+See more projects [good for contributors](https://github.com/xenia-project/xenia/labels/good%20first%20issue). It's a good idea to ask on Discord/the issues before beginning
```
DrChat: "...It's a good idea to ask on Discord/check the issues page before beginning..."

README.md
```diff
@@ -80,17 +80,18 @@ Fixes and optimizations are always welcome (please!), but in addition to
...
 
 For more see the main [frequently asked questions](https://xenia.jp/faq/) page.
 
 ### Can I get an exe?
 
-Check [Appveyor's artifacts](https://ci.appveyor.com/project/benvanik/xenia/build/artifacts) to see
-what's there.
+[Master (Vulkan)](https://ci.appveyor.com/api/projects/benvanik/xenia/artifacts/xenia-master.zip?branch=master&job=Configuration%3A%20Release&pr=false)
+
+[D3D12](https://ci.appveyor.com/api/projects/benvanik/xenia/artifacts/xenia-d3d12.zip?branch=d3d12&job=Configuration%3A%20Release&pr=false)
```
DrChat: Let's not link to the d3d12 branch from the master branch. The d3d12 branch could probably link to the d3d12 appveyor artifacts.

README.md
```diff
@@ -80,17 +80,18 @@ Fixes and optimizations are always welcome (please!), but in addition to
...
 
 ## FAQ
 
 For more see the main [frequently asked questions](https://xenia.jp/faq/) page.
 
 ### Can I get an exe?
 
-Check [Appveyor's artifacts](https://ci.appveyor.com/project/benvanik/xenia/build/artifacts) to see
-what's there.
+[Master (Vulkan)](https://ci.appveyor.com/api/projects/benvanik/xenia/artifacts/xenia-master.zip?branch=master&job=Configuration%3A%20Release&pr=false)
```
DrChat: This is a good idea, and could you change "Master (Vulkan)" -> "Master"?

README.md
gibbed: I think there's merit to this change as the server has a vanity link for that purpose (apparently, didn't know it did).

README.md
gibbed: I think that was changed to match how I labeled them on the website?

DrChat: Ah, good point. I suppose we can keep it this way, though I don't like to think of branches as being tied to any one renderer backend.

gibbed: That's just a title, doesn't necessarily have to be renderer.

https://github.com/xenia-project/xenia/blob/gh-pages/_config.yml#L66

README.md
Margen67: What do you mean?

README.md
Margen67: People download from the master branch and get confused when it's not d3d12 so that's not a good idea.

README.md
Margen67: Does PS4/DS4 work?

README.md
gibbed: Please undo this change. I didn't notice the `.me` TLD change and this page has ads.

README.md
```diff
@@ -82,17 +82,17 @@ is wide open greenfield fun.
 Fixes and optimizations are always welcome (please!), but in addition to
 that there are some major work areas still untouched:
 
-* Help work through [missing functionality/bugs in games](https://github.com/xenia-project/xenia/labels/compat)
-* Add input drivers for [PS4 controllers](https://github.com/xenia-project/xenia/issues/60) (or anything else)
+* Help work through [missing functionality/bugs in games](https://github.com/xenia-project/xenia/issues?labels=compat)
+* Add input drivers for [PS4/DS4 controllers](https://github.com/xenia-project/xenia/issues/60) (or anything else)
```
gibbed: I think for this, the text would be best `DualShock4 (PS4) controllers`.

Issue #1154, opened 2018-05-14
Automatically compress log file


zemzema:
It would be a good option that on the end Xenia automatically compress log file that can be submitted to the compatibility list issue page.

brodycas3:
And upload to paste.ee or similar service?

TGLTClutch:
Would be good if Xenia automatically submitted the compressed log to the correct issue

techflashYT:
What type of format would be best for this?  targz?  tarxz?  zip?   7z?  RAR?   In order for this to get done, this would need to be decided on.  I would be glad to implement this once it actually gets decided.

Nonary:
Sounds like it would be an easy first issue, but the problem I see here is that in order for it to be useful, it needs to be done in real time. 

> What type of format would be best for this? targz? tarxz? zip? 7z? RAR? In order for this to get done, this would need to be decided on. I would be glad to implement this once it actually gets decided.

It's a bit more challenging than it seems in my opinion, because you would need to compress the file in real time so to speak. You can't just have it auto zip something and then delete the original log file, because it would add a considerable delay to exiting. 

However, as a suggestion I would say .ZIP should be decided here since it has native support in practically every operating system and does pretty well with compressing text. I am open for ideas on how this should be done as well.

techflashYT:
Alright, well if it's ZIP, I have kind of a suggestion on how it should work.  A kind of checklist of possible ways.  Here's some psuedocode:
```
#ifdef __linux // Linux
    if (gzip exists) {
        gzip(log)
        return
    }
    else if (tar exists) {
        tar(log)
        return
    }
    else {
        compressUsingBuiltInMethodFromLibrary(log)
        return
    }
#elifdef WIN32 // Windows
    if (7zip) {
        7zip(log, "zip")
        return
    }
    else if (winrar) {
        winrar(log, "zip")
        return
    }
    else if (builtInWindowsZipTool) {
        builtInWindowsZipTool(log)
        return
    }
    else if (powershell) {
        powershellWithCompress-ArchiveArgumen(log)
        return
    }
    else {
        compressUsingBuiltInMethodFromLibrary(log)
        return
   }
#else // Something else, we don't know what it has, so just use builtin method
    compressUsingBuiltinMethodFromLibrary(log)
#endif
```

PR #1302, opened 2018-12-06
[Base] Logs stored in folder with timestamp.

Unnoen:
Custom log flag untouched for now.

Planning on tackling https://github.com/xenia-project/xenia/issues/1154 in time.
This is my first ever foray into C++ so I might need some hand holding.

### DIFF
--- a/src/xenia/base/logging.cc
+++ b/src/xenia/base/logging.cc
@@ -52,9 +52,15 @@ thread_local std::vector<char> log_format_buffer_(64 * 1024);
 class Logger {
  public:
   explicit Logger(const std::wstring& app_name) : running_(true) {
+	//Create timestamp suffix for filename use.
+    time_t timeraw = time(nullptr);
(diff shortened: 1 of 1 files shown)


Razzile:
This could use up _a lot_ of storage after continued use. May be better waiting until the new UI is finished and have it as an option. Or as a command line flag

Unnoen:
The current implementation also uses a lot of storage, this just splits it up for easier debugging.

Razzile:
the current implementation clears the existing log on launch. this would cause new logs to be made on each run

Unnoen:
Huh, so it does, I thought it just kept appending...
I might make this keep only X amount of logs then. Once it hits that amount it will delete the oldest.

gibbed:
@Unnoen Don't worry, this is a good idea _in concept_ but I agree with Razzile that it should wait until we have a better UI that can allow for customization of these logs.

PR #1306, opened 2018-12-10
Change "Windows SDKs" to specify exact SDKs needed

twrist:
_"I'm helping!" ~Ralph Wiggum_


Margen67:
It should also mention that building the D3D12 branch requires the 15063 SDK.

gibbed:
No longer relevant since support for VS 2019 was added.

PR #1308, opened 2018-12-13
Update D3D12

YeimiK:
Update


Triang3l:
Only GitHub web interface changes there that affect nothing when on a separate branch, not very important.

Issue #399, opened 2015-08-18
Add wstring<->string conversion compat shim.


benvanik:
GCC's stdlib doesn't have codecvt, while clang and msvc do. Currently this is hacked to just truncate wstring, but that's not right and should be fixed using linux-specific APIs (until GCC gets its shit together): [string.cc](https://github.com/benvanik/xenia/blob/master/src/xenia/base/string.cc#L14)

PlasmaPower:
It looks like libiconv is the best choice for this. However, it is LGPG ([source](https://www.gnu.org/software/libiconv/) "The `libiconv` and `libcharset` *libraries* and their header files are under LGPL."). I'm not sure if it is too complicated (?), the documentation on that in `CONTRIBUTING.md` looks like it was not completed.

benvanik:
We wouldn't ship the source or headers, but could rely on them being installed (apt-get install libconv...). Could have xenia-build verify its installed in its setup command (with dpkg-query or something).

PlasmaPower:
That is what I was expecting, but I'm no expert in licensing. I'll see if I can implement that tomorrow.

PlasmaPower:
Side note, for checking if they are installed, I wouldn't rely on dpkg-query because a lot of systems aren't dpkg based. Just checking for the library should work (with tools like `locate` or `whereis`).

Elronnd:
Generally, the way this is done is with `pkg-config`.

PR #1316, opened 2019-01-13
[Base] Remove hacky fix for missing codecvt on linux

chris-hawley:
per #399 , these lines were only required because codecvt was missing in GCC. That has since been fixed per https://stackoverflow.com/questions/15615136/is-codecvt-not-a-std-header

### DIFF
--- a/src/xenia/base/filesystem_posix.cc
+++ b/src/xenia/base/filesystem_posix.cc
@@ -12,29 +12,58 @@
 #include "xenia/base/logging.h"
 #include "xenia/base/string.h"
 
+#include <assert.h>
 #include <dirent.h>
 #include <fcntl.h>
 #include <ftw.h>
+#include <libgen.h>
+#include <pwd.h>
+#include <stdio.h>
+#include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <iostream>
 
 namespace xe {
 namespace filesystem {
 
 std::wstring GetExecutablePath() {
-  assert_always();  // IMPLEMENT ME.
-  return std::wstring();
+  char buff[FILENAME_MAX] = "";
+  readlink("/proc/self/exe", buff, FILENAME_MAX);
+  std::string s(buff);
+  return to_wstring(s);
 }
 
 std::wstring GetExecutableFolder() {
-  assert_always();  // IMPLEMENT ME.
-  return std::wstring();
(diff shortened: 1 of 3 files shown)


arabek:
Shouldn't this be hidden behind a gcc version check, since gcc versions prior to 5.x don't support it?

Rationale: https://gcc.gnu.org/gcc-5/changes.html
```
Full support for C++11, including the following new features:
(…)
locale facets for Unicode conversion;
```

chris-hawley:
I fixed it so it should detect the right compiler versions.

Prism019:
Shouldn't we be using the contents of the `XDG_DATA_HOME` environment variable + `/xenia`, then falling back to `~/.xenia`? `~/.local/share` _is_ the default content of `XDG_DATA_HOME`, but users are free to change it at will, and programs should respect that. Much how Rick made the decision to change to `Documents\Xenia` on Windows due to Microsoft's standards, I feel that we should be using Freedesktop's standards on Linux.

[Freedesktop's Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
[Other examples of programs using what I've specified](https://wiki.archlinux.org/index.php/XDG_Base_Directory#Supported)

src/xenia/base/filesystem_posix.cc
```diff
@@ -15,26 +15,50 @@
...
+  if (dataHome == NULL) {
+    dataHome = std::getenv("HOME");
+  } else {
+    std::string home(dataHome);
+    return to_wstring(home);
+  }
+
+  // if HOME not set, fall back to this
+  if (dataHome == NULL) {
+    struct passwd* pw = getpwuid(getuid());
```
DrChat: You might use `getpwuid_r` for threadsafety and easy error checking.

src/xenia/base/string.cc
```diff
@@ -9,12 +9,28 @@
...
+#endif
+#elif defined(__GNUC__) || defined(__GNUG__)
+// using gcc
+#if (__GNUC__ < 5)
+// insufficient clang version
+#define NO_CODECVT 1
+#else
+#include <codecvt>
+#endif
+#elif defined( \
```
DrChat: Could you drop the comment onto a different line so this `#elif` is not split?

chris-hawley: oh...that was split by the auto formatting.

Issue #24, opened 2013-05-23
Add libjit overflow add/sub/etc that don't throw exceptions.


benvanik:
Getting back a value that indicates an overflow would produce much better code.

Issue #23, opened 2013-05-23
libjit may be storing useless local variables.


benvanik:
00000000000F1167 48 89 45 E0          mov         qword ptr [rbp-20h],rax  
00000000000F116B 0F C8                bswap       eax  

The stored value is never used again. Something is making libjit think it needs to keep it around in a local.

PR #1317, opened 2019-01-18
Threading unit tests and linux threading implementation

bwrsandman:
Current Status on master
---
On Linux, the Xenia app and many of the test ui tests do not run correctly due to waits on `WaitHandles` flowing through and returning early.

Before

Fix status of this PR
---
I have been working on this PR for about a year and it now has reached a point where I consider it done and ready for review.

This PR implements most of `threading_posix.cc` by TDD to make sure behaviour is the same as win32.
This is a follow-up to (#1108) in order to get xenia to actually run on Linux. Following this, a few GTK-Vulkan fixes need to be done as well as implementing POSIX file system stubs. With those, a GTK window actually appears and responds to events.

After

Fix description
---
Borrowed commit from #1076 for Linux debug symbols.

Fixed occurrences of micro/millisecond confusion on timers.

Posix threads do not support suspension in the same way as windows threads. [Real-time interrupts used to give this functionality](https://stackoverflow.com/a/7979922) and to implement thread callback when in alertable state.
More real-time interrupts used to signal timers.
Since there are more than two signals used, `SIGUSR1` and `SIGUSR2` are not used. Signals are allocated between `SIGRTMIN` and `SIGRTMAX` which are dynamic and can change per machine or kernel version but should start from `SIGUSR2 + 1`.

> Not all systems support real-time signals.  bits/signum.h indicates
> that they are supported by overriding __SIGRTMAX to a value greater
> than __SIGRTMIN.  These constants give the kernel-level hard limits,
> but some real-time signals may be used internally by glibc.  Do not
> use these constants in application code; use SIGRTMIN and SIGRTMAX
> (defined in signal.h) instead.
```c
/* Return number of available real-time signal with highest priority.  */
extern int __libc_current_sigrtmin (void) __THROW;
/* Return number of available real-time signal with lowest priority.  */
extern int __libc_current_sigrtmax (void) __THROW;

#define SIGRTMIN        (__libc_current_sigrtmin ())
#define SIGRTMAX        (__libc_current_sigrtmax ())
```

Real-time interrupts handlers are installed in a lazy-load fashion. This could be moved to an initialize phase to avoid lazy-load, but no obvious location for this was found.

Waiting is implemented using static condition variables. To do this, each `WaitHandle` is also an instance of `PosixConditionBase`, the latter which has the same functionality of wait and wait multiple. This implementation was more successful than using the existing `PosixFdHandle` which was then removed in order to use a more generalizable super class which gave the added bonus of waiting on multiple types of primitives at the same time since they all use the same conditional variables.

Alertable state is set as a thread local static variable.

I ran these tests multiple times concurrently in Windows and Linux on my machine, my laptop and on travis. I will also test it on my LattePanda Alpha SBC to test a lower-end machine. Throughout the development, there have been fails which are not always reproducible and I brought this number down with fixes and hopefully there aren't any rare bugs left.

Things left to check later
---
~~Suspending an already suspended thread was not tested nor implemented and left for a later time.~~

Implementing affinity and priority was attempted but in the end abandoned due to Linux's technicalities.
I could be wrong but there doesn't seem to be any way to set a thread's priority which is higher than the main thread's without running as a super user so we could only set it to Normal or Lower.

Sleeping for times lower than 1ms works on my setups but causes deadlocks on Travis. This is no problem for the unit tests. The times were just increased by factors of 10 to 100. This, however, means that tests which run in about 5s could potentially run in < 1s if that deadlock issue is fixed.

There will always be fundamental differences between Windows and Linux which could be tested against. My hope is that the existence of these unit tests can make pin pointing these differences more easy.

### DIFF
--- a/src/xenia/base/threading_posix.cc
+++ b/src/xenia/base/threading_posix.cc
@@ -164,75 +164,64 @@ std::unique_ptr<HighResolutionTimer> HighResolutionTimer::CreateRepeating(
   return std::unique_ptr<HighResolutionTimer>(timer.release());
 }
 
-// TODO(dougvj) There really is no native POSIX handle for a single wait/signal
-// construct pthreads is at a lower level with more handles for such a mechanism
-// This simple wrapper class could function as our handle, but probably needs
-// some more functionality
+// There really is no native POSIX handle for a single wait/signal construct
+// pthreads is at a lower level with more handles for such a mechanism.
+// This simple wrapper class functions as our handle and uses conditional
+// variables for waits and signals.
 class PosixCondition {
  public:
-  PosixCondition() : signal_(false) {
-    pthread_mutex_init(&mutex_, NULL);
-    pthread_cond_init(&cond_, NULL);
-  }
-
-  ~PosixCondition() {
-    pthread_mutex_destroy(&mutex_);
-    pthread_cond_destroy(&cond_);
-  }
+  PosixCondition(bool manual_reset, bool initial_state)
+      : signal_(initial_state), manual_reset_(manual_reset) {}
+  virtual ~PosixCondition() = default;
 
   void Signal() {
-    pthread_mutex_lock(&mutex_);
+    auto lock = std::unique_lock<std::mutex>(mutex_);
     signal_ = true;
-    pthread_cond_broadcast(&cond_);
-    pthread_mutex_unlock(&mutex_);
+    if (manual_reset_) {
+      cond_.notify_all();
+    } else {
+      cond_.notify_one();
+    }
   }
 
   void Reset() {
-    pthread_mutex_lock(&mutex_);
+    auto lock = std::unique_lock<std::mutex>(mutex_);
     signal_ = false;
-    pthread_mutex_unlock(&mutex_);
   }
 
-  bool Wait(unsigned int timeout_ms) {
-    // Assume 0 means no timeout, not instant timeout
-    if (timeout_ms == 0) {
-      Wait();
+  WaitResult Wait(std::chrono::milliseconds timeout) {
+    bool executed;
+    auto predicate = [this] { return this->signaled(); };
+    auto lock = std::unique_lock<std::mutex>(mutex_);
+    if (predicate()) {
+      executed = true;
+    } else {
+      if (timeout == std::chrono::milliseconds::max()) {
+        cond_.wait(lock, predicate);
+        executed = true;  // Did not time out;
+      } else {
+        executed = cond_.wait_for(lock, timeout, predicate);
+      }
     }
-    struct timespec time_to_wait;
-    struct timeval now;
-    gettimeofday(&now, NULL);
-
-    // Add the number of seconds we want to wait to the current time
-    time_to_wait.tv_sec = now.tv_sec + (timeout_ms / 1000);
-    // Add the number of nanoseconds we want to wait to the current nanosecond
-    // stride
-    long nsec = (now.tv_usec + (timeout_ms % 1000)) * 1000;
-    // If we overflowed the nanosecond count then we add a second
-    time_to_wait.tv_sec += nsec / 1000000000UL;
-    // We only add nanoseconds within the 1 second stride
-    time_to_wait.tv_nsec = nsec % 1000000000UL;
-    pthread_mutex_lock(&mutex_);
-    while (!signal_) {
-      int status = pthread_cond_timedwait(&cond_, &mutex_, &time_to_wait);
-      if (status == ETIMEDOUT) return false;  // We timed out
+    if (executed) {
+      post_execution();
+      return WaitResult::kSuccess;
+    } else {
+      return WaitResult::kTimeout;
     }
-    pthread_mutex_unlock(&mutex_);
-    return true;  // We didn't time out
-  }
-
-  bool Wait() {
-    pthread_mutex_lock(&mutex_);
-    while (!signal_) {
-      pthread_cond_wait(&cond_, &mutex_);
-    }
-    pthread_mutex_unlock(&mutex_);
-    return true;  // Did not time out;
   }
 
  private:
+  inline bool signaled() const { return signal_; }
+  inline void post_execution() {
+    if (!manual_reset_) {
+      signal_ = false;
+    }
+  }
   bool signal_;
-  pthread_cond_t cond_;
-  pthread_mutex_t mutex_;
+  const bool manual_reset_;
+  std::condition_variable cond_;
+  std::mutex mutex_;
 };
 
 // Native posix thread handle
@@ -250,12 +239,14 @@ class PosixThreadHandle : public T {
   pthread_t handle_;
 };
 
-// This is wraps a condition object as our handle because posix has no single
+// This wraps a condition object as our handle because posix has no single
 // native handle for higher level concurrency constructs such as semaphores
 template <typename T>
 class PosixConditionHandle : public T {
  public:
-  ~PosixConditionHandle() override {}
+  PosixConditionHandle(bool manual_reset, bool initial_state)
+      : handle_(manual_reset, initial_state) {}
+  ~PosixConditionHandle() override = default;
 
  protected:
   void* native_handle() const override {
@@ -265,51 +256,10 @@ class PosixConditionHandle : public T {
   PosixCondition handle_;
 };
 
-template <typename T>
-class PosixFdHandle : public T {
- public:
-  explicit PosixFdHandle(intptr_t handle) : handle_(handle) {}
-  ~PosixFdHandle() override {
-    close(handle_);
-    handle_ = 0;
-  }
-
- protected:
-  void* native_handle() const override {
-    return reinterpret_cast<void*>(handle_);
-  }
-
-  intptr_t handle_;
-};
-
-// TODO(dougvj)
 WaitResult Wait(WaitHandle* wait_handle, bool is_alertable,
                 std::chrono::milliseconds timeout) {
-  intptr_t handle = reinterpret_cast<intptr_t>(wait_handle->native_handle());
-
-  fd_set set;
-  struct timeval time_val;
-  int ret;
-
-  FD_ZERO(&set);
-  FD_SET(handle, &set);
-
-  time_val.tv_sec = timeout.count() / 1000;
-  time_val.tv_usec = timeout.count() * 1000;
-  ret = select(handle + 1, &set, NULL, NULL, &time_val);
-  if (ret == -1) {
-    return WaitResult::kFailed;
-  } else if (ret == 0) {
-    return WaitResult::kTimeout;
-  } else {
-    uint64_t buf = 0;
-    ret = read(handle, &buf, sizeof(buf));
-    if (ret < 8) {
-      return WaitResult::kTimeout;
-    }
-
-    return WaitResult::kSuccess;
-  }
+  auto handle = reinterpret_cast<PosixCondition*>(wait_handle->native_handle());
+  return handle->Wait(timeout);
 }
 
 // TODO(dougvj)
@@ -329,40 +279,37 @@ std::pair<WaitResult, size_t> WaitMultiple(WaitHandle* wait_handles[],
   return std::pair<WaitResult, size_t>(WaitResult::kFailed, 0);
 }
 
-// TODO(dougvj)
-class PosixEvent : public PosixFdHandle<Event> {
+class PosixEvent : public PosixConditionHandle<Event> {
  public:
-  PosixEvent(intptr_t fd) : PosixFdHandle(fd) {}
+  PosixEvent(bool manual_reset, bool initial_state)
+      : PosixConditionHandle(manual_reset, initial_state) {}
   ~PosixEvent() override = default;
-  void Set() override {
-    uint64_t buf = 1;
-    write(handle_, &buf, sizeof(buf));
+  void Set() override { handle_.Signal(); }
+  void Reset() override { handle_.Reset(); }
+  void Pulse() override {
+    using namespace std::chrono_literals;
+    handle_.Signal();
+    MaybeYield();
+    Sleep(10us);
+    handle_.Reset();
   }
-  void Reset() override { assert_always(); }
-  void Pulse() override { assert_always(); }
-
- private:
-  PosixCondition condition_;
 };
 
 std::unique_ptr<Event> Event::CreateManualResetEvent(bool initial_state) {
-  // Linux's eventfd doesn't appear to support manual reset natively.
-  return nullptr;
+  return std::make_unique<PosixEvent>(true, initial_state);
 }
 
 std::unique_ptr<Event> Event::CreateAutoResetEvent(bool initial_state) {
-  int fd = eventfd(initial_state ? 1 : 0, EFD_CLOEXEC);
-  if (fd == -1) {
-    return nullptr;
-  }
-
-  return std::make_unique<PosixEvent>(fd);
+  return std::make_unique<PosixEvent>(false, initial_state);
 }
 
 // TODO(dougvj)
 class PosixSemaphore : public PosixConditionHandle<Semaphore> {
  public:
-  PosixSemaphore(int initial_count, int maximum_count) { assert_always(); }
+  PosixSemaphore(int initial_count, int maximum_count)
+      : PosixConditionHandle(false, false) {
+    assert_always();
+  }
   ~PosixSemaphore() override = default;
   bool Release(int release_count, int* out_previous_count) override {
     assert_always();
@@ -378,7 +325,9 @@ std::unique_ptr<Semaphore> Semaphore::Create(int initial_count,
 // TODO(dougvj)
 class PosixMutant : public PosixConditionHandle<Mutant> {
  public:
-  PosixMutant(bool initial_owner) { assert_always(); }
+  PosixMutant(bool initial_owner) : PosixConditionHandle(false, false) {
+    assert_always();
+  }
   ~PosixMutant() = default;
   bool Release() override {
     assert_always();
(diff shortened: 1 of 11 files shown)


Razzile:
Hot damn this is amazing work!

bwrsandman:
Note to reviewers:
Please let me know if there are any changes to be made regarding architecture, code style, naming, spelling or if there are any tests to add. 

If you check out the branch and run into issues, please send me a coredump if one is produced. Let me know what distro you run, which kernel version, the compiler with version, machine specs (cpu, number of cores, memory, load) and reproduceability. Having all this information is incredibly useful to find and fix threading bugs. 

Any comment is appreciated and I will try to make requested changes within a week.

chris-hawley:
First of all. This is excellent work. This has been a long-standing missing component of the emulator. 

I cloned your repository and then rebased it against the master for the latest fixes. I implemented the filesystem headers required and tried running it. I had some issues with it crashing and ran it under valgrind a few times and noticed this. To be clear, the crash wasn't caused by this output. The vulkan backend was crashing. Given my troubles with Vulkan, could you verify this behavior and fix if necessary? I have also added the three functions that were required to get the filesystem working.  

````
==15421== Thread 2 :
==15421== Invalid read of size 1
==15421==    at 0x201FB4: xe::threading::PosixCondition<xe::threading::Thread>::ThreadStartRoutine(void*) (threading_posix.cc:906)
==15421==    by 0x55BEA9C: start_thread (in /usr/lib/libpthread-2.28.so)
==15421==    by 0x5D98B22: clone (in /usr/lib/libc-2.28.so)
==15421==  Address 0x87beab0 is 32 bytes inside a block of size 48 free'd
==15421==    at 0x48389AB: free (vg_replace_malloc.c:530)
==15421==    by 0x201F73: xe::threading::PosixCondition<xe::threading::Thread>::ThreadStartRoutine(void*) (threading_posix.cc:901)
==15421==    by 0x55BEA9C: start_thread (in /usr/lib/libpthread-2.28.so)
==15421==    by 0x5D98B22: clone (in /usr/lib/libc-2.28.so)
==15421==  Block was alloc'd at
==15421==    at 0x483777F: malloc (vg_replace_malloc.c:299)
==15421==    by 0x569AE19: operator new(unsigned long) (in /usr/lib/libc++.so.1.0)
==15421==    by 0x203EFA: xe::threading::PosixThread::Initialize(xe::threading::Thread::CreationParameters, std::function<void ()>) (threading_posix.cc:836)
==15421==    by 0x20227A: xe::threading::Thread::Create(xe::threading::Thread::CreationParameters, std::function<void ()>) (threading_posix.cc:938)
==15421==    by 0x1ECA0D: xe::Logger::Logger(std::__cxx11::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> > const&) (logging.cc:71)
==15421==    by 0x1EBDF2: xe::InitializeLogging(std::__cxx11::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> > const&) (logging.cc:245)
==15421==    by 0x1C9AFE: main (main_posix.cc:37)
````

**Comment by gibbed** [comment]
Consider this high on my priority list, I'll review it as soon as I am able. Great work!

**Comment by bwrsandman** [comment]
@chris-hawley I plan to tackle vulkan on linux next.
I have only spent a little time looking at the issues with vulkan (to get xenia-app running) as most of my time was spent with the unit tests.

Other than the filesystem stubs which are easy to fix with hard-coded values, I identified two issues :
* The vulkan GTK windows is not setup correctly
** The xcb surface requirement isn't asked for
** The xcb connection/native_platform_handle isn't set
* The VkDebugReportCallbackCreateInfoEXT isn't properly set (32a0e25fd01156da0bc3382ae0941a328cde5aae)

I am pretty familiar with Vulkan and graphics so I plan on taking a more in-depth look at the implementation and following up with fixes in another pull request.

**Comment by chris-hawley** [comment]
I'll create a branch called linux-testing in my xenia fork.

**Comment by bwrsandman** [comment]
Great. By the way, you can see my Vulkan work and changes on [this branch](https://github.com/bwrsandman/xenia/tree/linux_vulkan). Subject to change.

**Comment by chris-hawley** [comment]
branch is all set to go

**Comment by bwrsandman** [comment]
@chris-hawley this is the crash I get when running your branch:

```
           PID: 15353 (xenia)
           UID: 1000 (~)
           GID: 1000 (~)
        Signal: 11 (SEGV)
     Timestamp: Fri 2019-01-18 21:25:23 EST (37s ago)
  Command Line: ./build/bin/Linux/Debug/xenia --log_file stdout
    Executable: ~/code/xenia/build/bin/Linux/Debug/xenia
 Control Group: /user.slice/user-1000.slice/session-2.scope
          Unit: session-2.scope
         Slice: user-1000.slice
       Session: 2
     Owner UID: 1000 (~)
       Boot ID: 1b49fb70c0384b7b91b899362bcedca8
    Machine ID: b9db9b094d524ac3a55b543bb2a09cfa
      Hostname: BattleStation
       Storage: /var/lib/systemd/coredump/core.xenia.1000.1b49fb70c0384b7b91b899362bcedca8.15353.1547864723000000.lz4
       Message: Process 15353 (xenia) of user 1000 dumped core.
                
                Stack trace of thread 15357:
                #0  0x0000000000000000 n/a (n/a)
                #1  0x000055fd3633a8e7 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #2  0x000055fd361435fc n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #3  0x000055fd361433dd n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #4  0x000055fd35fb34fe n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #5  0x000055fd3630bcaf n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #6  0x000055fd3630bb5d n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #7  0x000055fd35fb34fe n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #8  0x000055fd3630cfe5 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #9  0x000055fd3630c45d n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #10 0x00007f80d958cb7b n/a (libgdk-3.so.0)
                #11 0x00007f80d90d38d1 g_main_context_dispatch (libglib-2.0.so.0)
                #12 0x00007f80d90d55e9 n/a (libglib-2.0.so.0)
                #13 0x00007f80d90d65c2 g_main_loop_run (libglib-2.0.so.0)
                #14 0x00007f80d989b93f gtk_main (libgtk-3.so.0)
                #15 0x000055fd3630c3c1 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #16 0x000055fd3630cc9c n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #17 0x000055fd3630cbed n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #18 0x000055fd3630cb7d n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #19 0x000055fd3630cb55 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #20 0x000055fd3630cb25 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #21 0x000055fd3630ca19 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #22 0x00007f80d8bdd063 execute_native_thread_routine (libstdc++.so.6)
                #23 0x00007f80d904ea9d start_thread (libpthread.so.0)
                #24 0x00007f80d88b7b23 __clone (libc.so.6)
                
                Stack trace of thread 15353:
                #0  0x00007f80d9054afc pthread_cond_wait@@GLIBC_2.3.2 (libpthread.so.0)
                #1  0x00007f80d8bd6cd1 __gthread_cond_wait (libstdc++.so.6)
                #2  0x000055fd35fc74ce n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #3  0x000055fd35fc66f1 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #4  0x000055fd3630b8db n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #5  0x000055fd36141cfe n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #6  0x000055fd36159dd5 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #7  0x000055fd35fbf748 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #8  0x000055fd35f6dbcf n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #9  0x000055fd35f77b21 n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
                #10 0x00007f80d87e0223 __libc_start_main (libc.so.6)
                #11 0x000055fd35f5c38e n/a (~/code/xenia/build/bin/Linux/Debug/xenia)
```

This is due to the xcb connection/native_platform_handle not being set (0e5af7d) and the crash happens within either gtk or the graphics driver (in my case nvidia).

If you cherry-pick a9c2053 and 0e5af7d from my branch, you won't get those issues.

**Comment by chris-hawley** [comment]
that fixed the vulkan crashing.

**Comment by chris-hawley** [comment]
Also, we have a discord if you would rather do this back and forth there. [discord](https://discord.gg/5g93S8H)

**Comment by Triang3l** [comment]
`GetUserFolder` should have the following logic (@Prism019's comment on #1316 is related to this):

- If environment variable XDG_DATA_HOME is set, the path is `$XDG_DATA_HOME/Xenia`
- If it's not set:
    - If environment variable HOME is set, the path is `$HOME/.local/share/Xenia`
    - Optionally (I don't know how realistic this situation is, other apps don't always handle it), if HOME is not set, use `getpwuid(getuid())->pw_dir` as $HOME (this is not always the same as $HOME — the user may override the home directory, but $HOME takes precedence in this case).

**Comment by bwrsandman** [comment]
I fixed some memory leaks that were pointed out by valgrind.
They occurred due to threads not being joined after being cancels as in the case of `Terminate()`.
There was a unit test not joining with threads.
There was a case of use after delete of thread start_data.

File: src/xenia/base/threading_posix.cc [hunk 1]
```diff
@@ -48,265 +97,697 @@ void MaybeYield() {
...
+    std::unique_lock<std::mutex> lock(PosixConditionBase::mutex_);
+
+    if (aggregate()) {
+      executed = true;
+    } else {
+      if (timeout == std::chrono::milliseconds::max()) {
+        PosixConditionBase::cond_.wait(lock, aggregate);
+        executed = true;
+      } else {
+        executed = PosixConditionBase::cond_.wait_for(lock, timeout, aggregate);
```
DrChat: Might change this to:
```executed = PosixConditionBase::cond_.wait_for(lock, timeout, aggregate) == std::cv_status::no_timeout;```

File: premake5.lua [hunk 2]
```diff
@@ -68,6 +68,11 @@ filter({"configurations:Debug", "platforms:Windows"})
     "/NODEFAULTLIB:MSVCRTD",
   })
 
+filter({"configurations:Debug", "platforms:Linux"})
+  buildoptions({
+    "-g",
```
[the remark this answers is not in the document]
bwrsandman: It was not.

bwrsandman:
Thanks for the review @DrChat. I have made the requested changes.
Have you had a chance to run it yourself?
Please let me know if there's anything else to do.

src/xenia/base/threading_posix.cc
bwrsandman: That wouldn't work. `std::condition_variable::wait_for` only returns a `std::cv_status` if there is no predicate. If there is one, the wait returns a `bool`. This case needs a predicate.

DrChat: Ah, you're right - indeed it does.

DrChat:
So, I didn't merge this at the time because of the complex nature of this pull request - it's modifying a delicate part of the emulator, which can introduce very subtle bugs. 

The changes here look pretty well thought out though. One thing though: What about that deadlock on Travis from the unit tests?

bwrsandman:
```c++
// TODO(bwrsandman): due_times of under 1ms deadlock under travis
```
This was something which only happened on Travis. I wasn't able to get to the bottom of this since none of my machines could replicate. If I had to guess it's the nature of the container and the kernel used is somehow interpreting less than 1 ms as infinite.
I would love to have this fixed as it could speed up the tests by up to 1000x (switching ms to ns).
Unfortunately, since I can't replicate it outside of travis, it's difficult.

bwrsandman:
Rebased to latest master

bwrsandman:
Suspending already suspended threads is implemented and tested.
Native handles have been reworked to work better with stack walker #1405

bwrsandman:
@DrChat With the last commit, I reduced wait times by about 10x. Have some microseconds and nanoseconds times and no deadlock.

bwrsandman:
It would seem that `Events` have a bug when they're created as auto reset. More tests are required to have feature parity with Win32.

For example, launching xenia and quitting without loading any ROM will produce a deadlock.

The following fixes this deadlock but renders the behaviour the same as as manual reset event.

```diff
diff --git a/src/xenia/base/threading_posix.cc b/src/xenia/base/threading_posix.cc
index 216d6714..9f1334e5 100644
--- a/src/xenia/base/threading_posix.cc
+++ b/src/xenia/base/threading_posix.cc
@@ -284,7 +284,7 @@ class PosixCondition<Event> : public PosixConditionBase {
     if (manual_reset_) {
       cond_.notify_all();
     } else {
-      cond_.notify_one();
+      cond_.notify_all();
     }
     return true;
   }
```

q4a:
Do you know that there is tests with errors: 3 on Linux and 1 on Windows
Ubuntu 19.10:
```
$ ./build/bin/Linux/Release/xenia-base-tests

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
xenia-base-tests is a Catch v1.12.2 host application.
Run with -? for options

-------------------------------------------------------------------------------
HighResolutionTimer
-------------------------------------------------------------------------------
../src/xenia/base/testing/threading_test.cc:147
...............................................................................

../src/xenia/base/testing/threading_test.cc:193: FAILED:
  REQUIRE( counter1 <= ratio1 + 1 )
with expansion:
  12 <= 6

-------------------------------------------------------------------------------
Wait on Multiple Timers
-------------------------------------------------------------------------------
../src/xenia/base/testing/threading_test.cc:687
...............................................................................

../src/xenia/base/testing/threading_test.cc:717: FAILED:
  REQUIRE( any_result.second == 0 )
with expansion:
  1 == 0

-------------------------------------------------------------------------------
Test Thread QueueUserCallback
-------------------------------------------------------------------------------
../src/xenia/base/testing/threading_test.cc:889
...............................................................................

../src/xenia/base/testing/threading_test.cc:936: FAILED:
  REQUIRE( is_modified == 0 )
with expansion:
  -1 == 0

===============================================================================
test cases:  43 |  40 passed | 3 failed
assertions: 411 | 408 passed | 3 failed
```
Windows: https://ci.appveyor.com/project/q4a/xenia/builds/32266275/job/g2vonuy266aofj1g#L1185
```
xb test --config=%CONFIGURATION% --no_build
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
p�ܧ� is a Catch v1.12.2 host application.
Run with -? for options
-------------------------------------------------------------------------------
HighResolutionTimer
-------------------------------------------------------------------------------
C:\projects\xenia\src\xenia\base\testing\threading_test.cc(147)
...............................................................................
C:\projects\xenia\src\xenia\base\testing\threading_test.cc(193): FAILED:
  REQUIRE( counter1 <= ratio1 + 1 )
with expansion:
  7 <= 6
===============================================================================
test cases:  43 |  42 passed | 1 failed
assertions: 418 | 417 passed | 1 failed
```
Can you check it?

bwrsandman:
@q4a What hardware are you running? Can you paste the result of `cat /proc/cpuinfo` to a gist?

q4a:
@bwrsandman Here is my `cat /proc/cpuinfo`: https://paste.ubuntu.com/p/SXtG5y6S8s/
Sorry for late answer.

src/xenia/base/threading_posix.cc
```diff
@@ -47,265 +95,746 @@ void MaybeYield() {
...
+    auto predicate = [](auto h) { return h->signaled(); };
+
+    // Construct a condition for all or any depending on wait_all
+    auto operation = wait_all ? std::all_of<iter_t, decltype(predicate)>
+                              : std::any_of<iter_t, decltype(predicate)>;
+    auto aggregate = [&handles, operation, predicate] {
+      return operation(handles.cbegin(), handles.cend(), predicate);
+    };
+
+    std::unique_lock<std::mutex> lock(PosixConditionBase::mutex_);
```
Triang3l, line 220: This will probably deadlock if the thread is suspended between locking and waiting.

Triang3l (changes requested):

src/xenia/base/threading_posix.cc
Triang3l: Suspension needs rework to only occur immediately when the thread is in guest code, but deferred until the return to the guest if it happens during a kernel call.

This should be somewhat easy to implement on POSIX because the thread itself will reply to the suspension signal — it can set a flag in its context structure when it enters a kernel function, and the signal handler can check whether to sleep immediately or to defer the suspension.

On Windows, however (while there is a native WaitForMultipleObjects implementation that will work fine, having control over suspension would allow us more to have more granular kernel mutexes rather than just a single global_critical_region which is our current way of preventing suspension when kernel state is being modified, with easier usage of condition variables in the kernel — would possibly be useful for GPU readback in the access violation handler), the target thread can't provide its own response to suspension, and there are no signals. A possible solution is making suspension synchronous (which is contrary to the way it actually works on Windows, but possibly should have no noticeable effect on anything), with the source thread, not the target, being partially responsible for checking whether the target is in kernel code: call SuspendThread, then GetThreadContext to synchronize with the suspended thread, and check whether it's sleeping in kernel code currently — if yes, notify it that suspension was requested and unsuspend it on the host, to let it self-suspend on the next kernel to guest return.

JoelLinn (changes requested):
I checked the parts that concern Windows builds, I can't see any regressions. Didn't check on Linux stuff since we don't have anything now - I suggest we merge it and improve on it against master.

JoelLinn:
https://github.com/JoelLinn/xenia/commit/4646c901af04b4a9b8faa6371d441c4931e8ac78
You may consider cherry picking this commit to fix xe::threading::Fence.

PR #1324, opened 2019-01-30
Fixed link

BenNottelling:
Minor change, this link didn't point to the right file, so I corrected it..
Well, it didn't point to any file really


DrChat:
Much appreciated!

PR #1327, opened 2019-02-01
Add files via upload

niko1point0:
Made a change to the ISSUE_TEMPLATE


DrChat:
Appreciated - though users do not need to be reminded of this because they likely do not have this content in the first place.

PR #1322, opened 2019-01-27
Linux misc fixes

bwrsandman:
These are miscellaneous fixes leftover from my initiative to get Xenia running on Linux. The changes help getting Xenia up and running but they need to be combined with the other Linux PRs to see the full working benefit. See #1430.

* Clock commit borrowed from #1076

PR #1339, opened 2019-03-09
[Linux] xenia-cpu-ppc-tests (Rebase of #803)

### LINKED ISSUE #670: xenia-cpu-ppc-tests Linux Support

bwrsandman:
Description
---

This is not complete and will fail on travis. I might require help. See **Issues** for details.

Currently on master, the ppc emulation tests are not tested on travis. If run manually, they fail due to gcc/clang using different registers than the MSVC compiler. Guest to host and host to guest calls fail.

Fix status of this PR
---
This PR takes the original commits of #803 which have merge conflicts and rebases them on current master. There was a large refactoring of `x64_backend.cc` and some work had to be done to resolve them.

On top of that, I also enabled the tests on travis.

Currently, the tests run well on Debug and with the combination of this branch, #1317 and #1322 running roms actually execute code. There is more work to be done before an actual rom is playable but I was able to get guest to host logging calls from RDR.

Issues
---

While Debug works for the most part, running in release has some serious issues which I have trouble fixing.
The issue of Release is related to optimized builds. Even debug builds with an added `-O1` flag have this issue.

This issue is also present in the original PR (#803).

My best guess at this is that optimized builds might either compile away some thunk call code or that the compiler uses different registers than non-optimized builds.
I am including this branch for review and for some helpful tips from the community since I haven't had the time to investigate this fully and didn't want to leave my partial work floating in the ether if someone wanted to pick up this particular task.

Original Description by @uytvbn in #803 
---
> Hi,
> 
> This pull request fixes #670 (and also partially #356). I've tried to change as little as possible in Windows code.
> 
> ```
> $ ./build/bin/Linux/Debug/xenia-cpu-ppc-tests --log_file stdout
> ...
> i> 0000293A Total tests: 1457
> i> 0000293A Passed: 1457
> i> 0000293A Failed: 0
> ```
> 
> Comments:
> 
> * src/xenia/base/exception_handler_posix.cc - straightforward implementation based almost entirely on Windows version,
> 
> * src/xenia/base/math.h - off-by-one error, ffs contrary to _BitScanForward returns one-based index,
> 
> * src/xenia/base/memory_posix.cc - porting this exactly is probably not possible as there are significant differences in virtual memory handling between Windows and Linux, tried to follow the behavior of Windows implementation as closely as possible,
> 
> * src/xenia/base/string_buffer.cc - fixed non-standard behavior that happens to work with MSVC,
> 
> * src/xenia/cpu/backend/x64/x64_backend.cc - conform to Linux calling convention,
> 
> * src/xenia/cpu/backend/x64/x64_sequences.cc - the only change here that affects (code-wise) MSVC, it forces Linux to pass XMM values by pointer which is expected by codegen (default calling convention on Linux uses XMM registers), I've checked and it doesn't affect code generated by Visual Studio 2017, could probably hide it behind a typedef if you want to feel extra safe - though you might end up with different types for input and output,
> 
> * src/xenia/cpu/hir/value.cc - fixes Value::MulHi for 64 bit values, returned low not high qword,
> 
> * xenia-build - changes binutils path to use the one you get when building in third-party/binutils.
> 
> 
> Thanks.


premake5.lua
```diff
@@ -68,6 +68,11 @@ filter({"configurations:Debug", "platforms:Windows"})
     "/NODEFAULTLIB:MSVCRTD",
   })
 
+filter({"configurations:Debug", "platforms:Linux"})
+  buildoptions({
+    "-g",
```
DrChat: Since this is a part of another pull request - can you create a separate pull request with _just_ this change so we can merge it? Appreciated!

bwrsandman: Done in #1341

bwrsandman:
Build is green on travis and appveyor

bwrsandman:
Branch is ready for review as it is green. Here were my issues that were resolved.

> Issues
> ---
> 
> While Debug passes the tests and works for the most part, running in release has some serious issues which I have trouble fixing. This can be seen in travis tests.
> The issue of Release is related to optimized builds. Even debug builds with an added `-O1` flag have this issue.
> 
> This issue is also present in the original PR (#803).
> 
> My best guess at this is that optimized builds might either compile away some thunk call code or that the compiler uses different registers than non-optimized builds.
> 
> After doing some searching, I've tracked the obj file which is sensitive to `-O1`, it is `constant_propagation_pass.o`. If I compile xenia with `-O1` and delete that obj file and recompile with -O0 then the tests run successfully. If I do this with any other file, the issue persists. 
> Here's the objdump of `constant_propagation_pass.o` with both `-O0` and `O1`: https://gist.github.com/bwrsandman/d218a2d8590e5b74310cfd6dac8f234e
> 
> After more drilling down, it seems to come from some header included in `constant_propagation_pass.cc`. Clang has a pragma which disables optimization and by searching up the included headers, I found the highest single header for which disabling optimization fixes the problem. This header is `src/xenia/cpu/hir/value.h`. Something within this header is causing issues.
> ```diff
> diff --git a/src/xenia/cpu/hir/value.h b/src/xenia/cpu/hir/value.h
> index dcc95ca8..70a159fe 100644
> --- a/src/xenia/cpu/hir/value.h
> +++ b/src/xenia/cpu/hir/value.h
> @@ -109,6 +109,7 @@ class Value {
>    Use* AddUse(Arena* arena, Instr* instr);
>    void RemoveUse(Use* use);
>  
> +#pragma clang optimize off
>    void set_zero(TypeName new_type) {
>      type = new_type;
>      flags |= VALUE_IS_CONSTANT;
> @@ -175,6 +176,7 @@ class Value {
>      flags = other->flags;
>      constant.v128 = other->constant.v128;
>    }
> +#pragma clang optimize on
>  
>    inline bool IsConstant() const { return !!(flags & VALUE_IS_CONSTANT); }
>    bool IsConstantTrue() const {
> ```
> 
> The issue in within the `set_zero()` function of in `src/xenia/cpu/hir/value.h`. The following fixes the issue:
> ```diff
> diff --git a/src/xenia/cpu/hir/value.h b/src/xenia/cpu/hir/value.h
> index dcc95ca8..ae59a68e 100644
> --- a/src/xenia/cpu/hir/value.h
> +++ b/src/xenia/cpu/hir/value.h
> @@ -109,7 +109,7 @@ class Value {
>    Use* AddUse(Arena* arena, Instr* instr);
>    void RemoveUse(Use* use);
>  
> -  void set_zero(TypeName new_type) {
> +  void set_zero(TypeName new_type) __attribute__ ((optnone)){
>      type = new_type;
>      flags |= VALUE_IS_CONSTANT;
>      constant.v128.low = constant.v128.high = 0;
> ```
> The difference in opcode is here https://gist.github.com/bwrsandman/390712c2f29a7645f03e67d9daf8ce81
> 
> There are also similar issues happening on travis which I cannot replicate on my Arch setup with clang7. I am able to replicate these crashes on docker ubuntu 16.04 with clang6. Two solutions are available:
> * Do the same work as above to find the offending functions
> * Update travis to use clang7
> 
> The issues are similiar:
> * `Processor::backend()` inlining was causing issues
> * `Value::Not()` optimization was causing issues.
> 
> The issues are not clear and I am affraid that these are just hacks and not fixing the real underlying problems.
> 
> 
> I am including this branch for review and for some helpful tips from the community since I haven't had the time to investigate this fully and didn't want to leave my partial work floating in the ether if someone wanted to pick up this particular task.

DrChat:
So, in that specific function the difference between optimizing and not is the usage of AVX instructions.
What were the issues, exactly? Crashes? Clobbering?

src/xenia/cpu/backend/x64/x64_backend.cc
```diff
@@ -521,10 +564,18 @@ void X64ThunkEmitter::EmitLoadVolatileRegs() {
...
 
 void X64ThunkEmitter::EmitSaveNonvolatileRegs() {
-  // Preserve nonvolatile registers.
+#if XE_PLATFORM_LINUX
+  mov(qword[rsp + offsetof(StackLayout::Thunk, r[0])], rbx);
+  mov(qword[rsp + offsetof(StackLayout::Thunk, r[1])], rbp);
+  mov(qword[rsp + offsetof(StackLayout::Thunk, r[2])], r12);
+  mov(qword[rsp + offsetof(StackLayout::Thunk, r[3])], r13);
+  mov(qword[rsp + offsetof(StackLayout::Thunk, r[4])], r14);
+  mov(qword[rsp + offsetof(StackLayout::Thunk, r[5])], r15);
```
DrChat: N.B: SysV does not have nonvolatile XMM registers.

bwrsandman:
> So, in that specific function the difference between optimizing and not is the usage of AVX instructions.
> What were the issues, exactly? Crashes? Clobbering?

Clang 6 would vectorize the arguments inside a different register in Release than in Debug. This led to inconsistent behavior depending on the optimization level.

src/xenia/cpu/backend/x64/x64_backend.cc
bwrsandman: What is the suggested way of resolving this?

DrChat: Sorry - I think this was a note for myself. I might suggest you just drop a comment here and state what I said, just so nobody goes "wut, why aren't xmm registers being saved"

bwrsandman:
Rebased onto master and added the comment as per @DrChat request

bwrsandman:
@Margen67 cpu would be a better tag than test since the tests are testing the cpu :P

bwrsandman:
Cherry-pick register fix from https://github.com/xenia-project/xenia/pull/1517
If that gets merged beforehand, then this branch will need to be rebased.

Prism019:
#1517 has been merged.

sl1pkn07:
seems need rebase

greetrings

src/xenia/base/testing/string_test.cc
```diff
@@ -0,0 +1,34 @@
...
+namespace xe {
+namespace base {
+namespace test {
+
+TEST_CASE("StringBuffer") {
+  StringBuffer sb;
+  uint32_t module_flags = 0x1000000;
+
+  std::string path_(R"(\Device\Cdrom0\default.xex)");
+  sb.AppendFormat("Module %s:\n", path_.c_str());
```
JoelLinn, line 24: {fmt}

src/xenia/base/testing/string_test.cc
```diff
@@ -0,0 +1,34 @@
...
+  REQUIRE(sb.to_string() == "Module \\Device\\Cdrom0\\default.xex:\n");
+  sb.AppendFormat("    Module Flags: %.8X\n", module_flags);
```
JoelLinn, line 26: {fmt}

bwrsandman: Wouldn't it be `{:08X}` instead?

JoelLinn: I guess so, I just pasted that everywhere I saw a {fmt} related bug (thats the stylized lib name) while I was in there

bwrsandman:
3 tests started to fail on linux with relatively small floating point differences:
* `instr_fmadds.s`
* `instr_fsqrt.s`
* `instr_fmadd.s`

src/xenia/cpu/processor.h
```diff
@@ -29,6 +29,12 @@
 #include "xenia/cpu/thread_state.h"
 #include "xenia/memory.h"
 
+#if XE_PLATFORM_LINUX
```
JoelLinn, line 32: `not XE_COMPILER_MSVC`

src/xenia/cpu/processor.h
```diff
@@ -67,7 +73,9 @@ class Processor {
   Memory* memory() const { return memory_; }
   StackWalker* stack_walker() const { return stack_walker_.get(); }
   ppc::PPCFrontend* frontend() const { return frontend_.get(); }
-  backend::Backend* backend() const { return backend_.get(); }
+  // Clang 6 on release has issues with ppc instructions when this function is
+  // inlined
+  backend::Backend* backend() const NOINLINE { return backend_.get(); }
```
JoelLinn, line 78: weirdness again... there has to be a proper reason. @gibbed

src/xenia/cpu/testing/pack_test.cc
```diff
@@ -51,7 +51,7 @@ TEST_CASE("PACK_FLOAT16_2", "[instr]") {
       },
       [](PPCContext* ctx) {
         auto result = ctx->v[3];
-        REQUIRE(result == vec128i(0, 0, 0, 0x7FFFFFFF));
```
JoelLinn, line 54: Ok I see these tests fail on master. But are you sure the test case is at fault here? (This comment counts for all changes in this commit)

JoelLinn (changes requested):
Sorry for my comment bomb. I know you aren't even the original author of some of the changes... But there is definitely weird things in here, which we (>us two people) should discuss.

Changes concerning windows builds are minimal but some things like the cpu pack tests need discussion one way or another.

I hope we can get this patchset in a "comfortable" state and get it merged.

src/xenia/base/exception_handler_posix.cc
```diff
@@ -9,10 +9,135 @@
...
+      Exception::AccessViolationOperation access_violation_operation =
+          ((ucontext_t*)sigctx)->uc_mcontext.gregs[REG_ERR] & 0x2
+              ? Exception::AccessViolationOperation::kRead
+              : Exception::AccessViolationOperation::kWrite;
+      ex.InitializeAccessViolation(&thread_context,
+                                   reinterpret_cast<uint64_t>(siginfo->si_addr),
+                                   access_violation_operation);
+    } break;
+    default:
+      assert_always("Unhandled signum");
```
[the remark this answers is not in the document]
gibbed: Any `XELOG` you think would be useful, absolutely.

src/xenia/base/string_buffer.cc
```diff
@@ -68,11 +68,10 @@ void StringBuffer::Append(const std::string_view value) {
...
-  if (result <= 0) {
-    return;
-  }
-  auto length = static_cast<size_t>(result);
+  va_list size_args;
+  va_copy(size_args, args);  // arg is indeterminate after the return so copy it
+  int length = vsnprintf(nullptr, 0, format, size_args);
+  va_end(size_args);
   Grow(length + 1);
   vsnprintf(buffer_ + buffer_offset_, buffer_capacity_, format, args);
```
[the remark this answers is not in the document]
gibbed: Yes and yes.

src/xenia/cpu/hir/value.h
```diff
@@ -109,7 +115,9 @@ class Value {
   Use* AddUse(Arena* arena, Instr* instr);
   void RemoveUse(Use* use);
 
-  void set_zero(TypeName new_type) {
+  // Set optnone to prevent clang from vectorizing the assignment to 0 which
+  // would happen on different registers.
```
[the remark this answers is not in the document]
gibbed: Is it possible that the zeroing of rax by MSVC is somehow dodging a bug on Windows in following code, that clang is avoiding by vectorizing the code?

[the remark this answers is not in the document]
bwrsandman: It has been a while but `-O0` used rax but `-O1` used rdi. This was happening with the current clang version at the time.

src/xenia/cpu/testing/pack_test.cc
bwrsandman: I was using win32 as a working reference.
All the changes are done to match what is seen in win32.
I am not sure if these are the right outputs because these tests are pretty old.

src/xenia/cpu/processor.cc
```diff
@@ -197,7 +197,7 @@ Function* Processor::DefineBuiltin(const std::string_view name,
   function->set_end_address(address + 4);
   function->set_name(name);
 
-  auto builtin_function = static_cast<BuiltinFunction*>(function);
+  auto builtin_function = dynamic_cast<BuiltinFunction*>(function);
```
Triang3l, line 200: dynamic_cast (here and in module.cc) is only needed when you actually need to check if an object is an instance of a specific subclass — here there are no null checks afterwards, and as this function is declared in builtin_module_, it's clear that this will be a BuiltinFunction (and Symbol::Type::kFunction in module.cc also implies that it will be a Function), using static_cast is natural here.

Triang3l (changes requested):

src/xenia/cpu/processor.cc
bwrsandman: Rebased and removed the commit that adds dynamic cast

src/xenia/cpu/backend/x64/x64_seq_vector.cc
```diff
@@ -671,7 +671,7 @@ EMITTER_OPCODE_TABLE(OPCODE_VECTOR_SUB, VECTOR_SUB);
 // OPCODE_VECTOR_SHL
 // ============================================================================
 template <typename T, std::enable_if_t<std::is_integral<T>::value, int> = 0>
-static __m128i EmulateVectorShl(void*, __m128i src1, __m128i src2) {
+static __m128i EmulateVectorShl(void*, __m128i& src1, __m128i& src2) {
```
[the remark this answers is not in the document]
Triang3l: `const` also, by the way.

Triang3l (changes requested):

Triang3l:
The exception handler part has been merged into the main branch — thanks @uytvbn! 💚 I've also added RIP updating after an exception handler as that's needed for MMIO (it executes the `mov` manually and advances the instruction pointer). One additional change that I've made though is questionable — the original commit had bit `1 << 1` of the error code for a `SIGSEGV` interpreted as "read access violation" when it's set. However, `trap_pf.h` (not doing an `#include` of it though as it's not available on Android apparently) says that it's the opposite — `X86_PF_WRITE` is set to 1 for a write page fault. Which behavior is the correct in this situation?

PR #1341, opened 2019-03-10
Add debug symbols when building Debug on linux

bwrsandman:
The premake config doesn't include the debug symbols on linux in the debug config this adds them.

Commit cherry-picked from #1076.


DrChat:
Appreciated!

PR #1348, opened 2019-04-04
Instruction tests generated from console

DrChat:


halotroop2288 (approved):
There's no real reason to leave this open. You could take it or leave it.

Wunkolo:
Would greatly benefit test-coverage to get this in!

Wunkolo:
Just capturing this, but this is a very very huge and expansive set of tests, but might be too huge to add to the main repository due to the sheer size and amount of tests that it adds to the current set of unit tests. I've use these generated tests when adding and verifying new JIT optimizations though. It might be worth putting into its own repository.

PR #1351, opened 2019-04-08
[optimizations] More fast

chris-hawley:
Through changes in compile flags, games now run 3-4 fps faster. I also made a few changes so that we didn't have unnecessary returns on functions with static returns.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -79,7 +79,9 @@ filter("configurations:Release")
     "NDEBUG",
     "_NO_DEBUG_HEAP=1",
   })
-  optimize("On")
+  optimize("speed")
+  inlining("Auto")
+  floatingpoint("Fast")
   flags({
     "LinkTimeOptimization",
   })
@@ -88,6 +90,9 @@ filter({"configurations:Release", "platforms:Windows"})
   linkoptions({
     "/NODEFAULTLIB:MSVCRTD",
   })
+  buildoptions({
+    "/GT", -- enable fiber-safe optimizations
+   })
 
 filter("platforms:Linux")
   system("linux")


psennermann:
Is there a way to test this with Dx12?
Thanks

chris-hawley:
There is a build in the discord that has this and the other fixes currently pending in PR status. In the future, these types of questions should be made there. Plus, we got some great memes.

premake5.lua
```diff
@@ -79,7 +79,9 @@ filter("configurations:Release")
     "NDEBUG",
     "_NO_DEBUG_HEAP=1",
   })
-  optimize("On")
+  optimize("speed")
+  inlining ("Auto")
```
gibbed: Fix formatting.

premake5.lua
```diff
@@ -79,7 +79,9 @@ filter("configurations:Release")
...
+  floatingpoint ("Fast")
```
gibbed: Extraneous space.

gibbed: Extraneous space.

premake5.lua
gibbed: Extraneous space.

premake5.lua
```diff
@@ -157,7 +164,7 @@ filter("platforms:Windows")
     "/wd4127",  -- 'conditional expression is constant'.
     "/wd4324",  -- 'structure was padded due to alignment specifier'.
     "/wd4189",  -- 'local variable is initialized but not referenced'.
-    "/utf-8",   -- 'build correctly on systems with non-Latin codepages'.
+    "/utf-8",	-- 'build correctly on systems with non-Latin codepages'.
```
gibbed: Tab instead of spaces.

premake5.lua
Margen67: Does changing this to Fast break anything?

src/xenia/cpu/backend/x64/x64_function.cc
```diff
@@ -30,13 +30,12 @@ void X64Function::Setup(uint8_t* machine_code, size_t machine_code_length) {
...
 }
 
-bool X64Function::CallImpl(ThreadState* thread_state, uint32_t return_address) {
+void X64Function::CallImpl(ThreadState* thread_state, uint32_t return_address) {
   auto backend =
       reinterpret_cast<X64Backend*>(thread_state->processor()->backend());
   auto thunk = backend->host_to_guest_thunk();
   thunk(machine_code_, thread_state->context(),
         reinterpret_cast<void*>(uintptr_t(return_address)));
-  return true;
```
gibbed: While we don't currently do anything with the return value, that's left open for in case we want to in the future.

I really doubt this has an impact on performance, do you have any metrics on this specific change alone?

premake5.lua
chris-hawley: I need to just revert that since I didn't actually change anything

src/xenia/cpu/backend/x64/x64_function.cc
chris-hawley: no, it probably doesn't. I'll revert it

premake5.lua
chris-hawley: what extraneous space?

gibbed: `inlining ("Auto")` vs `inlining("Auto")`

premake5.lua
chris-hawley: Not that I've noticed. It does seem to increase the problem with audio. However, increasing the audio thread priority resolves that.

chris-hawley:
Changes made

premake5.lua
```diff
@@ -88,6 +90,11 @@ filter({"configurations:Release", "platforms:Windows"})
   linkoptions({
     "/NODEFAULTLIB:MSVCRTD",
   })
+  buildoptions({
+	"/Ot",
```
gibbed: Tab instead of spaces.

premake5.lua
```diff
@@ -88,6 +90,11 @@ filter({"configurations:Release", "platforms:Windows"})
...
+	"/GL",
```
gibbed: Tab instead of spaces.

premake5.lua
```diff
@@ -88,6 +90,11 @@ filter({"configurations:Release", "platforms:Windows"})
...
+	"/GT",
```
gibbed: Tab instead of spaces.

premake5.lua
```diff
@@ -88,6 +90,11 @@ filter({"configurations:Release", "platforms:Windows"})
   linkoptions({
     "/NODEFAULTLIB:MSVCRTD",
   })
+  buildoptions({
```
gibbed: It'd be nice if you could explain what these build options actually do in a comment so we don't have to reference MSDN docs for it.

premake5.lua
```diff
@@ -79,7 +79,9 @@ filter("configurations:Release")
     "NDEBUG",
     "_NO_DEBUG_HEAP=1",
   })
-  optimize("On")
+  optimize("speed")
+  inlining("Auto")
+  floatingpoint("Fast")
```
gibbed: I think you mentioned side effects to this change, or was that for AVX2? (which isn't in this PR?)

chris-hawley: It was only for AVX2 and it was limited to only a few sections of games. Blue Dragon being the most noticeable and it only happens sometimes. So not sure if it's actually a problem somewhere else that just shows up with AVX2.

premake5.lua
chris-hawley: Ahh, yes. They are because premake doesn't have an option for those. I'll add in a comment.

premake5.lua
chris-hawley: Are you saying that you want tabs instead of spaces?

gibbed: No, I'm saying there is a tab instead of spaces. The rest of the file is space delimited.

chris-hawley:
changes made

chris-hawley:
should be gucci now

PR #1353, opened 2019-04-13
[Kernel] Add ObReferenceObjectByName

emoose:
Adds ObReferenceObjectByName to xboxkrnl_ob.cc, this is used by the XEFU Xbox OG emu and required for it to progress.

This seems to be a working impl for it, though two of the args for it go unused (attributes & parse_context), we don't seem to handle those in any other Ob* funcs though so I guess it's fine for now.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
@@ -144,6 +144,21 @@ dword_result_t ObReferenceObjectByHandle(dword_t handle,
 }
 DECLARE_XBOXKRNL_EXPORT1(ObReferenceObjectByHandle, kNone, kImplemented);
 
+dword_result_t ObReferenceObjectByName(lpstring_t name, dword_t attributes,
+                                       dword_t object_type_ptr,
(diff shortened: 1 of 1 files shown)


gibbed:
`parse_context`?

gibbed:
> `parse_context`?

To clarify, I'm asking what this is since you declare it but never touch it again.

emoose:
Sorry for the late reply, been away from home a little while.

I got the name for that arg [from Cxbx](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/0077eb7ab296a1636bac54f32fb44f6d2c35e3ae/src/core/kernel/exports/EmuKrnlOb.cpp#L872) (they also name what we have as "unk" in ObOpenObjectByName the same thing too), in Cxbx the arg gets passed along to the objects "ParseProcedure" it seems, but afaik we don't have anything setup for that atm, so I let the arg go unused for now.

(Also the only actual use of it by an XEX that I've seen had it as 0, so I don't think it matters too much, not sure about ObOpenObjectByName though)

It could be worth looking into Cxbx sometime, their Object code looks a bit more fleshed out than ours at least, maybe they've documented more currently-unknown params too.

PR #1354, opened 2019-04-13
[Kernel] Fix loading import library modules

emoose:
This should fix loading import libraries, so that any executable XEXs that require imported funcs from them should now work fine, this is done by attempting to load the library module if KernelState::GetModule can't find the module already.

The second commit should allow running any entrypoint that import libraries may have (ie DllMain), there was code in place for this already, but it seems changes to how Xenia handles threads must have broken that, so for this I had to change XThread's ctor params to allow specifying a "StartupType" (what kind of arguments to pass to the thread's start address), with that we can now use LaunchModule to load the DLL's entrypoint just like it was a normal XEX module.

XEFU requires these changes to work, and maybe Halo 3 Beta too (it gets further now at least, but still doesn't seem to run... cache:\ drive needs to be mapped for it, and also needs to be populated with folders before loading it too since Xenia doesn't seem to create folders properly for it atm, but now it gets stuck after loading Q10.dll..)

It might be a good idea to test out any games that have import libraries included, if they worked prior to these fixes then these changes could break things now that the libraries are actually being loaded in.


emoose:
Fixed the issue with Halo 3, the check in LaunchModule for running the DllMain properly was broken, making any loaded DLLs not get ran correctly...

Fixed now, and Halo 3 Beta now seems to get a little bit further too. Need to test with some more games.

Razzile:
I can confirm this allows Fuzion Frenzy 2 to boot now (it required a ELEA.dll to be loaded) although the game now freezes shortly after boot. not sure if that's related to dll loading yet

jacob1218:
While this did fix COD: Classic, it broke bf3, and split/second. Logs attached.
[xenia.log](https://github.com/xenia-project/xenia/files/3087565/xenia.log)
[xenia.log](https://github.com/xenia-project/xenia/files/3087567/xenia.log)

src/xenia/kernel/kernel_state.cc
```diff
@@ -229,37 +229,53 @@ object_ref<XModule> KernelState::GetModule(const char* name, bool user_only) {
...
+  xe::kernel::object_ref<XThread> thread = nullptr;
+
+  if (!module->is_dll_module()) {
+    // Not a DLL module, run entrypoint as normal
+    thread = object_ref<XThread>(new XThread(
+        kernel_state(), module->stack_size(), XThread::StartupType::Normal, 0,
+        module->entry_point(), 0, X_CREATE_SUSPENDED, true, true));
+  } else {
+    // Run entrypoint as DllMain, using module handle as start context
+    thread = object_ref<XThread>(
```
DrChat: So, usually DllMain is called from the same thread that called the loader, i.e. `<game code> -> LoadLibrary -> <DLL code>`.
I believe it's this way on the xbox as well - so changing this behavior could introduce race conditions in guest code.

src/xenia/kernel/kernel_state.cc
```diff
@@ -229,37 +229,53 @@ object_ref<XModule> KernelState::GetModule(const char* name, bool user_only) {
       return retain_object(user_module.get());
     }
   }
-  return nullptr;
+
+  // Module not found, try loading it
+  return LoadUserModule(name);
```
DrChat: We shouldn't implicitly load a module here - this function is called from places like XexGetModuleHandle that explicitly do not load new modules.

What we can do instead is in SetuplibraryImports, if `kernel_state_->IsKernelModule(name)` is false, call `GetModule`, and if no module is returned then implicitly load one.

src/xenia/cpu/xex_module.cc
```diff
@@ -486,8 +486,6 @@ int XexModule::ReadImage(const void* xex_addr, size_t xex_length,
     return 0;
   }
 
-  memory()->LookupHeap(base_address_)->Reset();
```
DrChat: This isn't strictly necessary - but there also isn't a reason to remove this.

src/xenia/kernel/kernel_state.cc
```diff
@@ -277,9 +293,6 @@ object_ref<UserModule> KernelState::GetExecutableModule() {
 }
 
 void KernelState::SetExecutableModule(object_ref<UserModule> module) {
-  if (module.get() == executable_module_.get()) {
```
DrChat: We can probably leave this here - I see no reason to remove this check.

PR #1356, opened 2019-04-14
[Kernel] Add XeCryptAes* and XeCryptRc4* functions

emoose:
This adds all the XeCrypt functions needed for AES & RC4, the XeCryptAes* adds seem to fix strings in Halo Reach, not sure if many other games will be affected by it though.

Reach

The AES functions make use of https://github.com/kokke/tiny-AES-c, since the context struct needed for that is smaller than the actual XECRYPT_AES_STATE struct we can just store tiny-AES-c's context in there instead, seems to work fine that way.

RC4 code is from code I wrote for supporting 360 profiles, worked fine for decrypting the 360's Account file, but I'm not sure if there's any games which use it that we could test against...

In my profiles branch I split the RC4 code into a separate file so that our profile code could make use of it, but since that code isn't in master I've just kept it inside xboxkrnl_crypt for now, maybe it can be split off in future once we have other code that needs it.

### DIFF
--- a/src/xenia/kernel/util/aes.cc
+++ b/src/xenia/kernel/util/aes.cc
@@ -1,0 +1,555 @@
+// tiny-AES-c from https://github.com/kokke/tiny-AES-c
+/*
+
+This is an implementation of the AES algorithm, specifically ECB, CTR and CBC
+mode. Block size can be chosen in aes.h - available choices are AES128, AES192,
+AES256.
+
+The implementation is verified against the test vectors in:
+  National Institute of Standards and Technology Special Publication 800-38A
+2001 ED
+
+ECB-AES128
+----------
+
+  plain-text:
+    6bc1bee22e409f96e93d7e117393172a
+    ae2d8a571e03ac9c9eb76fac45af8e51
+    30c81c46a35ce411e5fbc1191a0a52ef
+    f69f2445df4f9b17ad2b417be66c3710
+
+  key:
+    2b7e151628aed2a6abf7158809cf4f3c
+
+  resulting cipher
+    3ad77bb40d7a3660a89ecaf32466ef97
+    f5d3d58503b9699de785895a96fdbaaf
+    43b1cd7f598ece23881b00e3ed030688
+    7b0c785e27e8ad3f8223207104725dd4
+
+
+NOTE:   String length must be evenly divisible by 16byte (str_len % 16 == 0)
+        You should pad the end of the string with zeros if this is not the case.
+        For AES192/256 the key size is proportionally larger.
+
+*/
+
+/*****************************************************************************/
+/* Includes:                                                                 */
+/*****************************************************************************/
+#include "aes.h"
+#include <stdint.h>
+#include <string.h>  // CBC mode, for memset
+
+/*****************************************************************************/
+/* Defines:                                                                  */
+/*****************************************************************************/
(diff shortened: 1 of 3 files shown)


Triang3l:
Merged RC4, but going to try a different library for AES that would let us keep the original context structure and modify the initialization vector for chained XeCryptAesCbc calls.

emoose:
Ah nice, I did try making my own to fit the context along with adding AES-NI support, but couldn't get it to decrypt properly again :/

If you can get the context filled properly adding AES-NI support should be pretty straight forward from there though I think, https://gist.github.com/acapola/d5b940da024080dfaf5f has some loop-unrolled AES-NI code which could come in useful.

AFAIK AES-NI only needs the exact data that's stored in the context, so should be perfect for us to use (also think there's some instructions to help generate/"keygen" that context data too), we'd have to check for support for AES-NI though of course (or do we? haven't CPUs had that for almost 10 years now?)

Triang3l:
Wow, that AES-NI sample is really short and clean. Would be even more useful for XEX loading I guess. But as a fallback I'm going to try https://github.com/openluopworld/aes_128, not sure if it generates `keytabenc` with the exact same layout though, wish I could compare.

emoose:
xorloser's [windows xecrypt.lib](https://github.com/jogolden/XeCrypt) does everything the same as 360 afaik, you could maybe try comparing against that if you need to.

That code looks pretty good too, should be a way to get that working with the contexts hopefully... Also glad it has both input & output params, less stuff needing to be memcpy'd around :P

Once you're done I'll try looking into adding AES-NI, + making all the other loose AES stuff use that instead if you haven't already.

Where are you thinking of storing the AES code btw? A crypto folder that all the projects could access would be nice, since there's also RC4 that needs to be split off too, and probably stuff like RSA eventually, not really sure how to go about doing something like that though.

Triang3l:
Is the source for that xecrypt.lib available?

emoose:
Sadly not :(, you should be able to create an exe with it to look at the context struct it makes though, but I'm not sure if it might need a certain VS version or anything.

Triang3l:
Our crypto dependencies are stored in third_party/crypto (though I'll add a separate third_party submodule for aes_128 so submodule structure is more unified), but for our own code (like AES-NI) — probably in a file in src/xenia/base/, where general-purpose utility functions are stored. `keytabdec` generation is only needed for XeCryptAesKey (since in other places we can use the inverse function directly, without the need for `keytabdec`), so I'll hardcode it in XeCryptAesKey at least for now.

Comparing context data sounds like a good idea, by the way.

Triang3l:
Implemented AES in 85e26c5dde4f5bb9d1e46fc4caa42ea3354d8fc5.
Unfortunately xorloser's XeCrypt.lib has a different context structure on PC (since it uses OpenSSL behind the scenes — allows 14 rounds for AES-256 and stores round count). I still don't know what the layout of `keytabdec` should be — inverted (as opposed to `keytabenc`) or not, what it should actually store even. I'm filling it just to make it somewhat valid at least, but still using `keytabenc` for decryption too (that's not right probably, but I wouldn't really expect games to upload the decryption round keys manually rather than using XeCryptAesKey).

ObiKKa:
[This video](https://youtu.be/KSzflZiuKe4) from Leafnation, showing Halo Reach running with fixed menu text, also claims that menu text in Halo 4 is also fixed.

PR #1358, opened 2019-04-17
Replace GFLAGS with a config system + config file

jgoyvaerts:
This pullrequest adds a config system to Xenia which functions in a similar way like GFLAGS but with a lot more features.
- The config file is located in the Xenia user folder (or in the same folder as the executable when running in portable mode), called xenia.config.toml
- If a config file does not exist, it will be created on launch, filled with default values. This allows users to delete their config if they messed it up and ensures everyone has a config file.
- If new config variables are added in a later version, the existing config file will be updated with the new config variables and their default values without overwriting any other user modified variables.
- All config flags can still be specified as launch parameter, which will take priority over the config variable

1. The possibility exists to create game specific config files, in the form of xenia.config.toml. This file should be placed in the content folder, in the subfolder with the correct game title id. The config file will be loaded upon game launch. For example, this allows a user to create a one line config file which switches a specific game to vulkan instead of D3D12. (However, this specific case as described will not work since several components (such as the graphics system) are loaded before game launch. Some rewriting of code is needed in the future to take full advantage of these game specific config files.)

2. The possibility also exists to specify a --config launch parameter to load a specific config file which is not in the default location. This, when combined with the --target launch parameter allows shortcuts to individual games with individual config files, for use in launchers such as [Bottlenose ](https://github.com/quinton-ashley/bottlenose/) and others.

3. It should be possible to create a console which launches together with xenia which allows you to modify config variables at runtime (if they support it). This is not in scope of this pullrequest.

For previous comments regarding this pull request when it was in construction, see https://github.com/xenia-project/xenia/pull/1286


gibbed:
1. Needs better handling for config file location. Next to the executable should only be in portable mode.
1. Game configs probably shouldn't be next to app config but probably in a subdirectory?
1. Not sure if I like `.cfg` as the file extension. Would prefer something like `.config.toml`.

jgoyvaerts:
Done done and done

jgoyvaerts:
Updated PR:
- Config is now dynamic typed (so boolean is true instead of "true")
- Renamed --cfg launch parameter to --config
- Added some logging

src/xenia/base/cvar.h
```diff
@@ -0,0 +1,258 @@
...
+
+#define DEFINE_string(name, defaultValue, description, category) \
+  DEFINE_CVar(name, defaultValue, description, category, std::string)
+
+#define DEFINE_bool(name, defaultValue, description, category) \
+  DEFINE_CVar(name, defaultValue, description, category, bool)
+
+#define DEFINE_CVar(name, defaultValue, description, category, type) \
+  namespace cvar {                                                   \
+  type var_##name = defaultValue;                                    \
```
gibbed: Pick one:
1. `type VAR_##name = defaultValue;`, `using cvar::VAR_##name`. `VAR_name` for access.
2. `type name = defaultValue;`, `cvar::name` for access.

Upper case prefix makes things stand out clearly if we're not going to use the namespace. Personally, I'd very much prefer the latter, but either is acceptable.

If you go with the latter, probably better to stuff them into `cvars` namespace instead of `cvar` to make things distinct.

src/xenia/base/logging.cc
```diff
@@ -328,7 +331,7 @@ void FatalError(const char* fmt, ...) {
                 MB_OK | MB_ICONERROR | MB_APPLMODAL | MB_SETFOREGROUND);
   }
 #endif  // WIN32
-
+  logger_->~Logger();
```
gibbed: I don't think this belongs to this PR. What is this changing?

src/xenia/base/string_util.h
```diff
@@ -24,13 +24,13 @@ namespace string_util {
 
 inline std::string to_hex_string(uint32_t value) {
   char buffer[21];
-  std::snprintf(buffer, sizeof(buffer), "%08" PRIX32, value);
+  snprintf(buffer, sizeof(buffer), "%08" PRIX32, value);
```
gibbed: I don't think this belongs to this PR. Why did you drop `std::`?

src/xenia/base/string_util.h
```diff
@@ -24,13 +24,13 @@ namespace string_util {
...
   return std::string(buffer);
 }
 
 inline std::string to_hex_string(uint64_t value) {
   char buffer[21];
-  std::snprintf(buffer, sizeof(buffer), "%016" PRIX64, value);
+  snprintf(buffer, sizeof(buffer), "%016" PRIX64, value);
```
gibbed: I don't think this belongs to this PR. Why did you drop `std::`?

src/xenia/base/string_util.h
```diff
@@ -54,8 +54,8 @@ inline std::string to_hex_string(double value) {
 
 inline std::string to_hex_string(const vec128_t& value) {
   char buffer[128];
-  std::snprintf(buffer, sizeof(buffer), "[%.8X, %.8X, %.8X, %.8X]",
-                value.u32[0], value.u32[1], value.u32[2], value.u32[3]);
+  snprintf(buffer, sizeof(buffer), "[%.8X, %.8X, %.8X, %.8X]", value.u32[0],
```
gibbed: I don't think this belongs to this PR. Why did you drop `std::`?

src/xenia/base/string_util.h
```diff
@@ -66,7 +66,7 @@ inline std::string to_hex_string(const __m128& value) {
   char buffer[128];
   float f[4];
   _mm_storeu_ps(f, value);
-  std::snprintf(
+  snprintf(
```
gibbed: I don't think this belongs to this PR. Why did you drop `std::`?

src/xenia/base/string_util.h
```diff
@@ -77,15 +77,22 @@ inline std::string to_string(const __m128& value) {
   char buffer[128];
   float f[4];
   _mm_storeu_ps(f, value);
-  std::snprintf(buffer, sizeof(buffer), "(%F, %F, %F, %F)", f[0], f[1], f[2],
-                f[3]);
+  snprintf(buffer, sizeof(buffer), "(%F, %F, %F, %F)", f[0], f[1], f[2], f[3]);
```
gibbed: I don't think this belongs to this PR. Why did you drop `std::`?

src/xenia/base/vec128.h
```diff
@@ -266,8 +266,8 @@ static inline vec128_t vec128b(uint8_t x0, uint8_t x1, uint8_t x2, uint8_t x3,
 
 inline std::string to_string(const vec128_t& value) {
   char buffer[128];
-  std::snprintf(buffer, sizeof(buffer), "(%g, %g, %g, %g)", value.x, value.y,
-                value.z, value.w);
+  snprintf(buffer, sizeof(buffer), "(%g, %g, %g, %g)", value.x, value.y,
```
gibbed: I don't think this belongs to this PR. Why did you drop `std::`?

src/xenia/kernel/util/shim_utils.h
```diff
@@ -408,8 +405,8 @@ inline void AppendParam(StringBuffer* string_buffer,
 }
 inline void AppendParam(StringBuffer* string_buffer,
                         pointer_t<X_EXCEPTION_RECORD> record) {
-  string_buffer->AppendFormat("%.8X(%.8X)", record.guest_address(),
-                              uint32_t(record->exception_code));
+  /*string_buffer->AppendFormat("%.8X(%.8X)", record.guest_address(),
```
gibbed: I don't think this belongs to this PR.

tools/build/scripts/test_suite.lua
```diff
@@ -25,8 +25,7 @@ local function combined_test_suite(test_suite_name, project_root, base_path, con
       project_root.."/"..build_bin,
     }))
     links(merge_arrays(config["links"], {
-      "gflags",
-    }))
+	}))
```
gibbed: Tab instead of spaces.

tools/build/scripts/test_suite.lua
```diff
@@ -50,8 +49,7 @@ local function split_test_suite(test_suite_name, project_root, base_path, config
         project_root.."/"..build_bin,
       }))
       links(merge_arrays(config["links"], {
-        "gflags",
-      }))
+	  }))
```
gibbed: Tab instead of spaces.

src/xenia/app/premake5.lua
```diff
@@ -87,6 +83,7 @@ project("xenia-app")
     if not os.isfile(user_file) then
       debugdir(project_root)
       debugargs({
-        "--flagfile=scratch/flags.txt",
```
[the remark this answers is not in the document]
jgoyvaerts: A flagsfile is a concept of GFLAGS but this PR gets rid of GFLAGS all together, so this option is obsolete. If you specify a --flagfile option at launch you would get an error that it's an invalid launch argument.

We could replace this by using the new --config=scratch/xenia.config.toml launch option but you can also just change the config itself so I opted to remove the debug arg all together

src/xenia/app/premake5.lua
```diff
@@ -87,6 +83,7 @@ project("xenia-app")
...
+        "2>&1",
```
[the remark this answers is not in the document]
jgoyvaerts: Removed, wasn't meant to be there in the first place

src/xenia/app/premake5.lua
```diff
@@ -87,6 +83,7 @@ project("xenia-app")
...
+        "1>scratch/stdout.txt",
```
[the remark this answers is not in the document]
jgoyvaerts: Removed, wasn't meant to be there in the first place

src/xenia/base/cvar.h
jgoyvaerts: Good suggestion, went with the latter approach. All cvars are now in cvars::name

src/xenia/base/string_util.h
jgoyvaerts: Clarified the drop of std:: in the commit message, copying here: 
This seems to be an issue with Microsoft declaring a snprintf macro in an included header which causes issues.
Related issues:
- https://github.com/nlohmann/json/issues/1408
- https://github.com/kodi-pvr/pvr.vuplus/issues/162

src/xenia/kernel/util/shim_utils.h
jgoyvaerts: Reversed

tools/build/scripts/test_suite.lua
jgoyvaerts: Fixed

tools/build/scripts/test_suite.lua
jgoyvaerts: Fixed

src/xenia/base/logging.cc
jgoyvaerts: See commit message, by manually destroying the logger when a fatalerror is triggered, we trigger a log flush so the fatal error actually shows up in the log file.

src/xenia/base/string_util.h
jgoyvaerts: Clarified the drop of std:: in the commit message, copying here:
This seems to be an issue with Microsoft declaring a snprintf macro in an included header which causes issues.
Related issues:

nlohmann/json#1408
kodi-pvr/pvr.vuplus#162

src/xenia/base/string_util.h
jgoyvaerts: Clarified the drop of std:: in the commit message, copying here:
This seems to be an issue with Microsoft declaring a snprintf macro in an included header which causes issues.
Related issues:

nlohmann/json#1408
kodi-pvr/pvr.vuplus#162

src/xenia/base/string_util.h
jgoyvaerts: Clarified the drop of std:: in the commit message, copying here:
This seems to be an issue with Microsoft declaring a snprintf macro in an included header which causes issues.
Related issues:

nlohmann/json#1408
kodi-pvr/pvr.vuplus#162

src/xenia/base/vec128.h
jgoyvaerts: Clarified the drop of std:: in the commit message, copying here:
This seems to be an issue with Microsoft declaring a snprintf macro in an included header which causes issues.
Related issues:

nlohmann/json#1408
kodi-pvr/pvr.vuplus#162

src/xenia/base/string_util.h
jgoyvaerts: Clarified the drop of std:: in the commit message, copying here:
This seems to be an issue with Microsoft declaring a snprintf macro in an included header which causes issues.
Related issues:

nlohmann/json#1408
kodi-pvr/pvr.vuplus#162

src/xenia/base/logging.cc
gibbed: It would probably do to move this change out into its own PR all on its own. Instead of being buried in unrelated code.

jgoyvaerts: Moved to a separate PR. This PR is not dependent on that fix anyway, was just a quality of life improvement

jgoyvaerts:
Updated to resolve all requested changes, travis build works this time

Razzile:
This might not be the correct place to list this but @jgoyvaerts asked me to put these issues in the PR:

in the example config file `win32_high_freq = true                      # Audio system. Use: [any, nop, xaudio2]` has the wrong comment next to it.

Also there's a semantics issue in that `cpu = "any"` is under the [CPU] section but `gpu = "any"` and `apu = "any"` are under [General]

one other thing, would it be cleaner to have Vulkan and DX12 specific options under [GPU.Vulkan] and [GPU.D3D12] instead of their own standalone section (e.g. just [Vulkan] in its own)? This is supported by the toml format

Prism019:
I like the proposition of each section having its own config section. Perhaps they could be organized based on the folders in the source tree? For example, XAudio2 settings could be under `[APU.XAudio2]`, or the x64 backend settings could be under `[CPU.Backend.x64]`.

Because of that, `gpu = "any"` and `apu = "any"` would be under [App], so either they should be moved to their respective `[GPU]` and `[APU]` settings, or maybe `cpu = "any"` should be moved to `[App]`?

Margen67:
Will a config option for Discord need to be added first, or can that be added in a later PR?

PR #1361, opened 2019-04-25
Add a check to see if divisor is -1 in integer division

### LINKED ISSUE #1085: DIV_I32 and others don't handle INT_MIN / -1
https://github.com/benvanik/xenia/blob/190108dab6732d9db5c8595e1358b7b034cb5a14/src/xenia/cpu/backend/x64/x64_sequences.cc#L4332

It was pointed out to me that x86 will raise an exception when dividing INT_MIN by -1.
I haven't encountered any games doing this yet - but it could potentially be an issue.

Could be worth investigating if you're a new contributor.

IceTDrinker:
Resolve #1085

1) In that case use neg on the register containing the dividend.

This avoids an x86 CPU exception when dividend is INT_MIN or similar and
divisor is -1

The result will be wrong in that last case and still equal to INT_MIN as
the true result == INT_MAX + 1 would require an additional bit to store

For PPC, the behavior is apparently undefined in that case, so hopefully
x360 game devs were not relying on such values

2) Code now returns 0 to comply with results from RPCS3 when encountering
INT_MIN / -1 case

3) Corrected code after testing with cpu-tests.

Oddity with Xbyak handling of imm: in source code, imm have to be 32 bits
when using 16 bits registers (potential bug in opRM_I ) for the cmp op.

64 bits imm are not usable with cmp (see intel ref manual), used intermediate register
for DIV_I64

Stopped using CodeGenerator:T_SHORT, as default T_AUTO already tries to use
T_SHORT

4) Code clean-up based on better understanding of Xbyak

### DIFF
--- a/src/xenia/cpu/backend/x64/x64_sequences.cc
+++ b/src/xenia/cpu/backend/x64/x64_sequences.cc
@@ -1678,6 +1678,9 @@ EMITTER_OPCODE_TABLE(OPCODE_MUL_HI, MUL_HI_I8, MUL_HI_I16, MUL_HI_I32,
 struct DIV_I8 : Sequence<DIV_I8, I<OPCODE_DIV, I8Op, I8Op, I8Op>> {
   static void Emit(X64Emitter& e, const EmitArgType& i) {
     Xbyak::Label skip;
+    Xbyak::Label normalDiv;
+    constexpr int8_t minusOne{-1};
+    constexpr int8_t minInt8{INT8_MIN};
     e.inLocalLabel();
 
     if (i.src2.is_constant) {
@@ -1688,12 +1691,19 @@ struct DIV_I8 : Sequence<DIV_I8, I<OPCODE_DIV, I8Op, I8Op, I8Op>> {
         e.div(e.cl);
       } else {
         e.movsx(e.ax, i.src1);
+        e.cmp(e.al, minInt8);
+        e.jne(normalDiv);
+        e.cmp(e.cl, minusOne);
+        e.jne(normalDiv);
+        e.xor_(e.ax, e.ax);
+        e.jmp(skip);
+        e.L(normalDiv);
         e.idiv(e.cl);
       }
     } else {
       // Skip if src2 is zero.
       e.test(i.src2, i.src2);
-      e.jz(skip, CodeGenerator::T_SHORT);
+      e.jz(skip);
 
       if (i.instr->flags & ARITHMETIC_UNSIGNED) {
         if (i.src1.is_constant) {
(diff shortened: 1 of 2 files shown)


IceTDrinker:
There probably is a dire need for unit testing, I modified the file for addition testing locally but I did not manage to add a file for division testing any tips so that I (maybe) update the PR ?

IceTDrinker:
I talked to the developer of Xbyak, the code I wrote can be improved, I'll push a corrected commit when I get the time at home

src/xenia/cpu/backend/x64/x64_sequences.cc
```diff
@@ -1816,6 +1895,10 @@ struct DIV_I32 : Sequence<DIV_I32, I<OPCODE_DIV, I32Op, I32Op, I32Op>> {
 struct DIV_I64 : Sequence<DIV_I64, I<OPCODE_DIV, I64Op, I64Op, I64Op>> {
   static void Emit(X64Emitter& e, const EmitArgType& i) {
     Xbyak::Label skip;
+    Xbyak::Label neg;
+    Xbyak::Label normalDiv;
+    constexpr const int32_t minusOne{-1};
```
elad335: const applies implicitly for constexpr.

src/xenia/cpu/backend/x64/x64_sequences.cc
```diff
@@ -1708,6 +1724,17 @@ struct DIV_I8 : Sequence<DIV_I8, I<OPCODE_DIV, I8Op, I8Op, I8Op>> {
         } else {
           e.movsx(e.ax, i.src1);
         }
+        e.cmp(i.src2, minusOne);
+        e.jne(normalDiv);
+        e.cmp(e.al, minInt8);
```
elad335: dividend = int_min is extremely unlikely while divisor = -1 is much more likely, so you can reverse the order of the checks (first check dividend) so it will skip the second check when dividend != int_min, leading to little reduced overall amount of checks.

src/xenia/cpu/backend/x64/x64_sequences.cc
IceTDrinker: Will get on it, it was an habit in a company I worked for to still specify const after constexpr

src/xenia/cpu/backend/x64/x64_sequences.cc
IceTDrinker: Good point

IceTDrinker:
Any thoughts on the division tests added ? Some other specific values to test ?

PR #1362, opened 2019-04-27
FPS Counter at the end of the display chain

IceTDrinker:
FPS display based on profiler data, toggleable by pressing the the h key
on the keyboard, use Shift + h to modify FPS text scale (x1, x2 or x4)

### DIFF
--- a/src/xenia/ui/window.cc
+++ b/src/xenia/ui/window.cc
@@ -17,6 +17,8 @@
 #include "xenia/base/logging.h"
 #include "xenia/ui/imgui_drawer.h"
 
+#include "xenia/base/profiling.h"
+
 namespace xe {
 namespace ui {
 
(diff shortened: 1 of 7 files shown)


illusion98:
What happened to the repository?
@IceTDrinker

IceTDrinker:
I deleted it, I realized I would not have time to work on xenia, code of the PR is still accessible though from what I see on github

src/xenia/ui/window.cc
```diff
@@ -17,6 +17,8 @@
 #include "xenia/base/logging.h"
 #include "xenia/ui/imgui_drawer.h"
 
+#include "xenia/base/profiling.h"
+
```
illusion98: Maybe an option for the user to toggle showing the FPS number in the titlebar.
Since applications like OBS won't be able to hook due to the title kept changing. 1/2
Note: i can't start a review at line 15 so when merging include cvar before logging.
```suggestion
#include "xenia/base/cvar.h"

DEFINE_bool(fps_titlebar, true, "Show FPS in titlebar", "General");

File: src/xenia/ui/window.cc [hunk 2]
```diff
@@ -180,6 +182,33 @@ void Window::OnPaint(UIEvent* e) {
...
+    }
+    float fMs = fToMs * nFlipTicks;
+
+    if (fMs != 0.0f) {
+      game_fps_ = static_cast<uint32_t>(1000.0f / fMs);
+    }
+#endif
+
+    title_fps_text_ = base_title_;
+    title_fps_text_ += L" | ";
```
illusion98: Maybe an option for the user to toggle showing the FPS number in the titlebar.
Since applications like OBS won't be able to hook due to the title kept changing. 2/2
```suggestion
    if (cvars::fps_titlebar) {
      title_fps_text_ += L" | ";
      title_fps_text_ += std::to_wstring(game_fps_);
      title_fps_text_ += L" FPS";
    }

PR #1363, opened 2019-05-02
Log build info and module info with error level

IceTDrinker:
Allows to have useful log 0 with build info and module info

### DIFF
--- a/src/xenia/base/main_win.cc
+++ b/src/xenia/base/main_win.cc
@@ -139,8 +139,8 @@ int Main() {
   }
 
   // Print version info.
-  XELOGI("Build: %s / %s on %s", XE_BUILD_BRANCH, XE_BUILD_COMMIT,
-         XE_BUILD_DATE);
+  xe::LogLineFormat(LogLevel::Error, 'i', "Build: %s / %s on %s",
+                    XE_BUILD_BRANCH, XE_BUILD_COMMIT, XE_BUILD_DATE);
 
   // Request high performance timing.
   if (FLAGS_win32_high_freq) {
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -620,6 +620,7 @@ X_STATUS Emulator::CompleteLaunch(const std::wstring& path,
   module->GetOptHeader(XEX_HEADER_EXECUTION_INFO, &info);
   if (info) {
     title_id_ = info->title_id;
+    xe::LogLineFormat(xe::LogLevel::Error, 'i', "Title ID : %.8X\n", title_id_);
   }
 
   // Try and load the resource database (xex only).


src/xenia/kernel/user_module.cc
```diff
@@ -804,7 +804,7 @@ void UserModule::Dump() {
     sb.AppendFormat("\n");
   }
 
-  xe::LogLine(xe::LogLevel::Info, 'i', sb.GetString());
+  xe::LogLine(xe::LogLevel::Error, 'i', sb.GetString());
```
gibbed: I don't think we want to output all module information like this. It would be preferable to log the title ID of the main module once it's available, before any module dumps occur.

src/xenia/kernel/kernel_state.cc
```diff
@@ -394,6 +394,9 @@ object_ref<UserModule> KernelState::LoadUserModule(const char* raw_name,
     user_modules_.push_back(module);
   }
 
+  xe::LogLineFormat(xe::LogLevel::Error, 'i', "Title ID : %.8X\n",
```
gibbed: What's happening now for games that load multiple modules?

IceTDrinker: It would log the title ID each time LoadUserModule is called, but then it would also log the module info in the dump function called right after so...

gibbed: That's not what I mean, are we going to get "Title ID" output for each module loaded?

Games can have multiple modules, for the relevance of this logging, we only care about the initial module loaded's title ID.

IceTDrinker: "It would log the title ID each time LoadUserModule"

gibbed: Sorry, I wasn't clear. What I meant was, this isn't desired behavior.

IceTDrinker: What do you propose then to handled this ? This seems the earliest part of the code where we can get title ID info.

Something like a LogOnce ?

jgoyvaerts: Why don't you do it somewhere here: https://github.com/xenia-project/xenia/blob/master/src/xenia/emulator.cc#L626 ?

IceTDrinker: Is this part of the code called only once per game launch ?
The reason is I don't know the code base that well, so I put it somewhere where it worked

IceTDrinker: Moved the logging where you suggested

DrChat:
So, ideally we wouldn't be logging the title information at an error level, but maybe instead we could have an "always output" info level or something?

IceTDrinker:
I was told logging levels were fine when I mentioned an additional logging level on the discord server.

But yeah in theory I agree with you.

PR #1369, opened 2019-05-08
Flush the logger before exiting on a fatal error

jgoyvaerts:
We manually call the logger destructor when a fatal error occurs, this fixes the error not showing up in the log file when a fatal error is triggered

### DIFF
--- a/src/xenia/base/logging.cc
+++ b/src/xenia/base/logging.cc
@@ -325,7 +325,7 @@ void FatalError(const char* fmt, ...) {
                 MB_OK | MB_ICONERROR | MB_APPLMODAL | MB_SETFOREGROUND);
   }
 #endif  // WIN32
-
+  ShutdownLogging();
   exit(1);
 }
 


Razzile:
would it be better semantically to have a Logger::Flush() method which can be called both manually and through the destructor? That way we can avoid calling the destructor directly

jgoyvaerts:
No strong opinion either way, but I agree it seems weird to call a destructor in a random method

gibbed:
I would agree, if `Flush` is currently publicly accessible it would be clearer to call that instead.

jgoyvaerts:
There was already a method called ShutdownLogging so I used that one instead.

2 ways to test this fix:
- Launch xenia with a command line argument that is not a valid game file (for example ./xenia.exe xenia.pdb
- Launch xenia with an config file which cannot be parsed

In both cases, the error doesn't show up in the xenia.log, but after this fix it does.

PR #1373, opened 2019-05-13
ppc disasm: BC formatting improvements

elad335:
Add extended mnemonics of BC instruction for the disassmebler.

How it looks:
image


gibbed:
The code changes look fine, but the code style needs to be fixed to match the rest of the project. Unfortunately, clang-format is turned off for this file.

Because of that, I'm willing to make necessary changes to bring it in line, if you're OK with that?

elad335:
Yes, go ahead.

PR #1383, opened 2019-06-21
Update imgui to latest stable (v1.71)

Prism019:
Use the latest stable dear imgui version. Lots of breaking changes have happened so I've changed what needed to be changed. Everything seems to work fine from my testing, including the UI test program, so there shouldn't be a problem at all.


TGLTClutch:
latest is v1.72 now 😮  https://github.com/ocornut/imgui/commit/ecb9b1e2eba5becc38019b5c9f75fba711cd881b

PR #1386, opened 2019-06-25
[xenia-build] Update old link and fix VS Preview not being found

Margen67:
-prerelease allows vswhere to find preview/prerelease versions of Visual Studio, fixing xb with Visual Studio Preview.
https://github.com/microsoft/vswhere/issues/185#issuecomment-499962802


DrChat:
LGTM! Appreciated :)

PR #1388, opened 2019-06-28
Search for Python3 specifically in xb.bat

chris-hawley:
This commit serves to address the issue that everyone coming in with the new build of windows 10 will have with xb. I also added 3.7 to list of python versions we manually check for now.

### DIFF
--- a/xb.bat
+++ b/xb.bat
@@ -1,4 +1,4 @@
-@ECHO OFF
+@echo off
 REM Copyright 2015 Ben Vanik. All Rights Reserved.
 
 SET DIR=%~dp0
@@ -29,12 +29,10 @@ REM Utilities
 REM ============================================================================
 
 :check_python
-SETLOCAL
+setlocal enabledelayedexpansion
 SET FOUND_PYTHON_EXE=""
-1>NUL 2>NUL CMD /c where python3
-IF NOT ERRORLEVEL 1 (
-  SET FOUND_PYTHON_EXE=python3
-)
+
+
 IF %FOUND_PYTHON_EXE% EQU "" (
   IF EXIST c:\\python34\\python.exe SET FOUND_PYTHON_EXE=C:\\python34\\python.exe
 )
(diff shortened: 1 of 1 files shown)


chris-hawley:
fixed this so it properly works around the problem

chris-hawley:
fixed the logical flaw with the previous version

xb.bat
```diff
@@ -29,12 +29,27 @@ REM Utilities
...
 
 :check_python
-SETLOCAL
+setlocal enabledelayedexpansion
 SET FOUND_PYTHON_EXE=""
-1>NUL 2>NUL CMD /c where python3
-IF NOT ERRORLEVEL 1 (
-  SET FOUND_PYTHON_EXE=python3
+FOR /F %%I IN ('where python') DO (
+	CMD /C %%I -c "import sys; sys.exit(1 if not sys.version_info[:2] >= (3, 4) else 0)"
```
DrChat: Why Python 3.4 in particular? I'd imagine we could do just Python v3, no?

xb.bat
```diff
@@ -29,12 +29,27 @@ REM Utilities
...
+	IF ERRORLEVEL 0 (
+	  SET _RESULT=0
+	  SET PYTHON_EXE=%%I
+	  ENDLOCAL enabledelayedexpansion
+	  GOTO :eof
+	)
 )
+FOR /F %%I IN ('where python3') DO (
+	CMD /C %%I -c "import sys; sys.exit(1 if not sys.version_info[:2] >= (3, 4) else 0)"
```
DrChat: Not awfully familiar with bash, but maybe we can abstract this into its own function just so the check isn't duplicated across the two executables. Something like `check_python_version 3` or such.

xb.bat
chris-hawley: that's what was there before. I'd imagine there was a good reason for choosing that

xb.bat
chris-hawley: Not really sure what you mean here. The check won't be duplicated. It'll check all python3.exe's in PATH and then all python.exe's in PATH

chris-hawley:
This fails in the build system and I'm not entirely sure why. We can just detect that it's in the build system and hard code a python path..or just assume that it's there.

xb.bat
DrChat: Yeah - I'm just referring to code duplication. Though this is batch, so I'm not really sure if functions are that elegant either.

chris-hawley: I'm not too familiar with batch either. There might be a more elegant solution but I'm not aware of it. Well..I supposed I could create another goto and have it work that way but that seems equally as bad.

DrChat: Yeah, that's alright - not a big deal

xb.bat
```diff
@@ -29,12 +29,27 @@ REM Utilities
 REM ============================================================================
 
 :check_python
-SETLOCAL
+setlocal enabledelayedexpansion
 SET FOUND_PYTHON_EXE=""
-1>NUL 2>NUL CMD /c where python3
-IF NOT ERRORLEVEL 1 (
-  SET FOUND_PYTHON_EXE=python3
+FOR /F %%I IN ('where python') DO (
```
Brinsky: Thanks for this fix - I tried it out and ran into issues because my paths have spaces in them, which caused FOR to truncate them. The following fix (based on https://stackoverflow.com/a/5553316) worked for me:
```suggestion
FOR /F "delims=" %%I IN ('where python') DO (
```
If you'd like to confirm this for yourself with an example, run
```
(ECHO example one & ECHO demo two) > test.txt 
FOR /F %A IN (test.txt) DO ECHO %A 
```
and then add `"delims="` after the `/F` and try it again.

xb.bat
```diff
@@ -29,12 +29,27 @@ REM Utilities
...
+	CMD /C %%I -c "import sys; sys.exit(1 if not sys.version_info[:2] >= (3, 4) else 0)"
+	IF ERRORLEVEL 0 (
+	  SET _RESULT=0
+	  SET PYTHON_EXE=%%I
+	  ENDLOCAL enabledelayedexpansion
+	  GOTO :eof
+	)
 )
+FOR /F %%I IN ('where python3') DO (
```
Brinsky: Same as above:
```suggestion
FOR /F "delims=" %%I IN ('where python3') DO (
```

xb.bat
Brinsky: Looking at the continuous integration failure, I think it's caused by this same issue.

chris-hawley:
@Brinsky Thank you for your contribution. The fixes didn't actually help the build system.

Brinsky:
Too bad. I suppose you could temporarily remove `@ECHO OFF` and let AppVeyor run again to see if that makes the issue clear? I briefly looked into running AppVeyor locally, but it sounds like a pain and seems to require a project-specific authorization key.

chris-hawley:
rip appveyor. I'll just put in the detector for it being ran on there and map it to the environment variable that they provide

chris-hawley:
I'm not sure why this keeps failing in appveyor. At this point, I'm not entirely sure it's worth fixing since it breaks appveyor. If some piece of this can be adopted into the existing script to improve it, let me know. However, I don't think pursuing this potential bug is going to fruitful without someone more knowledgeable about batch.

PR #1395, opened 2019-07-09
List of recently opened files

ObsidianNA:
RecentsMenuAlpha

Overview:
Added a Pop-up menu named 'Open Recent' under the 'File' menu bar item that lists 10 recently opened files.

- It stores the paths of opened games inside a file named 'recent.txt' in the same folder where the xenia.exe is located.

- If the file doesn't exits, the program creates it, then paths are stored when a new file is opened.

- Each time a file is opened, the program compares stored paths with the new one, if it exists then the path is moved to the top of the list to indicate that it is the most recent file opened.


gibbed:
You need to squash your commits, you've completely obliterated any ability to view diffs with the way you just deleted two files then readded them with changes.

PR #1396, opened 2019-07-09
Recent Files Menu Bar

ObsidianNA:
RecentsMenuAlpha

Overview: 
Added a Pop-up menu named 'Open Recent' under the 'File' menu bar item that lists 10 recently opened files.

- It stores the paths of opened games inside a file named 'recent.txt' in the same folder where the xenia.exe is located.

- If the file doesn't exits, the program creates it, then paths are stored when a new file is opened.

- Each time a file is opened, the program compares stored paths with the new one, if it exists then the path is moved to the top of the list to indicate that it is the most recent file opened.

- 'Open Recent' items refresh once xenia is re-started

### DIFF
--- a/src/xenia/gpu/spirv_builder.cc
+++ b/src/xenia/gpu/spirv_builder.cc
@@ -203,5 +203,95 @@ spv::Id SpirvBuilder::IfBuilder::createMergePhi(spv::Id then_variable,
                               getElsePhiParent());
 }
 
+SpirvBuilder::SwitchBuilder::SwitchBuilder(spv::Id selector,
+                                           unsigned int selection_control,
+                                           SpirvBuilder& builder)
+    : builder_(builder),
+      selector_(selector),
+      selection_control_(selection_control),
+      function_(builder.getBuildPoint()->getParent()),
+      header_block_(builder.getBuildPoint()),
+      default_phi_parent_(builder.getBuildPoint()->getId()) {
+  merge_block_ = new spv::Block(builder_.getUniqueId(), function_);
+}
+
+void SpirvBuilder::SwitchBuilder::makeBeginDefault() {
+  assert_null(default_block_);
+
+  endSegment();
+
+  default_block_ = new spv::Block(builder_.getUniqueId(), function_);
+  function_.addBlock(default_block_);
+  default_block_->addPredecessor(header_block_);
+  builder_.setBuildPoint(default_block_);
+
+  current_branch_ = Branch::kDefault;
+}
+
+void SpirvBuilder::SwitchBuilder::makeBeginCase(unsigned int literal) {
+  endSegment();
+
+  auto case_block = new spv::Block(builder_.getUniqueId(), function_);
+  function_.addBlock(case_block);
+  cases_.emplace_back(literal, case_block->getId());
+  case_block->addPredecessor(header_block_);
+  builder_.setBuildPoint(case_block);
+
+  current_branch_ = Branch::kCase;
+}
+
+void SpirvBuilder::SwitchBuilder::addCurrentCaseLiteral(unsigned int literal) {
+  assert_true(current_branch_ == Branch::kCase);
+
+  cases_.emplace_back(literal, cases_.back().second);
+}
+
+void SpirvBuilder::SwitchBuilder::makeEndSwitch() {
+  endSegment();
+
+  builder_.setBuildPoint(header_block_);
+
+  builder_.createSelectionMerge(merge_block_, selection_control_);
+
+  std::unique_ptr<spv::Instruction> switch_instruction =
+      std::make_unique<spv::Instruction>(spv::OpSwitch);
+  switch_instruction->addIdOperand(selector_);
+  if (default_block_) {
+    switch_instruction->addIdOperand(default_block_->getId());
+  } else {
+    switch_instruction->addIdOperand(merge_block_->getId());
+    merge_block_->addPredecessor(header_block_);
+  }
+  for (const std::pair<unsigned int, spv::Id>& case_pair : cases_) {
+    switch_instruction->addImmediateOperand(case_pair.first);
+    switch_instruction->addIdOperand(case_pair.second);
+  }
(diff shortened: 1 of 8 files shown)


illusion98:
Should be in a single commit now 😺

src/xenia/app/emulator_window.cc
```diff
@@ -174,10 +182,71 @@ bool EmulatorWindow::Initialize() {
...
+  if (LINES[9] != "" && LINES[9] != "\n")
+    recent_list->AddChild(
+        MenuItem::Create(MenuItem::Type::kString, GLOBAL_RECENT_PATHS[9],
+                         std::bind(&EmulatorWindow::RecentList9, this)));
+//****************************************************************************
   {
     file_menu->AddChild(
         MenuItem::Create(MenuItem::Type::kString, L"&Open...", L"Ctrl+O",
                          std::bind(&EmulatorWindow::FileOpen, this)));
+    // Add Recent List Tab ti file tab
```
Margen67: *to

src/xenia/app/emulator_window.cc
```diff
@@ -174,10 +182,71 @@ bool EmulatorWindow::Initialize() {
   // FIXME: This code is really messy.
   auto main_menu = MenuItem::Create(MenuItem::Type::kNormal);
   auto file_menu = MenuItem::Create(MenuItem::Type::kPopup, L"&File");
+  // Recent Sub tap
+  auto recent_list = MenuItem::Create(MenuItem::Type::kPopup, L"&Open Recent");
+  //******Add Recent PATHS to menu***************************************
+  std::string LINES[10];
+  std::ifstream RECENT_FILE("recent.txt");
+  int i = 0;
+  // Conver from UTF-8 to whatever tf
```
Margen67: *Convert

src/xenia/app/emulator_window.cc
```diff
@@ -174,10 +182,71 @@ bool EmulatorWindow::Initialize() {
...
+  std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
```
gibbed: We have helper functions for this. `xe::to_string`/`xe::to_wstring`.

src/xenia/app/emulator_window.cc
```diff
@@ -174,10 +182,71 @@ bool EmulatorWindow::Initialize() {
...
+      GLOBAL_RECENT_PATHS[i] = converter.from_bytes(LINES[i]);
+      i++;
+    }
+    RECENT_FILE.close();
+  }
+
+  // Added most recently opened files to menu
+  if (LINES[0] != "" && LINES[0] != "\n")
+    recent_list->AddChild(
+        MenuItem::Create(MenuItem::Type::kString, GLOBAL_RECENT_PATHS[0],
```
gibbed: This should all be a for loop. And std::bind can take static arguments, ie: `std::bind(&EmulatorWindow::RecentList, this, 0)`.

src/xenia/app/emulator_window.cc
```diff
@@ -174,10 +182,71 @@ bool EmulatorWindow::Initialize() {
   // FIXME: This code is really messy.
   auto main_menu = MenuItem::Create(MenuItem::Type::kNormal);
   auto file_menu = MenuItem::Create(MenuItem::Type::kPopup, L"&File");
+  // Recent Sub tap
+  auto recent_list = MenuItem::Create(MenuItem::Type::kPopup, L"&Open Recent");
+  //******Add Recent PATHS to menu***************************************
+  std::string LINES[10];
```
gibbed: Please follow naming conventions. This applies to all variables and members introduced by your changes. I won't make this comment for every instance.

src/xenia/app/emulator_window.cc
```diff
@@ -174,10 +182,71 @@ bool EmulatorWindow::Initialize() {
...
+  std::ifstream RECENT_FILE("recent.txt");
```
gibbed: There's things to be said about how you're creating this file here, but I will leave this out until other problems are resolved.

src/xenia/app/emulator_window.cc
ObsidianNA: is the last parameter (0) in 'std::bind(&EmulatorWindow::RecentList, this, 0)' passed to the function?

gibbed: Yes.

gibbed: https://en.cppreference.com/w/cpp/utility/functional/bind

src/xenia/app/emulator_window.cc
ObsidianNA: Variable LINES[10] should be lines[10]? Or recent_list to recent_menu?

src/xenia/app/emulator_window.cc
ObsidianNA: ifstream is used to read the file if it exists. If it doesn't then ofstream creates it when a file is opened. Is there still a better way of doing this, if so how?

src/xenia/app/emulator_window.cc
ObsidianNA: Code is using helper functions now

src/xenia/app/emulator_window.cc
ObsidianNA: I read the naming conversion and made the changes. Could you check the code to let me know that there are no more instances of unconventional naming please?

src/xenia/app/emulator_window.cc
```diff
@@ -174,10 +176,37 @@ bool EmulatorWindow::Initialize() {
   // FIXME: This code is really messy.
   auto main_menu = MenuItem::Create(MenuItem::Type::kNormal);
   auto file_menu = MenuItem::Create(MenuItem::Type::kPopup, L"&File");
+  // Recent Sub tab
+  auto recent_list = MenuItem::Create(MenuItem::Type::kPopup, L"&Open Recent");
```
x1nixmzeng: Minor point, but here the keyboard alt-combo is assigned to alt+o which duplicates open - considering moving this to alt+r (`Open &Recent`)

PR #1399, opened 2019-07-17
[CPU] Use window size for LZX ref_data_size, fixes some problematic XEXP patches

emoose:
This minor change should fix some problematic patches which would cause "match offset beyond LZX stream" mspack errors.

Tested with Band of Bugs & Perfect Dark which both failed to apply any XEXP previously, now seem working with this fix.

### DIFF
--- a/src/xenia/cpu/lzx.cc
+++ b/src/xenia/cpu/lzx.cc
@@ -113,7 +113,7 @@ int lzx_decompress(const void* lzx_data, size_t lzx_len, void* dest,
       std::memset(lzxd->window, 0, window_data_len);
       std::memcpy(lzxd->window + (window_size - window_data_len), window_data,
                   window_data_len);
-      lzxd->ref_data_size = (uint32_t)window_data_len;
+      lzxd->ref_data_size = window_size;
     }
 
     result_code = lzxd_decompress(lzxd, (off_t)dest_len);


gibbed:
Isn't this going to allow a buffer overrun? If we do this shouldn't we be copying the window data into a `window_size sized` buffer?

emoose:
The window data buffer should already be sized to window_size AFAIK, the window_data_len value that was used previously was just the size of the data that would be copied into the end of the window.

It seemed to patch fine for me without any overruns, though I do still need to try something a bit larger than an XBLA game..

gibbed:
The length here isn't the length of the window though, it's the length of the data being copied into the window. So it'll end up copying past that.

emoose:
This shouldn't change anything being copied though? Only the ref_data_size used during LZX is being changed here, at least I can't see that being used for copying anywhere.

gibbed:
And that's the problem, this length is telling the LZX decoder how much data there is available there. Giving it a length more than what's actually there is the buffer overrun.

gibbed:
Er, I'm an idiot. I just realized I've been saying buffer overrun instead of buffer overread.

gibbed:
Okay. After reviewing the code properly I'm wrong on all counts. This is already happening when we memcpy in the buffer to the window. What threw me off was we're not memset-ing the correct length in the buffer.

PR #1400, opened 2019-07-18
Fix typo in xenia-build

Margen67:
Reasons to merge:
* One-line change
* Generates noise™


illusion98:
tiday -> tidy. Good eye

PR #1404, opened 2019-07-19
Clean up .appveyor.yml

Margen67:
Changes:
* Removed comments
* Changed `version` into `'{build}-{branch}'` since xenia doesn't really have a version number.
* Removed unneeded/unused stuff:
  * `clone_folder` (it defaults to `c:\dev\xenia` anyway)
  * environment variables (leftovers from example yml)
  * Removed `cmd: set PATH=C:\Program Files (x86)\MSBuild\15.0\Bin;%PATH%` from `install`
  * Removed `.bat` from `xb`
  * Removed `Platform: Windows` and replaced `%PLATFORM%` with `Windows` since the steps that use it are only for Windows. Proper cross-platform support might require this: https://www.appveyor.com/docs/build-configuration/#specializing-matrix-job-configuration (not that it matters since Linux on master is broken, and the linux branch is outdated)
* Put `after_build` in proper order and made it use AppVeyor's multi-line syntax.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,18 +1,9 @@
-#---------------------------------#
-#      general configuration      #
-#---------------------------------#
+version: '{build}-{branch}'
 
-# version format
-version: 1.0.{build}-{branch}
-
-# branches to build
-# Only building master (and pull requests)
 branches:
-  # blacklist
   except:
     - gh-pages
 
-# Do not build on tags (GitHub only)
 skip_tags: true
 
 skip_commits:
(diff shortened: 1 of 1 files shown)


illusion98:
😕 why closed?

gibbed:
It was four folders deep because of how he changed the add paths on `7z`.

@Margen67 Does AppVeyor support the versioning scheme you wanted to do here without barfing?

Margen67:
It might be doing that since it's a PR. This is how it looks on my AppVeyor: https://ci.appveyor.com/project/Margen67/xenia/builds/26130674

gibbed:
I don't think changing it away from the automated versioning is any different than using the build number. They both mechanically work the same. One looks nicer (version numbers).

We could just add the version number stuff from AppVeyor into the UI as well?

Margen67:
Maybe, but I don't know how to do that.
Also, are these lines necessary? https://github.com/xenia-project/xenia/blob/master/.appveyor.yml#L40-#L41

gibbed:
For safety, yes.

Margen67:
@gibbed GitHub won't let me reopen this. Can I make a new PR?

gibbed:
Yeah that's fine. I can add new comments there.

PR #1405, opened 2019-07-19
[cpu linux] Implement stack walking

bwrsandman:
Implement stack walking using libunwind and libiberty.
These two libraries would be need dependencies.
The libunwind library is quite common on a linux system and libiberty is
part of the GNU compiler.
Add tests for stack_walker.

This, in combination with the other xenia linux fixes allows the backtrace to be visible at runtime:
Screenshot from 2019-07-19 11-17-20


bwrsandman:
Test failing due to dependency on linux_threads

bwrsandman:
As with all the other PRs, you can test a combination https://github.com/bwrsandman/xenia/tree/linux

sl1pkn07:
conflicts with #1317 because create the same file .gdbint, but with different content

greetings

bwrsandman:
See bottom of the first comment of #1430 about that conflict.

bwrsandman:
It seems like the inclusion of libunwind causes travis to fail. I cannot replicate this on my system.

src/xenia/cpu/stack_walker_posix.cc
```diff
@@ -9,16 +9,293 @@
...
+};
+}  // namespace std
+
+namespace xe {
+namespace cpu {
+
+#include <libiberty/demangle.h>
+#include <memory>
+std::string demangle(const char* mangled_name) {
+  if (mangled_name == std::string()) {
```
Triang3l, line 123: For checking if an `std::string` is empty, `.empty()` can be used. In this case (a C string), `[0]` can be compared to 0.

src/xenia/cpu/stack_walker_posix.cc
```diff
@@ -9,16 +9,293 @@
...
+          return false;
+        case UNW_ENOINFO:
+          return false;
+        case UNW_ENOMEM:
+          return false;
+        default:
+          return false;
+      }
+      auto demangled_host_symbol_name = demangle(host_symbol_name.data());
+      std::strncpy(frame.host_symbol.name, demangled_host_symbol_name.c_str(),
```
Triang3l, line 276: `strncpy` is dangerous (doesn't insert a null terminator in case of truncation). Overall, C strings should be avoided unless it's a hard requirement to use a preallocated number of bytes for a string.

src/xenia/cpu/stack_walker_posix.cc
```diff
@@ -9,16 +9,293 @@
...
+namespace xe {
+namespace cpu {
+
+#include <libiberty/demangle.h>
+#include <memory>
+std::string demangle(const char* mangled_name) {
+  if (mangled_name == std::string()) {
+    return "";
+  }
+  std::unique_ptr<char, decltype(&std::free)> ptr(
```
Triang3l, line 126: This usage of `unique_ptr` is highly confusing, and looks a lot like that's use-after-free until we notice that the pointer is being used for implicit `std::string` initialization rather than returned directly. An explicit `std::free` call should express the intention much more clearly, I think, and is short enough (`std::free` is safely defined as a no-op for `nullptr`). Though copy elision of the returned string is another question, I'm not sure how it will work with explicit additional code after the creation of the string — if this is done specifically to save copy elision, and there are no more straightforward options to force it, this approach to freeing should be fine.

Triang3l (changes requested):

PR #1410, opened 2019-07-23
[Kernel] Add APC support to NtWriteFile

emoose:
This PR adds APC support to NtWriteFile, based on the APC code inside NtReadFile.

This is needed for some games when using mount_cache, such as Halo 4 and the Halo 3 Epsilon build, without it the game seems to get stuck waiting for something from the APC routine.

I've just copied over most of the body from NtReadFile here, which seems to work fine, but I'm not sure whether the code inside NtReadFile was just a temporary workaround or not... if it was maybe it'd be better to redo these properly instead.

(edit: again with travis issues :()


emoose:
Rebased onto latest master, this is still needed for Halo games to work with cache partitions properly (though they also need something like #1740 to work too - with #1849 being merged I'll look into making a new PR for that soon)

PR #1411, opened 2019-07-25
[CI] Add Azure Pipelines

Margen67:
Reasons to add:
* Can build both Release and Checked configurations at the same time
* Is faster than AppVeyor

https://margen67.visualstudio.com/xenia/_build/results?buildId=117

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,6 +12,7 @@ skip_commits:
     - .github/*
     - LICENSE
     - README.md
+    - .azure-pipelines.yml
     - .travis.yml
 
 skip_branch_with_pr: true
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -53,9 +53,9 @@ jobs:
     displayName: 'Tests'
 
   - script: |
-      cd build\bin\Windows\$(Configuration)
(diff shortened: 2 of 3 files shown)


illusion98:
Okay, looks like azure only allows teams to download artifacts.

TGLTClutch:
Close PR 🤔 it's useless then

Margen67:
Once Rick registers for Azure he can still download test results, along with any other registered users, so it's not completely useless.

JohnHolmesII:
May as well add the banner to the readme while you're at it

Margen67:
@JohnHolmesII Soon**™**

I'll add another note here: Azure doesn't have a URL for the latest artifacts like AppVeyor does, so I made a [feature request](https://developercommunity.visualstudio.com/idea/698386/download-latest-build-artifacts-via-url.html).

Margen67:
https://github.com/xenia-project/xenia/commit/7b0367da5f94d5cad0cada1993d5416d00015eae causes the build to fail. @gibbed

Margen67:
GitHub Actions (which is a fork of Azure) is coming out in 2 months, so I don't see a need for this when it's only going to be used for a limited amount of time.
GitHub Actions will also be easier to implement since it can be added by just merging a PR.

PR #1417, opened 2019-08-04
[Kernel] Make XamShow*UI exports send XN_SYS_UI notifications, & small XNotifyListener fix

emoose:
It seems both Project Sylpheed & Sonic Unleashed require seeing ```XN_SYS_UI = true```, eventually followed by ```XN_SYS_UI = false```, for the game to realise all system UIs are closed, but ATM adding a notification that already exists (before the game gets a chance to dequeue it) will just overwrite the older notifications value, stopping the game from ever getting to see the older notification.

To accommodate sending the two notifications to those games I've changed XNotifyListener::notifications_ to a std::vector instead, allowing for multiple notifications to share the same ID, which should let the game see both XN_SYS_UI notifications instead of just the later XN_SYS_UI = false one.

(it seems my older threaded-xoverlapped stuff could workaround this issue as I guess the game would have a chance to run XNotifyGetNext before we sent over the second one, IMO this fix is probably better though, since it doesn't rely on needing to create new threads/hoping the game calls XNotifyGetNext/etc)

I don't think this should cause any regressions, since it seems the only time notifications are ever queued up is when Xenia adds them itself, and AFAIK it was always intended for the game to actually see the notifications we send (in other words, I'm pretty sure nothing was depending on this replace-if-exists behavior)

(Also, we currently send multiple notifys with the same ID inside kernel_state already: https://github.com/xenia-project/xenia/blob/master/src/xenia/kernel/kernel_state.cc#L581, atm the two on/off statements are useless since the game will only ever see the second statement, so maybe the current behavior wasn't intended)

### DIFF
--- a/src/xenia/kernel/xam/xam_ui.cc
+++ b/src/xenia/kernel/xam/xam_ui.cc
@@ -111,6 +111,9 @@ dword_result_t XamShowMessageBoxUI(dword_t user_index, lpwstring_t title_ptr,
     buttons.push_back(button);
   }
 
+  // Broadcast XN_SYS_UI = true
+  kernel_state()->BroadcastNotification(0x9, true);
+
   uint32_t chosen_button;
   if (cvars::headless) {
     // Auto-pick the focused button.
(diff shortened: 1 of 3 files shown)


gibbed:
Notifications getting overwritten is definitely unintended by the intention of the original code you linked where it queues up a bunch of initial ones.

PR #1421, opened 2019-08-05
[CPU] Store ymm registers in thunks

Triang3l:
Since enabling AVX2 in MSVC added a lot of glitches to games, this is probably important.

I'm not sure though if vmovups is a good solution though, in cases when accesses cross cache lines (will happen every other 32-byte vmovups if they're unaligned), maybe we should try to align the stack (with one or two marker xmms for checking whether a padding was added).

### DIFF
--- a/src/xenia/cpu/backend/x64/x64_backend.cc
+++ b/src/xenia/cpu/backend/x64/x64_backend.cc
@@ -498,22 +498,22 @@ void X64ThunkEmitter::EmitSaveVolatileRegs() {
   mov(qword[rsp + offsetof(StackLayout::Thunk, r[5])], r10);
   mov(qword[rsp + offsetof(StackLayout::Thunk, r[6])], r11);
 
-  // vmovaps(qword[rsp + offsetof(StackLayout::Thunk, xmm[0])], xmm0);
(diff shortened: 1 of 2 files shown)


gibbed:
Stack is supposedly to be aligned to begin with?

Triang3l:
It's aligned to 16 bytes, not 32, if I understand correctly.

gibbed:
Yeah, that 16-byte alignment should be corrected to 32-bit here if we're going to store ymm instead of xmm.

gibbed:
Noticed this in the [documentation](https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2019#callercallee-saved-registers):

> The registers RAX, RCX, RDX, R8, R9, R10, R11, XMM0-5, and the **upper portions of YMM0-15** and ZMM0-15 are considered volatile and must be considered destroyed on function calls (unless otherwise safety-provable by analysis such as whole program optimization).

Thus, this shouldn't be necessary and if turning on AVX2 optimization is breaking something, it's _probably_ unrelated to this.

gibbed:
We do save off some `xmm` registers which are used for JIT temporary variables, but we never use them as `ymm` so that also should be unrelated.

PR #1423, opened 2019-08-07
Stagger thread creation when affinity not specified

randprint:
feels more smooth ¯\_(ツ)_/¯

### DIFF
--- a/src/xenia/kernel/xthread.cc
+++ b/src/xenia/kernel/xthread.cc
@@ -156,10 +156,12 @@ void XThread::set_name(const std::string& name) {
 }
 
 uint8_t next_cpu = 0;
+uint8_t fakedhwthreads[6] = {0,2,4,1,3,5};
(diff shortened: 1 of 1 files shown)


TGLTClutch:
is this a professional PR I'm seeing. 

👀👀👀

src/xenia/kernel/xam/user_profile.cc
```diff
@@ -18,8 +18,8 @@ namespace kernel {
 namespace xam {
 
 UserProfile::UserProfile() {
-  xuid_ = 0xBABEBABEBABEBABE;
-  name_ = "User";
+  xuid_ = 0xE000BABEBABEBABE;
```
Razzile: This change isn't related to the PR right?

randprint: oof no, clicked the wrong button

d4rkne55:
What happened with this?

PR #1424, opened 2019-08-07
Add support for logging transports 

Razzile:
Adds the concept of "transports" to the logging system which can be used to receive, store and write logs as they are received.

This doesn't really have much use in xenia as-is but would allow for things like a logging console window to be created by creating a console transport, for example.

This is a similar concept to the transport system found [in Winston](https://github.com/winstonjs/winston)

I still need to clean up the code a bit so keeping as a draft for now


Razzile:
I think this PR has everything I wanted to do in it now. Let me know if there's anything you would change

Razzile:
superseded by https://github.com/xenia-project/xenia/pull/1647

PR #1426, opened 2019-08-10
Add game_language as a config variable

jgoyvaerts:
Adds game_language as a variable in the config file.

I think adding it to the UI is pointless for now, since a new UI is in the works. This at least gives users the possibility to change it without having to compile xenia.


gibbed:
This isn't the only place language gets pulled from, but it'll be fine to add this to start.

PR #1428, opened 2019-08-12
Fix vpkd3d

psychoholic14:


Triang3l:
Seems like something else is affecting the order (maybe those parameters with shifts? Or other instructions) since the test (which was written by @benvanik, so I assume it was checked on a real Xenon) is failing now.

PR #1429, opened 2019-08-13
Improve Error Logging

TranzRail:
* Added new definitions for XELOGCORE for specific types of modules (such as CPU, GPU, APU, KERNEL).
* Added configuration flag to exclude error messages from specific types of modules (see above). Any logged message using the above new definitions can be excluded from the log by setting the corresponding config flag (in xenia.config.toml) to true (e.g. exclude_gpu_from_log = true).
* Improved error messages by adding what module generated it at the beginning of the string (e.g. "[GPU] Failed to read backbuffer.").
* Identified the function that generated the VdSwap log message, and tagged it with one of the new logging definitions (XELOG_GPU_I). This allows this message to be excluded from the log by setting 'exclude_gpu_from_log' to 'true'.

Rationale for the above changes is to improve the bug hunting experience. For example, instead of a few XELOGE calls, you can now see at a glance what the error is. The new format goes XELOG_%MODULE%_%ERROR_LEVEL% (e.g. 'XELOG_GPU_I' would be an Information level log entry for the GPU module.

In addition, the purpose of excluding items from the log is useful in the case where a particular XEX spews repetitive entries, thus generating large log files, while you might be working on a completely different module, making it harder to identify relevant entries to what you are working on.

If the PR is accepted, I'd be happy to go through and finish updating existing modules.

### DIFF
--- a/src/xenia/apu/audio_system.cc
+++ b/src/xenia/apu/audio_system.cc
@@ -252,7 +252,7 @@ bool AudioSystem::Save(ByteStream* stream) {
 
 bool AudioSystem::Restore(ByteStream* stream) {
   if (stream->Read<uint32_t>() != 'XAUD') {
-    XELOGE("AudioSystem::Restore - Invalid magic value!");
+    XELOG_APU_E("[APU] AudioSystem::Restore - Invalid magic value!");
     return false;
   }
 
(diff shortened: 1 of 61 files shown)


TranzRail:
Closed due to reworking based on dev comments.

d4rkne55:
"dev comments"? On Discord? As there's nothing here.
And what do you mean with reworking? Did the team already start this or would quite some changes be needed here so the team wants to take over?

illusion98:
Scroll up in #dev and you'll see it.

TranzRail:
> "dev comments"? On Discord? As there's nothing here.
> And what do you mean with reworking? Did the team already start this or would quite some changes be needed here so the team wants to take over?

Significant rework is required as several devs were either not convinced this would be useful, or the approach needs drastic improvement.

d4rkne55:
@TranzRail Thanks for the summary, I'm currently not in the Discord channel. Unfortunate, seems like a useful change to me. Maybe I will check the discussion shortly.

Issue #1430, opened 2019-08-14
Linux Support


bwrsandman:
## Current State

Xenia on Linux compiles for Travis tests in order to check C++ errors, style and very basic functionality. It has some nice abstract function definition. Most of these definitions are stubs with failing asserts which mean running Xenia in debug causes `SIGABRT` and running in release will eventually crash due to missing implementation.

The biggest missing part is the threading. Without threading implementation, the Xenia UI can't run and wait. The emulated threads and synchronization primitives which are built on top of these cannot work properly.

There is also PPC CPU emulation discrepancies between Windows and Linux. The registers used on 64-bit Windows and 64-bit Linux are not the same, therefore proper care must be made to use the correct registers when transitioning from guest to host code and vice-versa.

Memory mapping needs a bit of work due to the differences in shared memory and explicit memory ranges having different parameters and prefixes on Linux.

The GTK windowing and its interaction with Vulkan needs to fixed for several bugs.

The use of paths and the string functions need to better specify NT-style line endings (`\`) which are used by guest and Windows host. In the case of a Linux host, Unix line endings must be used when specifying guest paths.

Linux debugging and stack walking functions need to be implemented.

# Approach to Adding Support

My approach to fixing the following problem has been to first add unit tests (and to use the ones already there) to the already working Windows implementation in order to have a behavioural ground truth. Then I add implementation to Linux which satisfy these tests. Once the Linux implementation pass the tests, I activate those tests on Travis in order to prevent future regressions.

# Active PRs

Here are the PRs that I have worked on to have native Linux working. Some of these are built on top of the work of others and I preserved the authorship in the commits.
They are in descending order of importance.

- [ ] Threading: #1317 
- [ ] CPU PPC: #1339 
- [ ] Memory: #1397 
- [ ] Windowing, Filesystem and Misc: #1322
- [ ] Stack Walker: #1405 (depends directly on changes from #1317)

# New State

The fixes allow for all unit tests in the project to run and pass on Linux with clang.

The Xenia UI works and can select files with the open dialog.

Loading a xex will work up until emulation.

Calling a host function works as well as host calling a guest function.

Logging works.

Loading a ROM will eventually crash during emulation.

To see all the changes together, I maintain a branch on my repo: https://github.com/bwrsandman/xenia/tree/linux

If you are running Arch Linux, the folllowing AUR package installs it for you: https://aur.archlinux.org/packages/xenia-git/

Prism019:
[...]

PR #1431, opened 2019-08-14
Linux windowing fixes

bwrsandman:
Description
---
These are fixes to the windowing code in Linux. The changes help getting Xenia up and running but they need to be combined with the other Linux PRs to see the full working benefit. See #1430.

Changes
---
* Fix Vulkan on Xorg.
* GTK boots up without crashing, the file dialog works.
* Borrowed a commit from #1076.

Previous state
---
* Xenia window would crash in many ways but often due to a null pointer.
* The open file dialog was broken.
* Vulkan surface was not created properly on XOrg.

State after fix coupled with other linux fixes
---
* The xenia app boots up and you can select a ROM from filesystem.


sl1pkn07:
seems need rebase

greetings

src/xenia/ui/window_gtk.cc
```diff
@@ -160,7 +202,7 @@ void GTKWindow::set_focus(bool value) {
     if (value) {
       gtk_window_activate_focus(GTK_WINDOW(window_));
     } else {
-      // TODO(dougvj) Check to see if we need to do something here to unset
+      // TODO(dougvj) Check to see if we need to do somethign here to unset
```
Margen67, line 205: typo

bwrsandman: Good catch

JoelLinn:
Scrolling with mouse wheel seems to be inverted for me (tested with imgui y scrollbar).
Handled in `window_gtk.cc` in `case GDK_SCROLL:`.
I don't have a mouse with x scroll, can't test that.
It was also very slow stepped, one would need to consult the docs about the units of those values.

src/xenia/ui/window.h
```diff
@@ -172,8 +172,9 @@ class Window {
   Loop* loop_ = nullptr;
   std::unique_ptr<MenuItem> main_menu_;
   std::string title_;
-  int32_t width_ = 0;
-  int32_t height_ = 0;
+  // GTK must have a default value here that isn't 0
+  int32_t width_ = 1280;
```
Triang3l, line 176: Is this safe on Windows? It has:
```
  if (width != width_ || height != height_) {
    width_ = width;
    height_ = height;
    Layout();
  }
```
in WM_SIZE handling.

src/xenia/ui/window_gtk.cc
```diff
@@ -57,30 +52,81 @@ void gtk_event_handler_(GtkWidget* widget, GdkEvent* event, gpointer data) {
...
-  g_signal_connect(G_OBJECT(window_), "destroy", G_CALLBACK(gtk_main_quit),
-                   NULL);
-  g_signal_connect(G_OBJECT(window_), "event", G_CALLBACK(gtk_event_handler_),
+  gtk_window_set_resizable(GTK_WINDOW(window_), true);
+  gtk_window_set_title(GTK_WINDOW(window_), title_.c_str());
+  gtk_window_set_default_size(GTK_WINDOW(window_), width_, height_);
+  // Drawing area is where we will attach our vulkan/gl context
+  drawing_area_ = gtk_drawing_area_new();
+  // Don't allow resizing the window below this
+  gtk_widget_set_size_request(drawing_area_, 640, 480);
```
Triang3l, line 101: Xenia doesn't impose any size limit internally, this constraint is not needed.

Triang3l (changes requested):

src/xenia/ui/window.h
bwrsandman: I don't think this matters

src/xenia/ui/window_gtk.cc
```diff
@@ -350,18 +404,18 @@ bool GTKWindow::HandleKeyboard(GdkEventKey* event) {
...
   bool alt_pressed = modifiers & GDK_META_MASK;
   bool super_pressed = modifiers & GDK_SUPER_MASK;
+  uint32_t key_char = gdk_keyval_to_unicode(event->keyval);
   auto e =
       KeyEvent(this, event->hardware_keycode, 1, event->type == GDK_KEY_RELEASE,
                shift_pressed, ctrl_pressed, alt_pressed, super_pressed);
   switch (event->type) {
     case GDK_KEY_PRESS:
       OnKeyDown(&e);
+      if (key_char > 0) OnKeyChar(&e);
```
Triang3l, line 414: Xenia code style uses `{}` for all conditionals.

Triang3l (changes requested):

src/xenia/ui/window_gtk.cc
bwrsandman: Done in two places

Triang3l:
Seems to work pretty fine, and no issues with merging (though I haven't checked the Windows default width/height initialization yet, and its relation with things like DPI scaling, but overall we should be accepting the desired — though as a hint, not forced, as, for instance, on Android, the size will be taken from the layout instead — client area size when creating the window rather than relying on an explicit Resize call after the creation).

There's one issue though on Linux currently, the size passed to Resize is interpreted as including the menu — in the main Xenia window, with the Adwaita GNOME 3 theme, the client area is 1280x693 rather than 1280x720.

src/xenia/ui/window_gtk.cc
```diff
@@ -350,18 +405,20 @@ bool GTKWindow::HandleKeyboard(GdkEventKey* event) {
   bool ctrl_pressed = modifiers & GDK_CONTROL_MASK;
   bool alt_pressed = modifiers & GDK_META_MASK;
   bool super_pressed = modifiers & GDK_SUPER_MASK;
+  uint32_t key_char = gdk_keyval_to_unicode(event->keyval);
```
Triang3l, line 408: Is this unused, except for the `> 0` check? I think it should be written in the `KeyEvent` somewhere, or is requested in some other place too?

Triang3l (changes requested):

src/xenia/ui/window_gtk.cc
```diff
@@ -57,30 +52,80 @@ void gtk_event_handler_(GtkWidget* widget, GdkEvent* event, gpointer data) {
...
+                               reinterpret_cast<gpointer>(this), nullptr);
+  // Attach our event handler to both the main window (for keystrokes) and the
+  // drawing area (for mouse input, resize event, etc)
+  g_signal_connect(G_OBJECT(drawing_area_), "event",
+                   G_CALLBACK(gtk_event_handler),
                    reinterpret_cast<gpointer>(this));
+
+  GdkDisplay* gdk_display = gtk_widget_get_display(window_);
+  assert(GDK_IS_X11_DISPLAY(gdk_display));
+  connection_ = XGetXCBConnection(gdk_x11_display_get_xdisplay(gdk_display));
```
Triang3l, line 112: Is that the only XCB/Xlib-specific part of our window handling? It seems that we can use Xlib directly instead of Xlib-XCB if we only need the display and the window.

bwrsandman: Perhaps     
```c++
connection_ = XGetXCBConnection(gdk_display);
```

Triang3l: Yes, this seems like the only place, so it looks like we can use VK_KHR_xlib_surface without any XCB dependencies.

PR #1432, opened 2019-08-14
Add license mask as a config variable

illusion98:
This at least gives users the possibility to change the license mask without having to compile xenia.

### DIFF
--- a/src/xenia/kernel/xam/xam_content.cc
+++ b/src/xenia/kernel/xam/xam_content.cc
@@ -14,6 +14,14 @@
 #include "xenia/kernel/xenumerator.h"
 #include "xenia/xbox.h"
 
+DEFINE_int32(license_mask, 0,
+             "Set license mask for activated content: "
+             "0 - disable all licenses / "
+             "1 - enable the first license - usually the full version license "
+             "in Xbox Live Arcade games / "
+             "-1 or 0xFFFFFFFF - enable all possible licenses.", 
+             "Content");
+
 namespace xe {
 namespace kernel {
 namespace xam {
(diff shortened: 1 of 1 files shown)


src/xenia/kernel/xam/xam_content.cc
```diff
@@ -13,7 +13,7 @@
 #include "xenia/kernel/xam/xam_private.h"
 #include "xenia/kernel/xenumerator.h"
 #include "xenia/xbox.h"
-
+DEFINE_int32(license_mask, 0, "Toggle license for XBLA games", "Kernel");
```
gibbed: Description needs to be adjusted. It's not a toggle, it's a bit set of activated licenses. Something like `Set license mask for activated content`. I'm not sure how to make this easily understood though.

Category should also be `Content`, not `Kernel`.

src/xenia/kernel/xam/xam_content.cc
```diff
@@ -49,7 +49,7 @@ dword_result_t XamContentGetLicenseMask(lpdword_t mask_ptr,
   // Each bit in the mask represents a granted license. Available licenses
   // seems to vary from game to game, but most appear to use bit 0 to indicate
   // if the game is purchased or not.
-  *mask_ptr = 0;
+  *mask_ptr = (cvars::license_mask);
```
gibbed: Needs a cast. `static_cast<uint32_t>(cvars::license_mask)` I think should work?

illusion98: Resolved.

src/xenia/kernel/xam/xam_content.cc
illusion98: Changed to `Set license mask for activated content` and added value description from @Triang3l

illusion98:
Changed to your suggestion. Should be ready to merge now.

gibbed:
One final thing: reword your commit so it has `[Kernel]` in front of the message.

illusion98:
Added [Kernel] to commit message.

TGLTClutch:
Why is Appveyor so slow 
bild

PR #1433, opened 2019-08-14
Linux filesystem

bwrsandman:
These are filesystem and string changes to the emulator to work better on Linux.  The changes help getting Xenia up and running but they need to be combined with the other Linux PRs to see the full working benefit. See #1430.

Changes
---
* Paths utilities are tested for linux.
* GTK boots up without crashing, the file dialog works.
* Logging works when loading an xex.
* Borrowed two commits from #1076.

Previous state
---
* Xenia was failing to log as it was creating a directory called `xenia.log` and was unable to open a file there.
* The open file dialog was broken.

State after fix coupled with other Linux fixes
---
* The xenia app boots up and you can select a ROM from filesystem.
* Xenia will load the ROM and logs data about it. It also logs guest-to-host calls.


x1nixmzeng:
This looks good; in future it may be worth defining the expected path seperator used by the xbox somewhere useful

The tests you've added could perhaps be expanded to test resolving canonicalized paths correctly

sl1pkn07:
seems need rebase

greetings

bwrsandman:
Removed the use of templates to unify wstrings and strings as was mentioned by one of the maintainers in an earlier PR.

src/xenia/base/filesystem_posix.cc
```diff
@@ -156,7 +153,7 @@ bool CreateFile(const std::filesystem::path& path) {
 
 bool DeleteFile(const std::filesystem::path& path) {
   // TODO: proper implementation.
-  return (path.c_str()) == 0 ? true : false;
+  return path.c_str() == nullptr;
```
JoelLinn, line 156: [`unlink()`](https://linux.die.net/man/2/unlink) ?

src/xenia/base/logging.cc
```diff
@@ -30,16 +30,15 @@
 
 // For MessageBox:
 // TODO(benvanik): generic API? logging_win.cc?
+// TODO(dougvj): For now I just inlined a small difference for linux
 #if XE_PLATFORM_WIN32
 #include "xenia/base/platform_win.h"
 #endif  // XE_PLATFORM_WIN32
 
-#include "third_party/fmt/include/fmt/format.h"
```
JoelLinn, line 37: Removed by accident?

src/xenia/base/testing/string_test.cc
```diff
@@ -0,0 +1,39 @@
...
+  // TODO(bwrsandman):
+  REQUIRE(false);
+}
+
+TEST_CASE("StringBuffer") {
+  StringBuffer sb;
+  uint32_t module_flags = 0x1000000;
+
+  std::string path_(R"(\Device\Cdrom0\default.xex)");
+  sb.AppendFormat("Module %s:\n", path_.c_str());
```
JoelLinn, line 29: Needs fix for new {fmt} library.

src/xenia/base/testing/string_test.cc
```diff
@@ -0,0 +1,39 @@
...
+  REQUIRE(sb.to_string() == "Module \\Device\\Cdrom0\\default.xex:\n");
+  sb.AppendFormat("    Module Flags: %.8X\n", module_flags);
```
JoelLinn, line 31: {fmt}
maybe best to search the changesets for `%`s

src/xenia/base/filesystem_posix.cc
```diff
@@ -122,18 +121,16 @@ static int removeCallback(const char* fpath, const struct stat* sb,
...
+static uint64_t convertUnixtimeToWinFiletime(const timespec& unixtime) {
+  // Linux uses number of nanoseconds since 1/1/1970, and Windows uses
   // number of nanoseconds since 1/1/1601
-  // so we convert linux time to nanoseconds and then add the number of
-  // nanoseconds from 1601 to 1970
+  // so we add the number of nanoseconds from 1601 to 1970
   // see https://msdn.microsoft.com/en-us/library/ms724228
-  uint64_t filetime = filetime = (unixtime * 10000000) + 116444736000000000;
+  uint64_t filetime =
+      (unixtime.tv_sec * 10000000) + unixtime.tv_nsec + 116444736000000000;
```
JoelLinn, line 133: It's supposed to be in 100ns intervals.
`(unixtime.tv_sec * 10000000) + unixtime.tv_nsec / 100 + 116444736000000000`

src/xenia/base/filesystem_posix.cc
bwrsandman: [remove()](https://en.cppreference.com/w/cpp/filesystem/remove)?

JoelLinn: Ok, Std is (almost) always better than sys functions

JoelLinn: Although I am not sure if we should allow folders getting deleted which remove does. (Contrary to unlink). 
Check if its a folder and return false? (And some log message)

bwrsandman: I'm wondering if there's any point to having this function anymore.
There are [one place](https://github.com/xenia-project/xenia/blob/master/src/xenia/vfs/devices/host_path_entry.cc#L105) where `DeleteFile` is called and it is specifically expecting `path` to  not be a directory.

It could easily be replaced with
```c++
// Delete File.
return !std::filesystem::is_directory(full_path) && std::filesystem::remove(full_path);
```

Unless there is a reason not to do this, I would like to remove `IsFolder`, `PathExists`, `CreateFolder`, `DeleteFolder` and `DeleteFile` with `std::filesystem` functions in the place where they are called.

I realize that there might be a reason not to do this. Slippery slope argument: the same thing could be done with file handle functions and even with `std::threads`. So I am experimenting and waiting for feedback.

JoelLinn: About the slippery slope:
It depends. For example I decided against using `std::chrono::high_resolution_clock` for the rework of the clock code because I wanted full control over the arithmetics. Using `std` lib could have resulted in two rescales. The risk of that going ooopsie in some games was too high for my taste.
There could be similar arguments with `std::threads` for example. Reading std documentation may not be enough, for chrono stuff I even peeked into the MS implementation of it...

bwrsandman: PR for those functions is up
https://github.com/xenia-project/xenia/pull/1589

src/xenia/kernel/util/object_table.cc
```diff
@@ -21,6 +21,23 @@ namespace xe {
 namespace kernel {
 namespace util {
 
+namespace {
+#if defined(XE_PLATFORM_LINUX)
+#include <cxxabi.h>
+#include <memory>
+std::string demangle(const char* mangled_name) {
```
Triang3l, line 28: As demangling may be useful in other places, it'd possibly be nice to have this somewhere in `xenia/base`.

JoelLinn (commented):
There is some things duplicated from other PRs so I didn't comment on those, assuming they will be discussed/fixed and merged first.

src/xenia/kernel/util/object_table.cc
bwrsandman: Done

JoelLinn:
@bwrsandman You can remove the last clock commit from this PR, it's in master now

Triang3l:
> @bwrsandman You can remove the last clock commit from this PR, it's in master now
Done! :)

PR #1434, opened 2019-08-14
[CPU] Use c++ style cast

bwrsandman:


bwrsandman:
This is a small fix I had leftover from my Linux fixes.
It's out of scope but I didn't want to lose it.

bwrsandman:
Done

PR #1435, opened 2019-08-14
Implement debug functions for Linux

bwrsandman:
Check `TracerPid` in `/proc/self/status` for attached debugger.
Add `SIGTRAP` handler to prevent signal from halting app while not running
in a debugger.
Log `DebugPrint` in `clog` (`stderr`).

### DIFF
--- a/src/xenia/base/debugging_posix.cc
+++ b/src/xenia/base/debugging_posix.cc
@@ -9,21 +9,51 @@
 
 #include "xenia/base/debugging.h"
 
-#include <signal.h>
+#include <csignal>
 #include <cstdarg>
+#include <fstream>
+#include <iostream>
+#include <mutex>
+#include <sstream>
 
 #include "xenia/base/string_buffer.h"
 
 namespace xe {
 namespace debugging {
 
-bool IsDebuggerAttached() { return false; }
-void Break() { raise(SIGTRAP); }
+bool IsDebuggerAttached() {
+  std::ifstream proc_status_stream("/proc/self/status");
+  if (!proc_status_stream.is_open()) {
+    return false;
+  }
+  std::string line;
+  while (std::getline(proc_status_stream, line)) {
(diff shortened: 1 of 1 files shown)


src/xenia/base/debugging_posix.cc
```diff
@@ -11,14 +11,46 @@
...
+}
+
+static bool SigTrapInstalled = false;
+static void SigTrapHandler(int signum) {
+  signal(SIGTRAP, SIG_DFL);
+  SigTrapInstalled = true;
+}
+
+void Break() {
+  if (!SigTrapInstalled) {
```
DrChat: Use [`std::call_once`](https://en.cppreference.com/w/cpp/thread/call_once) for this instead.

bwrsandman: Updated and using `<csignal>` header instead.

src/xenia/base/debugging_posix.cc
```diff
@@ -9,21 +9,51 @@
...
 #include "xenia/base/string_buffer.h"
 
 namespace xe {
 namespace debugging {
 
-bool IsDebuggerAttached() { return false; }
-void Break() { raise(SIGTRAP); }
+bool IsDebuggerAttached() {
+  std::ifstream proc_status_stream("/proc/self/status");
+  if (proc_status_stream.is_open()) {
```
JoelLinn, line 26: I think you meant `if (!proc_status_stream.is_open()) {` instead

JoelLinn (changes requested):
Although it's a small changeset, I tested all these functions and there indeed was an issue. With that fixed everything works.

My recommendation is to merge this immediately after the typo is fixed.

src/xenia/base/debugging_posix.cc
bwrsandman: Done

PR #1440, opened 2019-08-18
[Kernel] Exports & fixes needed for Forza 4 saving/loading

emoose:
This adds two functions used by Forza 4 when saving (XeKeysObscureKey & XeKeysAesCbcUsingKey), a stub for an XLiveBase message that it needs to return success, and a fix for the game truncating save files by accident.

The two functions should work pretty much like they do inside the HV, though on X360 they use a per-console key for the obscuring-crypto, I've just set that to an all zero key in Xenia.

The XLiveBase message doesn't seem to write anything to the message buffer, just reads something from it AFAIK. Forza needs it to return success for the profile to be recognised so I've just added a stub for it.

The truncation bug was a tricky one, but I've narrowed it down to how we handle the file size of an XFile.
When writing data to a save, the game uses NtQueryInfoFile to make sure there's enough room in the file for the data.
If there isn't, the game uses NtSetInfoFile(EndOfFileInfo) to set the length of the file to (position + data size)
Unfortunately after NtSetInfoFile(EndOfFileInfo) is ran we don't update the information inside vfs::Entry, the same information that NtQueryInfo retrieves when the game tries to get the file size...

This makes the game always think that there isn't enough space in the file whenever it writes something, even though the game explicitly set the EOF to 0x5000 when the file was opened!

The result of this is that important save data gets truncated when the game tries updating info inside the saves header, eg:
- sets EOF to 0x5000
- writes 8 byte header
- writes variable-size data
- sets position to 4, to update data-size in header
- uses NtQueryInformationFile to check if the file is big enough to write the data-size in
- if not, calls NtSetInformationFile(EndOfFileInfo) to set end of file to data-size offset + 4 (0x8)

Because the size wasn't updated when we set the EOF, NtQueryInformationFile returns size 0, making the game set a new EOF which deletes any data written past the 8 byte header

To fix this I've added an update() method to vfs::Entry, and made NtQueryInfo/NtSetInfo call that before doing anything size-related with the vfs::Entry.
This seemed like the best way to solve this, seeing as the vfs::Entry doesn't have anything to do with the opened file handle, but I'm open to any other suggestions for it of course!

Anyway, with all this Forza 4 should now load/save with no issues.
I'm curious if this might help with other games - maybe the whole NtQueryInfo/NtSetInfo stuff comes from some library function that other games share, probably worth checking out.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
@@ -402,6 +402,9 @@ dword_result_t NtSetInformationFile(
         assert_true(length == 8);
         auto eof = xe::load_and_swap<uint64_t>(file_info);
         result = file->SetLength(eof);
+
+        // Update the files vfs::Entry information
+        file->entry()->update();
         break;
       }
       case XFileCompletionInformation: {
@@ -484,6 +487,11 @@ dword_result_t NtQueryInformationFile(
         //   ULONG         Unknown;
         // };
         assert_true(length == 56);
+
+        // Make sure we're working with up-to-date information, just in case the
+        // file size has changed via something other than NtSetInfoFile
+        // (eg. seems NtWriteFile might extend the file in some cases)
+        file->entry()->update();
 
         auto file_info = file_info_ptr.as<X_FILE_NETWORK_OPEN_INFORMATION*>();
         file_info->creation_time = file->entry()->create_timestamp();
--- a/src/xenia/kernel/xam/apps/xlivebase_app.cc
+++ b/src/xenia/kernel/xam/apps/xlivebase_app.cc
@@ -55,6 +55,14 @@ X_RESULT XLiveBaseApp::DispatchMessageSync(uint32_t message,
              buffer_length);
       return X_STATUS_UNSUCCESSFUL;
     }
+    case 0x00058046: {
+      // Required to be successful for Forza 4 to detect signed-in profile
+      // Doesn't seem to set anything in the given buffer, probably only takes
+      // input
+      XELOGD("XLiveBaseUnk58046(%.8X, %.8X) unimplemented", buffer_ptr,
+             buffer_length);
+      return X_ERROR_SUCCESS;
+    }
   }
   XELOGE(
       "Unimplemented XLIVEBASE message app=%.8X, msg=%.8X, arg1=%.8X, "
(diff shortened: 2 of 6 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
```diff
@@ -491,6 +488,9 @@ dword_result_t NtQueryInformationFile(
         // };
         assert_true(length == 56);
 
+        // Make sure we're working with up-to-date information
```
gibbed: Shouldn't this no longer be necessary with commit 4b27aa1fca3063b3c4b3ff47ed7a7870af98e131?

emoose: I'm not really sure, mostly kept that there in case anything else besides SetFileInfo changes the file size, eg. Windows docs on NtWriteFile mention that it can extend it under some circumstances, but I don't know if that's the case on 360 (if so, why does Forza do all this weird stuff with QueryFileInfo/SetFileInfo?)

IMO it's probably best to keep it there just to be safe, but if you like I can remove it, pretty sure Forza should still work without it.

emoose:
Squashed as requested.

src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
gibbed: Could you add this rational to a comment next to this then? For future reference.

emoose: Sure thing, extended that comment with some more context + squashed it.

gibbed: I feel like nothing has changed with the last commit?

emoose: I guess I could mention that stuff about NtWriteFile too, would something like this be better?

```
// Make sure we're working with up-to-date information, just in case the
// file size has changed via something other than NtSetInfoFile
// (eg. seems NtWriteFile might extend the file in some cases)
```

gibbed: Oh I just meant it seemed like the new force push didn't have any new actual changes compared to what it was previously.

That'd probably be a nice note, yes.

emoose: Oh right lol, that was just a quick typo fix, NtSetFileInfo -> NtSetInfoFile.

Added the note about NtWriteFile & squashed again.

PR #1444, opened 2019-08-19
[Kernel] XBLA Fixes

randprint:
[kernel] 

Updated PR from Permanulled https://github.com/xenia-project/xenia/pull/1150/files#diff-358a33abb42660137d33300bbc8188c8R51

XGI messages to fix Jetpac XBLA

gets Aegis Wing, Jetpac, and MetalSlug 3 in game

### DIFF
--- a/src/xenia/kernel/xam/apps/xgi_app.cc
+++ b/src/xenia/kernel/xam/apps/xgi_app.cc
@@ -2,7 +2,7 @@
  ******************************************************************************
  * Xenia : Xbox 360 Emulator Research Project                                 *
  ******************************************************************************
- * Copyright 2015 Ben Vanik. All rights reserved.                             *
+ * Copyright 2021 Ben Vanik. All rights reserved.                             *
  * Released under the BSD license - see LICENSE in the root for more details. *
  ******************************************************************************
  */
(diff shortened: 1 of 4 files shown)


Margen67:
Two things:
Rebase and squash so it doesn't conflict and have unneeded commits
The commit should say [Kernel] not [kernel]

illusion98:
Force pushed.

randprint:
updated to correct behaviour of NetDll_WSARecvFrom() to bring State of Decay in game

src/xenia/kernel/xam/user_profile.cc
```diff
@@ -18,8 +18,8 @@ namespace kernel {
 namespace xam {
 
 UserProfile::UserProfile() {
-  xuid_ = 0xBABEBABEBABEBABE;
-  name_ = "User";
+  xuid_ = 0xE000BABEBABEBABE;
+  name_ = "Xenia";
```
Razzile: This should probably be moved to a separate issue/PR as it isn't related to this PR

randprint: The Xuid change is most definitely related. The name, admittedly, is just cosmetic and could be left out from a merge. I'll probably leave it as-is until Rick addresses it.

PR #1448, opened 2019-08-24
[XMA] Fixed sound issues related to guest<->physical addresses mismatch

Gliniak:
Specific games stores guest addresses instead of physical. We need to convert it again into physical addresses.

context_array_ptr always stores physical address, but context_ptr not, that difference was causing unknown XMA registers to show up. After switching into physical address they disappeared.

(For more info I can provide screenshot of memory allocated for XMA_CONTEXT_DATA)

Benefits:
Sound works again in _Skate_ series games (and in every game that after implementing 0xE range lost audio) plus _Forza motorsport_ goes ingame again 

Casualities:
Unknown/Not visible


illusion98:
Motorsport 2 is now in-game again!

PR #1450, opened 2019-08-24
VFS changes to allow definition of local paths to devices

randprint:
VFS changes to allow definition of local paths to devices in the guest in Xenia as groundwork for relaunching games or apps from dashboards etc.


src/xenia/vfs/device.h
```diff
@@ -31,7 +31,7 @@ class Device {
   virtual bool is_read_only() const { return true; }
 
   virtual void Dump(StringBuffer* string_buffer) = 0;
-  virtual Entry* ResolvePath(const std::string& path) = 0;
+  virtual Entry* ResolvePath(std::string path) = 0;
```
Triang3l, line 34: These all should take string_view, not a copy of the string.

src/xenia/emulator.h
```diff
@@ -116,6 +116,12 @@ class Emulator {
   // Terminates the currently running title.
   X_STATUS TerminateTitle();
 
+  // takes array of emulated device paths, local paths, and symlinked drives
+  // handles initializing emulated devices
+  X_STATUS SetupVFSDevice(std::wstring devicepath, std::wstring localpath,
```
Triang3l, line 121: Arguments not on the same column, also these should take u16string views, not wstrings.

src/xenia/emulator.cc
```diff
@@ -47,6 +49,15 @@
 DEFINE_double(time_scalar, 1.0,
               "Scalar used to speed or slow time (1x, 2x, 1/2x, etc).",
               "General");
+DEFINE_string(HDD0_root, "", "Path to folder for HDD1.",
```
Triang3l, line 52: Config variables are normally lowercase.

Triang3l (changes requested):

PR #1452, opened 2019-08-25
[Kernel] distribute worker threads not created with affinity specified more evenly

randprint:
[Kernel] distribute worker threads not created with affinity specified more evenly


Triang3l:
If these are guest threads, we should possibly assign them to the same core as the one that created them.

https://docs.microsoft.com/en-us/windows/win32/dxtecharts/coding-for-multiple-cores

"On Xbox 360, you must explicitly assign software threads to a particular hardware thread by using XSetThreadProcessor. Otherwise, all child threads will stay on the same hardware thread as the parent."

Though I'm not aware what specific input should trigger such behavior, but an empty affinity mask on creation seems like one of the possible ways.

PR #1457, opened 2019-08-27
Add window dimension cvars

randprint:
add window dimensions to config.toml to make it easy to choose default window size at launch

### DIFF
--- a/src/xenia/app/emulator_window.cc
+++ b/src/xenia/app/emulator_window.cc
@@ -29,6 +29,9 @@
 
 DECLARE_bool(debug);
 
+DEFINE_int32(window_height, 720, "Window height", "UI");
+DEFINE_int32(window_width, 1280, "Window width", "UI");
+
 namespace xe {
 namespace app {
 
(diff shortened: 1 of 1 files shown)


src/xenia/app/emulator_window.cc
```diff
@@ -29,6 +29,9 @@
 
 DECLARE_bool(debug);
 
+DEFINE_int32(window_height, 720, "Window height", "UI");
```
Triang3l, line 32: 1280x720 hardcoded cvar defaults are not the way it should be handled, because that would block the addition of 480p and 1080p video output mode settings.

Triang3l (changes requested):

src/xenia/app/emulator_window.cc
randprint: This was just for convenience to have to resize the window every time I launch on a 4k screen. I don't know if it's even still applicable that xenia just stretches to fill the window without respecting aspect ratio now that you added the safezones. It would be a nice QOL feature if you can think of a better way to do it. As well as if Xenia would remember window position because my hack in fix for that is even uglier so I don't have to drag the window across the screen every time I launch too.

PR #1459, opened 2019-08-31
Removal of MAX_PATH limit on Windows

CookiePLMonster:
This PR lifts `MAX_PATH` limit from all code places which still used it. Most of current instances of it could be trivially removed either by calling functions twice (first to determine required buffer size, then to do actual work) or by replacing them with their memory-allocating counterparts.

The end result is a complete removal of `kMaxPath` constant on Windows. It's now unused and I think it's a good idea not to keep it to encourage not using any hardcoded constants for functions operating on paths.

### DIFF
--- a/src/xenia/base/platform.h
+++ b/src/xenia/base/platform.h
@@ -95,7 +95,6 @@ namespace xe {
 #if XE_PLATFORM_WIN32
 const char kPathSeparator = '\\';
 const wchar_t kWPathSeparator = L'\\';
-const size_t kMaxPath = 260;  // _MAX_PATH
 #else
 const char kPathSeparator = '/';
 const wchar_t kWPathSeparator = L'/';
--- a/src/xenia/gpu/shader.cc
+++ b/src/xenia/gpu/shader.cc
@@ -49,12 +49,21 @@ std::pair<std::string, std::string> Shader::Dump(const std::string& base_path,
     xe::filesystem::CreateFolder(target_path);
   }
 
-  char txt_file_name[kMaxPath];
-  std::snprintf(txt_file_name, xe::countof(txt_file_name),
-                "%s/shader_%s_%.16" PRIX64 ".%s", base_path.c_str(),
-                path_prefix, ucode_data_hash_,
-                shader_type_ == ShaderType::kVertex ? "vert" : "frag");
-  FILE* f = fopen(txt_file_name, "wb");
+  const std::string file_name_no_extension =
+      xe::format_string("%s/shader_%s_%.16" PRIX64, base_path.c_str(),
+                        path_prefix, ucode_data_hash_);
+
+  std::string txt_file_name, bin_file_name;
+
+  if (shader_type_ == ShaderType::kVertex) {
+    txt_file_name = file_name_no_extension + ".vert";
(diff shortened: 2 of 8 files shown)


src/xenia/gpu/shader.cc
```diff
@@ -49,41 +49,58 @@ std::pair<std::string, std::string> Shader::Dump(const std::string& base_path,
...
-      fprintf(f, "// ");
-      fwrite(ucode_disasm_ptr, 1, line_end - ucode_disasm_ptr + 1, f);
-      ucode_disasm_ptr = line_end + 1;
-    }
-    fprintf(f, "\n\n");
-    if (!host_disassembly_.empty()) {
-      fprintf(f, "\n\n/*\n%s\n*/\n", host_disassembly_.c_str());
+  std::string txt_file_name;
+  int required_size =
+      std::snprintf(nullptr, 0, "%s/shader_%s_%.16" PRIX64 ".%s",
```
gibbed: The way you ended up changing it here is excessively verbose. I think it would be better to `std::snprintf` the data hash into a static buffer then concatenate the string as normal instead of this giant repetition of code.

src/xenia/base/platform.h
```diff
@@ -95,7 +95,6 @@ namespace xe {
 #if XE_PLATFORM_WIN32
 const char kPathSeparator = '\\';
 const wchar_t kWPathSeparator = L'\\';
-const size_t kMaxPath = 260;  // _MAX_PATH
```
gibbed: Please split this into a separate commit.

src/xenia/gpu/shader.cc
CookiePLMonster: Can a reliable assumption be made about max length of this hash? If so, I can of course do it.

gibbed: `"%.16" PRIX64` is `%.16llx` -- so yes.

CookiePLMonster: Yes, but `path_prefix` doesn't seem to have size defined, so I would only save on reformatting `ucode_data_hash_` and maybe extension. Is it worth it?

CookiePLMonster: Nevermind, gave this a second look after a while and now I get what do you mean!

CookiePLMonster:
Concerns addressed, commits reworded.

src/xenia/gpu/shader.cc
```diff
@@ -49,12 +60,20 @@ std::pair<std::string, std::string> Shader::Dump(const std::string& base_path,
...
   }
 
-  char txt_file_name[kMaxPath];
-  std::snprintf(txt_file_name, xe::countof(txt_file_name),
-                "%s/shader_%s_%.16" PRIX64 ".%s", base_path.c_str(),
-                path_prefix, ucode_data_hash_,
-                shader_type_ == ShaderType::kVertex ? "vert" : "frag");
-  FILE* f = fopen(txt_file_name, "wb");
+  const std::string file_name_no_extension =
+      MakeShaderDumpPath(base_path, path_prefix, ucode_data_hash_);
```
DrChat: Use `xe::format_string` for this instead

CookiePLMonster:
Rebased and addressed review comments.

PR #1461, opened 2019-09-01
Fix SHA1 and SHA256 state endianness

CookiePLMonster:
Fixes malfunctioning SHA256 hash, corrupting its state between Update and Finish calls.

**Some** code paths were performing byteswapping from host to guest endianness (`XeCryptSha256Init`), but some were not (`XeCryptSha256Update`, `XeCryptSha256Final`), so SHA256 object was left in a corrupted state and hashing yielded wrong results.

Huge thanks to @Gliniak who scanned the logs and identified SHA256 functions as a possible culprit - I'd spend hours looking into I/O functions otherwise :)

Makes Forza Horizon go ingame:
image

**Question:** I have similar doubts about DES3 where `XECRYPT_DES_STATE` contains `uint32_t`'s in host endianness, but it's a guest structure. However, I'm not sure what's their true data type - they're defined as
```
uint32_t keytab[16][2];
```

but the way DES3 uses those fields implies it's an array of `uint64_t`'s.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
@@ -88,20 +88,16 @@ static_assert_size(XECRYPT_SHA_STATE, 0x58);
 
 void InitSha1(sha1::SHA1* sha, const XECRYPT_SHA_STATE* state) {
   uint32_t digest[5];
-  for (int i = 0; i < 5; i++) {
-    digest[i] = state->state[i];
-  }
+  std::copy(std::begin(state->state), std::end(state->state), digest);
 
   sha->init(digest, state->buffer, state->count);
 }
 
-void StoreSha1(sha1::SHA1* sha, XECRYPT_SHA_STATE* state) {
-  for (int i = 0; i < 5; i++) {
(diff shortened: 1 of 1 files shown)


gandalfthewhite19890404:
Causes MGS2/3 hang at selection screen.

CookiePLMonster:
> Causes MGS2/3 hang at selection screen.

Can you post logs (with level 3) from both **without** and **with** this PR? Assuming I implemented everything correctly this PR is nearly impossible to cause regressions, as it fixes something which can be trivially proven to be a bug. It's possible that hashing working correctly unlocked some other broken code path for those games.

gandalfthewhite19890404:
> > Causes MGS2/3 hang at selection screen.
> 
> Can you post logs (with level 3) from both **without** and **with** this PR? Assuming I implemented everything correctly this PR is nearly impossible to cause regressions, as it fixes something which can be trivially proven to be a bug. It's possible that hashing working correctly unlocked some other broken code path for those games.

Sorry, it was my mistake, I'll delete my post.

CookiePLMonster:
I have force pushed a fix for SHA1 implementation, existing one was incorrectly promoting bytes to dwords and was writing past the array.

This prepares for an upcoming PR getting Forza Horizon 2 to work.

PR #1462, opened 2019-09-01
Added new 0xFE app message

Gliniak:
### Overview:
Sometimes games uses not defined xam exports as a 0xFE app with different msgid.

Usually lack of even dummy support for such action causes game to return I/O Error or crash.

### Affected Games (by msgid):
0x21012:
- _FUSE_

0x22005: most EGO engine games
- _Grid 2_
- _Grid Autosport_
- _Toybox Turbos_


Gliniak:
Closed by request.

Changes in `XNotifyGetNext` requires further investigation

PR #1464, opened 2019-09-02
[Kernel] Add PVR opcode (includes cvars support)

TranzRail:
Adds the PVR opcode to the list of recognized opcodes. Includes useful comments.

### DIFF
--- a/src/xenia/cpu/cpu_flags.cc
+++ b/src/xenia/cpu/cpu_flags.cc
@@ -48,3 +48,6 @@ DEFINE_bool(break_condition_truncate, true, "truncate value to 32-bits", "CPU");
 
 DEFINE_bool(break_on_debugbreak, true, "int3 on JITed __debugbreak requests.",
             "CPU");
+
+// Other:
+DEFINE_uint64(pvr, 0x710700, "Processor version & revision number.", "CPU");
--- a/src/xenia/cpu/cpu_flags.h
+++ b/src/xenia/cpu/cpu_flags.h
@@ -25,7 +25,7 @@ DECLARE_bool(trace_function_data);
 DECLARE_bool(disable_global_lock);
 
 DECLARE_bool(validate_hir);
-
+// Breakpoints:
 DECLARE_uint64(break_on_instruction);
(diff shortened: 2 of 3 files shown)


TranzRail:
Should note that there's a conflict with #1243 but that's only because I've re-implemented work user: emoose already did (prior to me seeing that commit).

halotroop2288:
2 years, several months old and not a single review... ***bump***

Triang3l:
Since the branch appears to have been deleted, and I can't make comments directly, here are two suggestions here:

- Everything aside from the guest breakpoint is already "other", I think it'd be better to put that cvar above the breakpoint ones (and in `cpu_flags.h` as well).
- Unless an ampersand is a part of a name of some third-party entity, when there are no tight space limitations, the word "and" is preferable instead.
- The supported values for a cvar should be listed in the cvar description (as multiple `\n`-separated lines), not only in the code comment, so they're visible in the config file. As values are stored in the config as decimal (though we can add the functionality for changing the output format to hexadecimal in config storing later since that seems to be useful), both 0x-prefixed hexadecimal and decimal values should be specified in the cvar description if it's a bit field. On the code side, if there is a known set of numeric values with specific meanings, an `enum class` (or, for a bit field `struct`, which is preferred here as there already are descriptions of the bit ranges, `constexpr` instances of the `constexpr`-constructed structure) would improve readability and reusability of the values. The "Note: Some XEXs (such as mfgbootlauncher.xex) may check for a value that's less than 0x710700" observation should be in both the cvar description (so the user can see the expected side effects and the constraints) and the code comments (not to introduce a reverse dependency of the documentation of the internal logic on the existence of such documentation in the config user interface).

halotroop2288:
Since the branch was deleted, I'd like to revive the PR and add your changes. Is there any way to checkout this commit so I can preserve git blame?

TranzRail:
Thanks for fixing this PR. Unfortunately I don't have the repo and/or files locally anymore but to my recollection you haven't lost anything there.

gibbed:
Superseded by #1961.

PR #1465, opened 2019-09-02
[Kernel] Improve XexUnloadImage and module refcounts

CookiePLMonster:
This PR consists of two fairly separate parts:

1) The first commit corrects a few places in XEX loader where modules were incorrectly reference counted. Most notably, `object_ref`, which refcounts objects automatically, was used alongside manual `Retain()` calls. This probably caused handle leaks - there might be more to go, but at least some of them are fixed now.
2) The second commit fixes implementation flaws in `XexUnloadImage`. Previously, releasing the last reference to a module performed only a partial cleanup. This resulted in issues such as:
  - Attempting to notify an inexistant module when sending `DLL_THREAD_ATTACH` and `DLL_THREAD_DETACH` events. In my experience those were non-fatal but results may vary per game.
  - Not loading a XEX module fully after a chain of `Load` -> `Unload` -> `Load` calls. This actually was the main culprit of crashes I was trying to tackle.

When combined with #1461, this PR allows Forza Horizon to proceed past the first race, potentially making a good part of the game reachable and playable!
image

### DIFF
--- a/src/xenia/kernel/kernel_state.cc
+++ b/src/xenia/kernel/kernel_state.cc
@@ -405,6 +405,39 @@ object_ref<UserModule> KernelState::LoadUserModule(const char* raw_name,
   return module;
 }
 
+void KernelState::UnloadUserModule(const object_ref<UserModule>& module,
+                                   bool call_entry) {
+  auto global_lock = global_critical_region_.Acquire();
+
+  if (module->is_dll_module() && module->entry_point() && call_entry) {
+    // Call DllMain(DLL_PROCESS_DETACH):
+    // https://msdn.microsoft.com/en-us/library/windows/desktop/ms682583%28v=vs.85%29.aspx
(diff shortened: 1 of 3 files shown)


CookiePLMonster:
I have updated the PR with a change from
`object_table()->RemoveHandle(module->handle());`
to
`object_table()->ReleaseHandle(module->handle());`

As far as I can tell this doesn't change functionality at all, but now it properly decreases a handle reference counter from 1 to 0 before downloading - so it makes debugging easier, as it is valid to expect RemoveHandle to be called when handle refcount is 0.

CookiePLMonster:
Pushed a fixup for another "stray handle" remaining from when `LoadUserModule` fails to load (thanks @Gliniak for spotting this) - by calling `ReleaseHandle` instead of `RemoveHandle`, reference counters get decremented gracefully and module gets released this way.

PR #1473, opened 2019-09-08
[Kernel] Fixup GetProcAddressByOrdinal

CookiePLMonster:
A minor fix eyeballed while discussing related functionality. Fixes two tiny issues in `KernelModule::GetProcAddressByOrdinal`:

1. Now properly takes a global lock. Previously return value from `global_critical_region_.Acquire()` was discarded, which effectively locked and immediately unlocked the global lock.
2. Cases where a trampoline for ordinal was already created called `guest_trampoline_map_.find(ordinal)` twice for no reason. Now it's only called once - potentially a micro-optimization.

Chance of affecting games in any measurable way is near zero, unless there is a game which furiously calls `XexGetProcedureAddress` on the same module **concurrently** - in which case this pre-emptively prevents a race condition on populating the guest trampolines map.

### DIFF
--- a/src/xenia/cpu/export_resolver.cc
+++ b/src/xenia/cpu/export_resolver.cc
@@ -58,7 +58,7 @@ Export* ExportResolver::GetExportByOrdinal(const std::string_view module_name,
                                            uint16_t ordinal) {
   for (const auto& table : tables_) {
     if (xe::utf8::starts_with_case(module_name, table.module_name())) {
-      if (ordinal > table.exports_by_ordinal().size()) {
+      if (ordinal >= table.exports_by_ordinal().size()) {
         return nullptr;
       }
       return table.exports_by_ordinal().at(ordinal);
--- a/src/xenia/cpu/xex_module.cc
+++ b/src/xenia/cpu/xex_module.cc
@@ -127,7 +127,7 @@ uint32_t XexModule::GetProcAddress(uint16_t ordinal) const {
         xex_security_info()->export_table);
 
     ordinal -= export_table->base;
-    if (ordinal > export_table->count) {
+    if (ordinal >= export_table->count) {
       XELOGE("GetProcAddress({:03X}): ordinal out of bounds", ordinal);
       return 0;
     }
--- a/src/xenia/kernel/kernel_module.cc
+++ b/src/xenia/kernel/kernel_module.cc
@@ -107,12 +107,12 @@ uint32_t KernelModule::GetProcAddressByOrdinal(uint16_t ordinal) {
   } else {
     if (export_entry->function_data.trampoline ||
         export_entry->function_data.shim) {
-      global_critical_region_.Acquire();
+      auto global_lock = global_critical_region_.Acquire();
 
       // See if the function has been generated already.
-      if (guest_trampoline_map_.find(ordinal) != guest_trampoline_map_.end()) {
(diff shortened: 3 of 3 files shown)


CookiePLMonster:
Updated with an off-by-one fix in one of the functions `KernelModule::GetProcAddressByOrdinal` calls. With how the code was, calling `ordinals.at(ordinal)` while `ordinal == ordinals.size()` would have resulted in an exception.

Gliniak:
Small update:
I asked Cookie (Thanks) to add my change to this commit (very similar to his), as it is oneliner.

This will fix Bee Movie game.

Cause was simple. Ordinal have value 2 while count is also 2 (as elements: 0 and 1), so when he tried to read from array[2], he had a bad time.

legend800:
Looks like this is maybe causing freezes on loading gameplay in Need for Speed Most Wanted (2005). 

[xenia.zip](https://github.com/xenia-project/xenia/files/3588402/xenia.zip)

CookiePLMonster:
> Looks like this is maybe causing freezes on loading gameplay in Need for Speed Most Wanted (2005).

Are you able to compile this PR by yourself? If so, can you find out which of the two commits causes it?

ZolaKluke:
> Looks like this is maybe causing freezes on loading gameplay in Need for Speed Most Wanted (2005).
> 
> [xenia.zip](https://github.com/xenia-project/xenia/files/3588402/xenia.zip)

I tested it myself and it still works, you probably didn't do the workaround properly which is either delete the sound files or use timescalar 0.13x on loading into a race

src/xenia/kernel/kernel_module.cc
```diff
@@ -106,12 +106,12 @@ uint32_t KernelModule::GetProcAddressByOrdinal(uint16_t ordinal) {
...
         export_entry->function_data.shim) {
-      global_critical_region_.Acquire();
+      auto global_lock = global_critical_region_.Acquire();
 
       // See if the function has been generated already.
-      if (guest_trampoline_map_.find(ordinal) != guest_trampoline_map_.end()) {
-        auto entry = guest_trampoline_map_.find(ordinal);
-        return entry->second;
+      auto item = guest_trampoline_map_.find(ordinal);
+      if (item != guest_trampoline_map_.end()) {
```
DrChat: Hah, nice catch :)

src/xenia/cpu/export_resolver.cc
```diff
@@ -63,7 +63,7 @@ Export* ExportResolver::GetExportByOrdinal(const char* module_name,
   for (const auto& table : tables_) {
     if (std::strncmp(module_name, table.module_name(),
                      std::strlen(table.module_name())) == 0) {
-      if (ordinal > table.exports_by_ordinal().size()) {
+      if (ordinal >= table.exports_by_ordinal().size()) {
```
DrChat: Ditto, nice catch

src/xenia/cpu/xex_module.cc
```diff
@@ -126,7 +126,7 @@ uint32_t XexModule::GetProcAddress(uint16_t ordinal) const {
         xex_security_info()->export_table);
 
     ordinal -= export_table->base;
-    if (ordinal > export_table->count) {
+    if (ordinal >= export_table->count) {
```
DrChat: Ditto

PR #1474, opened 2019-09-08
[App] Fix discord cvar

Prism019:
The current placement of the check of the cvars to start the discord RPC (at the beginning of xenia_main) doesn't actually check the toml.

This PR moves the check to after the toml is imported and the cvars are populated, allowing `discord = false` in the toml to actually disable discord RPC.


Prism019:
@gibbed I updated the commit message. AppVeyor failed because this was updated back when AppVeyor broke, so this should still be safe to merge.

Prism019:
Closing and reopening to restart CI builds.

randprint:
tested, works great - thanks!

PR #1476, opened 2019-09-11
[XAM] Don't use strncpy for XamUserGetName.

gibbed:


src/xenia/kernel/xam/xam_user.cc
```diff
@@ -86,8 +86,19 @@ dword_result_t XamUserGetName(dword_t user_index, lpstring_t buffer,
...
+  if (!buffer_len) {
+    return X_ERROR_SUCCESS;
+  }
+
   const auto& user_profile = kernel_state()->user_profile();
-  std::strncpy(buffer, user_profile->name().data(), buffer_len);
+  const auto& user_name = user_profile->name();
+
+  // Real XAM will only copy a maximum of 15 characters out.
+  size_t copy_length = std::min(std::min(user_name.size(), size_t(15)),
```
CookiePLMonster: Is `std::min(std::initializer_list)` overload not available? Could probably simplify this to

```
std::min({user_name.size(), static_cast<size_t>(buffer_len - 1), size_t(15)});
```

CookiePLMonster: Also, constness - `user_profile`/`user_name` is const-ified so this could be too.

Issue #1419, opened 2019-08-04
cpptoml doesn't support UTF-8 files with a BOM


gibbed:
The TOML spec says a UTF-8 BOM should be supported.

Xenia won't end up writing a BOM to a config file on its own, but anyone editing the file by hand may end up adding one through that.

PR #1486, opened 2019-10-02
[Core] Fix reading UTF-8 BOM config files

Atvaark:
Since cpptoml appears to be dead I tried fixing #1419 by detecting and skipping a UTF-8 BOM before parsing the config file.

### DIFF
--- a/src/xenia/config.cc
+++ b/src/xenia/config.cc
@@ -15,6 +15,14 @@ std::shared_ptr<cpptoml::table> ParseFile(const std::wstring& filename) {
     throw cpptoml::parse_exception(xe::to_string(filename) +
                                    " could not be opened for parsing");
   }
+  // since cpptoml can't parse files with a UTF-8 BOM we need to skip them
+  char bom[3];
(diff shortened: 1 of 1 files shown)


src/xenia/config.cc
```diff
@@ -15,6 +15,14 @@ std::shared_ptr<cpptoml::table> ParseFile(const std::wstring& filename) {
     throw cpptoml::parse_exception(xe::to_string(filename) +
                                    " could not be opened for parsing");
   }
+  // since cpptoml can't parse files with a UTF-8 BOM we need to skip them
+  char bom[3];
+  file.read(bom, sizeof(bom));
```
CookiePLMonster: What if it reads less than 3 bytes?

Atvaark: Then `file.fail()` in the next line returns `true` and the stream is reset to position 0 before passing it to cpptoml.

DrChat:
We can create a fork for cpptoml on xenia-project and fix this there. Would you like to research a fix on that library directly instead?

PR #1488, opened 2019-10-09
[Base] File 'open_access' flag is uninitialised before use.

noshbar:
In the 'OpenExisting' function, 'open_access' is used to create the flags for opening the file.
However, it is not set to zero, so could end up being incorrect file modes, potentially even emptying the file for appending, which would be bad.

This simply initialises it to zero first.

### DIFF
--- a/src/xenia/base/filesystem_posix.cc
+++ b/src/xenia/base/filesystem_posix.cc
@@ -155,7 +155,7 @@ class PosixFileHandle : public FileHandle {
 
 std::unique_ptr<FileHandle> FileHandle::OpenExisting(std::wstring path,
                                                      uint32_t desired_access) {
-  int open_access;
+  int open_access = 0;
   if (desired_access & FileAccess::kGenericRead) {
     open_access |= O_RDONLY;
   }


noshbar:
Sorry for not using the correct format from the beginning, hopefully it's okay now.

gibbed:
It seems like you only updated the title of the PR? Not the commit message?

noshbar:
Right, I was being an idiot. Pull requests are still new to me, and I thought I could just change it "here" (I edited the title and -what turns out to be- my first comment).
I'll try do a proper job.

noshbar:
Yup, new to Git and contributing to others... thankfully no more Perforce and restrictive company policies for me!

DrChat:
I'm going to reword the commit and merge it.
"Initialize file 'open_access' flag before using it."

Not a huge deal, but prefer to word commit messages so that they describe the action taken, rather than the previous issue.
Otherwise thanks for spotting this!

DrChat:
I've also apparently managed to screw up this PR. Rest assured, it has been merged (and rebased).

PR #1493, opened 2019-10-20
[HID/Linux] Support Linux systems and non XInput controllers

JoelLinn:
xenia-hid-demo

## Current state on master
Input is Windows exclusive by _XInput_ compatible controllers or the _Keyboard_.
Both underlying APIs are Windows exclusive as well.

### Design decisions
The primary goal is to enable input on Linux.
The [SDL2](https://www.libsdl.org/) library was selected because of its clean API, active development, good availability on distributions as well as its wide controller support. As _SDL_ itself is a cross platform library, _xenia_ on windows benefits from its controller support and remap functionality. This implementation in turn can be fully tested on Windows.
### Implementation Completeness
All methods except for `GetKeystroke` are implemented. AFAIK that method is mainly used to query chatpad input but also includes controller input. Most games seem to use the `GetState` function though. Whether this is an issue would have to be studied in the field.
### SDL Integration
As this implementation is redundant to most Windows users, shipping the (comparably large) _SDL2_ library with _xenia_ is not desired. Also, static linking is [strongly discouraged](https://wiki.libsdl.org/Installation) by the authors. Therefore the library is dynamically linked with the addition of using [delayed loading on Windows](https://docs.microsoft.com/en-us/cpp/build/reference/linker-support-for-delay-loaded-dlls?view=vs-2019). When `SDL2.dll` can not be loaded the input driver will gracefully fail its `Setup()` routine - _xenia_ can be shipped without the dll which can be added if desired. [Official _SDL2_ development files](https://www.libsdl.org/download-2.0.php) with libraries for Windows have been added. On other platforms the build process uses the systems headers and libs (e.g. `libsdl2-dev` package on debian distros) and requires the library to be present at startup. This keeps the implementation clean and without explicit `LoadLibrary` or similar calls.
### Testing
This is tested with `xenia-hid-demo` and a couple of minutes of Red Dead Redemption gameplay. On Linux in order to run `xenia-hid-demo` the fixes from #1430 are needed. For that, merge this PR into @bwrsandman[s linux branch](https://github.com/bwrsandman/xenia/tree/linux).


Prism019:
Does this affect imgui at all? I know that currently imgui gets different keyvalues from gdk vs windows and it makes it nigh impossible to type anything.

bwrsandman:
Should be calling something like
```cpp
ImGui_ImplSDL2_ProcessEvent(&event);
```

JoelLinn:
Under the hood SDL initializes different stuff to catch those controller events. I was under the impression it ignores Keyboard events until they are explicitly enabled.
Where exactly are you having these issues?

Prism019:
Wait, are you saying you got in-game using this PR and bwrsandman's PRs? If so then this fix allows Linux to actually run games for the first time. Unless you meant you played RDR on Windows and not Linux.

JoelLinn:
**No ingame** on Linux, you can however test this PR in the hid demo on Linux.

See #1430 fox things missing.
Not to mention the Vulkan GPU backend is not really working great on Windows either (DX12 is win only)...

JoelLinn:
The SDL event processing is now done in the window loop thread using its `Post( )` function. Per documentation this is a requirement (although SDL internally does locking they still can't guarantee thread safety depending on the platform). Controller events are processed in a way that would allow another part of xenia to process different kind of events without interfering or the need for global SDL management code. SDL events to my understanding do not interfere with other window events.

I tried to keep locking minimal so that games do not get blocked by the window loop processing controller events. So no spikes in the profiler.

Rebased on latest master.

src/xenia/hid/sdlpad/sdlpad_input_driver.cc
```diff
@@ -0,0 +1,402 @@
...
+}
+
+X_STATUS SDLPadInputDriver::Setup() {
+  if (!TestSDLVersion()) {
+    return X_STATUS_UNSUCCESSFUL;
+  }
+
+  // SDL_PumpEvents should only be run in the thread that initialized SDL - we
+  // are hijacking the window loop thread for that.
+  window_->loop()->PostSynchronous([&]() {
```
bwrsandman: This actually requires linking with `xenia-ui`. This should probably be added to the premake file.

JoelLinn: mh, ok that link is missing on xinuput too

JoelLinn:
- Significantly reduced sdl event pumping by not queuing when previous pump has not finished.
- Renamed `sdlpad` to `sdl` since its already inside hid namespace.

JoelLinn:
Rebased onto master and retested.

PR #1494, opened 2019-10-22
Remove /gt & OpenGL leftovers

Margen67:
Fiber-safe optimizations actually worsen performance.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -90,9 +90,6 @@ filter({"configurations:Release", "platforms:Windows"})
   linkoptions({
     "/NODEFAULTLIB:MSVCRTD",
   })
-  buildoptions({
-    "/GT", -- enable fiber-safe optimizations
-   })
 
 filter("platforms:Linux")
   system("linux")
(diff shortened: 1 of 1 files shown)


premake5.lua
```diff
@@ -91,7 +91,7 @@ filter({"configurations:Release", "platforms:Windows"})
     "/NODEFAULTLIB:MSVCRTD",
   })
   buildoptions({
-    "/GT", -- enable fiber-safe optimizations
+    "/O2", -- Enable optimizations
```
chris-hawley: This is redundant with the speed optimization line. It's  `optimize("speed")`

premake5.lua
```diff
@@ -91,7 +91,7 @@ filter({"configurations:Release", "platforms:Windows"})
     "/NODEFAULTLIB:MSVCRTD",
   })
   buildoptions({
-    "/GT", -- enable fiber-safe optimizations
```
chris-hawley: You are correct, this should be removed.

PR #1497, opened 2019-10-28
[Kernel] Enable XEX1 loading

aerosoul94:
XEX1's are found in builds created before SDK 1888. Currently the only way to run them is by owning XeDK hardware (beta XDK), which aren't easy to come by. Though there still are more incompatibilities that need to be worked on, this allows them to be loaded and executed.

This fix is simple as the only difference appears to be in the security info structure (xex2_security_info). I just created a new structure named SecurityInfoContext which points to or loads all the info needed from the security info structure in the XEX header. 

Here are the changes made:
- Exception was made for the "XEX1" magic in `UserModule::LoadFromMemory()`.
- `xex1_security_info` structure was added to xex2_info.h
- `GetSecurityInfo()` now returns a void pointer which just points to either the XEX1 or XEX2 security info structure.
- Enum `XexFormat` was added to differentiate between XEX1 or XEX2, rather than relying on the magic.
- `SecurityInfoContext` was added to `XexModule`.
- `xex_security_info()` now returns the instance of `SecurityInfoContext`.

SecurityInfoContext's members share the same names as the security info structure, so no further changes were necessary within the rest of the loading process.

Here is a partial log provided as evidence of an XEX1 being loaded. Notice the SDK versions required are 1838: https://gist.github.com/aerosoul94/6ed631f38e041c16d8e11216bf9042f1

### DIFF
--- a/src/xenia/cpu/xex_module.h
+++ b/src/xenia/cpu/xex_module.h
@@ -43,6 +43,21 @@ class XexModule : public xe::cpu::Module {
     xe_xex2_version_t min_version;
     std::vector<ImportLibraryFn> imports;
   };
+  struct SecurityInfoContext {
+    const char* rsa_signature;
+    const char* aes_key;
+    uint32_t image_size;
+    uint32_t image_flags;
+    uint32_t export_table;
+    uint32_t load_address;
+    uint32_t page_descriptor_count;
+    const xex2_page_descriptor* page_descriptors;
+  };
+  enum XexFormat {
+    kFormatUnknown,
+    kFormatXex1,
+    kFormatXex2,
+  };
 
   XexModule(Processor* processor, kernel::KernelState* kernel_state);
   virtual ~XexModule();
@@ -51,8 +66,8 @@ class XexModule : public xe::cpu::Module {
   const xex2_header* xex_header() const {
     return reinterpret_cast<const xex2_header*>(xex_header_mem_.data());
   }
-  const xex2_security_info* xex_security_info() const {
-    return GetSecurityInfo(xex_header());
+  const SecurityInfoContext* xex_security_info() const {
+    return &security_info_;
   }
 
   uint32_t image_size() const {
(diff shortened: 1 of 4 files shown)


src/xenia/kernel/user_module.cc
```diff
@@ -135,7 +135,7 @@ X_STATUS UserModule::LoadFromMemory(const void* addr, const size_t length) {
   auto processor = kernel_state()->processor();
 
   auto magic = xe::load_and_swap<uint32_t>(addr);
-  if (magic == 'XEX2') {
+  if (magic == 'XEX2' || magic == 'XEX1') {
```
CookiePLMonster: Do you think it makes sense to compare just first three characters, aka `XEX` here? (with an appropriate mask)

CookiePLMonster:
Looks good, bar linting issues.

src/xenia/kernel/user_module.cc
aerosoul94: I had asked myself the same question but as I saw that it was comparing with `XEX2` I wasn't sure how specific it should be. There are actually a few more unhandled magics such as `XEX%`, `XEX-`, and `XEX?`.

CookiePLMonster: I actually wonder then if those are still 4-byte magics or 3-byte magics with some other data coming after them - but having unaligned data would be quite weird IMO.

src/xenia/cpu/xex_module.h
```diff
@@ -186,6 +200,9 @@ class XexModule : public xe::cpu::Module {
   uint32_t base_address_ = 0;
   uint32_t low_address_ = 0;
   uint32_t high_address_ = 0;
+
+  XexFormat xex_format_;
```
gibbed: You've left this uninitialized here, I think the best case here would be to have another enum (`kFormatXexUnknown`) to initialixe `xex_format_` to.

src/xenia/cpu/xex_module.h
```diff
@@ -186,6 +200,9 @@ class XexModule : public xe::cpu::Module {
...
+  SecurityInfoContext security_info_;
```
gibbed: Probably should be initialized to `{}` for clarity.

PR #1498, opened 2019-10-28
[APU/Linux] Support cross platform audio

JoelLinn:
## Current audio implementation
At the moment _XAudio2_ version 2.8 is used as the audio back-end.
_XAudio2_ is only available on Windows 8 and up.
### Library
In this PR, the [SDL2](https://www.libsdl.org/) library was used to implement another audio back-end for similar reasons as in #1493. It is much lighter than _XAudio2_. The _SDL2_ integration commit is is shared with that PR.
### State
I got sound working just fine with this PR and minimal code on Windows. I could however not test 6 channel audio since I just have headphones. Also, there is no frequency change when adjusting the CPU timer. Implementing that with _SDL2_ alone would be possible using the resampler but that would likely double the code complexity and only seems cosmetic to me. For Linux, I could just verify that it compiles.


JoelLinn:
Rebased onto master and retested.

src/xenia/app/premake5.lua
```diff
@@ -71,17 +72,24 @@ project("xenia-app")
       "xcb",
       "X11-xcb",
       "vulkan",
+      "SDL2",
     })
 
   filter("platforms:Windows")
     links({
+      "delayimp", -- Enable dll delayed loading for msvc
```
DrChat: Could you change the comment to something like "This library implements delayed loading on Windows" instead?
The current comment makes me think this is a linker option, which it isn't.

JoelLinn: Changed this and split linking related changes shared with #1493 into an extra commit.

PR #1506, opened 2019-11-15
[Kernel] Fixes to thread reference counters

CookiePLMonster:
This PR consists of three parts:

1) The first commit adds some useful assertions which I used to debug issues fixed in later commits.
2) Second commit stops `ExCreateThread` from retaining the handle. Despite the comment from 64b2be92d64361c437c6ba1de6f82296200be75b, I believe this handle should **not** be retained, as the thread is supposed to be freed once it finishes executing **and** it's handle is closed. Since every `XObject` contains a handle to itself on creation, retaining the handle further on creation would result in always having this one handle linger, and thus not freeing thread objects.

I have not extensively tested this but I've noticed threads started being properly deleted after intro videos in Forza Horizon 2. According to @Gliniak, numerous lingering threads in Lost Odyssey now close properly too.

3) The last commit changes logic of `ObXXX` functions to retain and release handles. This ties down to 2) and the case there `ExCreateThread` was created with `creation_flags & 0x80`, which results in returning a guest pointer and not a handle. To describe the problem, consider a typical scenario how games seem to use threads created with this flag (pay attention to bolded parts):
- Thread gets created. **Handle to thread gets created and stored internally.** Returned guest pointer gets saved.
- Guest pointer gets used to eg. modify thread priorities.
- Thread gets unpaused.
- Code waits for thread to finish.
- Code releases reference to the thread by calling `ObDereferenceObject`. This releases reference to the thread, releasing it. **Internal thread's handle does not get freed.**

As a result, this handle lingers making the game prone to crashing any time it gets used - for example, when pausing all threads to show a guest crash prompt!

My change makes the code operate on handles instead, so `ObDereferenceObject` releases it gracefully. For cases where `ObDereferenceObject` is used to release a pointer created by `ObReferenceObjectByHandle` or related, retaining the handle and not `XObject` allows to keep the symmetry of those actions and seems to be handled gracefully as well.

### DIFF
--- a/src/xenia/kernel/util/object_table.cc
+++ b/src/xenia/kernel/util/object_table.cc
@@ -194,6 +194,7 @@ X_STATUS ObjectTable::RemoveHandle(X_HANDLE handle) {
   if (entry->object) {
     auto object = entry->object;
     entry->object = nullptr;
+    assert_zero(entry->handle_ref_count);
     entry->handle_ref_count = 0;
 
     // Walk the object's handles and remove this one.
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
@@ -69,7 +69,7 @@ dword_result_t ObLookupThreadByThreadId(dword_t thread_id,
   }
 
   // Retain the object. Will be released in ObDereferenceObject.
-  thread->Retain();
+  thread->RetainHandle();
   *out_object_ptr = thread->guest_object();
   return X_STATUS_SUCCESS;
 }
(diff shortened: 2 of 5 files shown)


CookiePLMonster:
For the time being, this breaks Payday 2 (only tested Demo) in Checked builds:

From what I can tell, a thread gets created and instantly closed - and due `[Kernel] Do not retain handle in ExCreateThread` this causes thread's **handle** to be closed (thread **object** is still retained); in turn `xe::threading::set_current_thread_id(handle());` breaks because by the time thread starts executing, handles are dead.

CookiePLMonster:
Fix pushed. Retaining and releasing thread handles and not just thread objects allows them to continue gracefully and without leaking.

PR #1507, opened 2019-11-18
[imgui] Update to latest upstream master.

JoelLinn:
- Submodule is updated.
- All removed APIs are replaced.
- All APIs that have been deprecated are replaced.
- Biggest code changes are that ownership of the font atlas is now handled by imgui itself (xenia could do it as well but there is no benefit in that) and how the draw lists are processed (they are now processed once by the window class after all painting callbacks are returned because the old imgui callback mechanism is deprecated)

I tested this with the hid and ui demo, the profiler and guest debugger.
I didn't test trace viewers, please report if you find any regressions.

### DIFF
--- a/src/xenia/ui/imgui_drawer.h
+++ b/src/xenia/ui/imgui_drawer.h
@@ -18,6 +18,7 @@
 
 struct ImDrawData;
 struct ImFontAtlas;
+struct ImGuiContext;
 struct ImGuiIO;
 
 namespace xe {
@@ -56,7 +57,7 @@ class ImGuiDrawer : public WindowListener {
   Window* window_ = nullptr;
   GraphicsContext* graphics_context_ = nullptr;
 
-  std::vector<uint8_t> internal_state_;
+  ImGuiContext* internal_state_ = nullptr;
   std::unique_ptr<ImFontAtlas> font_atlas_;
   std::unique_ptr<ImmediateTexture> font_texture_;
 };
--- a/src/xenia/ui/imgui_drawer.h
+++ b/src/xenia/ui/imgui_drawer.h
@@ -17,7 +17,6 @@
 #include "xenia/ui/window_listener.h"
 
 struct ImDrawData;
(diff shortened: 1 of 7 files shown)


JoelLinn:
linter fixup...  3aa121c vs 9ebc4c8

gandalfthewhite19890404:
How cyrillic(and other non Latin) fonts are rendered? With outdated imgui as question marks (?).
Test-cases - any localized titles with changeable player's name as Diablo 3 (Vanilla and RoS), Fallout 3 etc.

JoelLinn:
Japanese is loaded in the SetupFonts() function. Adding a Cyrillic font would be a separate PR.

But guest (ingame) fonts aren‘t rendered by imgui. It could be that on the 360 there are system font files the game can not load (find) in the emulator.

gibbed:
I'm assuming they were referring to dialog prompts, which are provided by the kernel/system UI which would need native support for whatever language is being used. Not game rendering of text.

gibbed:
Is imgui master currently considered _stable_? If not, I would prefer updating to the latest stable release vs master.

JoelLinn:
As per their readme

> I occasionally tag Releases but it is generally safe and recommended to sync to master/latest. The library is fairly stable and regressions tend to be fixed fast when reported.

Did check the latest tag, it is two months old but has > 1000 changes to the relevant files.

gibbed:
That's fine then.

src/xenia/ui/imgui_drawer.h
```diff
@@ -16,8 +16,8 @@
 #include "xenia/ui/immediate_drawer.h"
 #include "xenia/ui/window_listener.h"
 
+struct ImGuiContext;
```
gibbed: Previous list was sorted alphabetically, this breaks it.

JoelLinn: sorted those forward declarations.

PR #1508, opened 2019-11-20
Remove unused icons

Margen67:


illusion98:
Reword your commit message to something like "Remove unused debugging icons"

PR #1509, opened 2019-11-20
[Base] Fix Clock

JoelLinn:
## Current state
There is a dedicated clock for each thread, updated independently. Upon thread save and restore these time values are preserved. 
I see no reason for this as time naturally progresses evenly for all threads (except the xbox360 is rotated very fast with one core as it's center and the power button near the speed of light which is hard to emulate anyways...). **This causes time to drift between different threads** and causes issues in many games that retrieve tick counts from more than one thread like RAGE engine titles (RDR, GTA). Additionally there is questionable **floating point arithmetic involved in translating the 64 bit tick count** between host and guest.

Both of these issues can cause guests to see time regress.

## Improved Clock implementation
The **Clock state now is unique** and dynamic fields are protected by a lock. Floating point arithmetics have been removed from the **perpetual tick calculations and replaced by rational number arithmetics**. Thread save / restore does not alter Clock states anymore. 

Additionally wo cvars where added to change Clock behavior from startup.
- `clock_no_scaling` removes all guest time scaling from equations, always reports host wall clock time to the guest and does not require locking, resulting in faster code. This would help guests querying ticks very often and across different threads simultaneously, as to them its just a simple ppc cpu instruction.
- `clock_source_raw` uses the cpu cycle counter as the timing source. Although it may not be that portable across CPUs its always higher resolution than the 360 tick counter. Windows' `QueryPerformanceCounter` is usually less precise than the 360. If for example a guest would try to measure the execution time of some code the resolution on most Windows system could not be enough.

If both cvars are set to true, the clock translation will be inlined into the JIT assembly for the PPC clock instruction `mftb`.

### Effects on games
Red Dead Redemption starts/stops audio fine and triggers missions/events normally.
Other games should be affected as well, please feel free to download this build and test those.

### DIFF
--- a/src/xenia/base/clock.cc
+++ b/src/xenia/base/clock.cc
@@ -20,13 +20,17 @@ DEFINE_bool(clock_no_scaling, false,
             "Disable scaling code. Time management and locking is bypassed. "
             "Guest system time is directly pulled from host.",
             "CPU");
+DEFINE_bool(clock_source_raw, false,
+            "Use the RDTSC instruction as the time source. "
+            "Host CPU must support invariant TSC. ",
+            "CPU");
 
 namespace xe {
 
 // Time scalar applied to all time operations.
 double guest_time_scalar_ = 1.0;
 // Tick frequency of guest.
-uint64_t guest_tick_frequency_ = Clock::host_tick_frequency();
+uint64_t guest_tick_frequency_ = Clock::host_tick_frequency_platform();
 // Base FILETIME of the guest system from app start.
 uint64_t guest_system_time_base_ = Clock::QueryHostSystemTime();
 // Combined time and frequency ratio between host and guest.
@@ -44,7 +48,7 @@ std::mutex tick_mutex_;
 void RecomputeGuestTickScalar() {
   // Create a rational number with numerator (first) and denominator (second)
   auto frac =
-      std::make_pair(guest_tick_frequency_, Clock::host_tick_frequency());
+      std::make_pair(guest_tick_frequency_, Clock::QueryHostTickFrequency());
   // Doing it this way ensures we don't mess up our frequency scaling and
   // precisely controls the precision the guest_time_scalar_ can have.
   if (guest_time_scalar_ > 1.0) {
@@ -102,6 +106,21 @@ inline uint64_t QueryGuestSystemTimeOffset() {
   reduce_fraction(numerator, denominator);
 
   return guest_tick_count * numerator / denominator;
+}
+
+uint64_t Clock::QueryHostTickFrequency() {
+  if (cvars::clock_source_raw) {
+    return host_tick_frequency_raw();
+  } else {
+    return host_tick_frequency_platform();
+  }
+}
+uint64_t Clock::QueryHostTickCount() {
+  if (cvars::clock_source_raw) {
+    return host_tick_count_raw();
+  } else {
+    return host_tick_count_platform();
+  }
 }
 
 double Clock::guest_time_scalar() { return guest_time_scalar_; }
--- a/src/xenia/base/clock.cc
+++ b/src/xenia/base/clock.cc
@@ -2,7 +2,7 @@
  ******************************************************************************
  * Xenia : Xbox 360 Emulator Research Project                                 *
  ******************************************************************************
- * Copyright 2015 Ben Vanik. All rights reserved.                             *
+ * Copyright 2019 Ben Vanik. All rights reserved.                             *
  * Released under the BSD license - see LICENSE in the root for more details. *
  ******************************************************************************
  */
@@ -10,9 +10,11 @@
 #include "xenia/base/clock.h"
 
 #include <algorithm>
-#include <climits>
+#include <limits>
+#include <mutex>
 
 #include "xenia/base/assert.h"
+#include "xenia/base/math.h"
 
 namespace xe {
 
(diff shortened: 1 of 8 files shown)


gibbed:
Looks fine other than breaking Travis, if you can sort that out (Linux issues) then it'll be mergeable.

JoelLinn:
Fixed linux builds.

altiereslima:
This fix Tropico 5. Thanks.

illusion98:
Fixed some sound trigger issues in GTA IV.

src/xenia/base/clock.cc
```diff
@@ -10,9 +10,20 @@
...
 #include <algorithm>
-#include <climits>
+#include <limits>
+#include <mutex>
 
 #include "xenia/base/assert.h"
+#include "xenia/base/math.h"
+
+DEFINE_bool(clock_no_scaling, false,
+            "Disable scaling code. Time managemend and locking is bypassed. "
```
Triang3l: Not very important, but "management".

src/xenia/base/clock.cc
```diff
@@ -68,43 +132,58 @@ void Clock::set_guest_system_time_base(uint64_t time_base) {
...
 }
 
 uint32_t Clock::ScaleGuestDurationMillis(uint32_t guest_ms) {
-  if (guest_ms == UINT_MAX) {
-    return UINT_MAX;
+  if (cvars::clock_no_scaling) {
+    return guest_ms;
+  }
+
+  uint64_t constexpr max = std::numeric_limits<uint32_t>::max();
```
Triang3l: It's `constexpr uint64_t` mostly according our clang-format, I think.

JoelLinn: Yes, in the morning that looks retarded.

MarioSonic2987:
Fixes missing sounds in MCLA , for example, when you press Start button in the title screen.

src/xenia/base/clock_win.cc
```diff
@@ -14,6 +14,36 @@
 namespace xe {
 
 uint64_t Clock::host_tick_frequency() {
+#if XE_ARCH_AMD64 && XE_COMPILER_MSVC
```
CookiePLMonster: Does this function still run multiple times or just once per session?

1. If it runs multiple times, this probably only needs to be checked once, perhaps via a lambda populating a static variable (so it's thread safe).
2. If it runs only once, path using `QueryPerformanceFrequency` doesn't need to have its frequency as a static variable. Additionally, that path is currently not thread safe, as concurrent access may result in concurrent writes/reads to this non-atomic variable, which is UB.

src/xenia/base/clock.cc
```diff
@@ -68,43 +132,58 @@ void Clock::set_guest_system_time_base(uint64_t time_base) {
...
+  return guest_system_time_base_ + guest_system_time_offset;
 }
 
-void Clock::SetGuestTickCount(uint64_t tick_count) {
-  last_host_tick_count_ = Clock::QueryHostTickCount();
-  guest_tick_count_ = tick_count;
+uint32_t Clock::QueryGuestUptimeMillis() {
+  return static_cast<uint32_t>(
+      std::min(QueryGuestSystemTimeOffset() / 10000,
+               static_cast<uint64_t>(std::numeric_limits<uint32_t>::max())));
```
CookiePLMonster: Cast to `uint64_t` can probably be removed by specializing `std::min<uint64_t>` here?

YeimiK:
Can confirm this fixes Castlevania LoS, as it goes ingme for the first 
time. Apparently the game had some clock based start routine, that wasn't 
working correctly. Good work : )

En 20 de noviembre de 2019 6:12:44 PM Joel Linn <notifications@github.com> 
escribió:

src/xenia/base/clock.cc
```diff
@@ -22,34 +33,87 @@ double guest_time_scalar_ = 1.0;
...
+    // Nothing to update, calculate on the fly
+    auto host_tick_count = Clock::QueryHostTickCount();
+    return host_tick_count * guest_tick_ratio_.first / guest_tick_ratio_.second;
+  }
+
+  std::unique_lock<std::mutex> lock(tick_mutex_, std::defer_lock);
+
+  if (lock.try_lock()) {
+    // Translate host tick count to guest tick count.
+    uint64_t host_tick_count = Clock::QueryHostTickCount();
```
CookiePLMonster: Does this need to be under a lock? The others do since they're accessing member variables, but this can likely be executed concurrently and discarded upon "losing the race".

Waifu-Stack:
Joel your a legend mate keep up the good work

src/xenia/base/clock_win.cc
JoelLinn: It runs at startup and every time time scaling is changed (+/-).
Just removed the `static` modifier.

JoelLinn:
- I pulled out the code for the raw clock from the platform dependent `clock_win.cc` since it not really is platform dependent.
- When both of the new cvars are set the JIT compiler now is aware of that and emits instructions to retrieve timing in place. Tick scaling values calculated in the Clock class are then baked into the code.

src/xenia/base/clock.cc
```diff
@@ -10,50 +10,136 @@
...
+    frac.first *= static_cast<uint64_t>(guest_time_scalar_ * 10.0);
+    frac.second *= 10;
+  } else {
+    frac.first *= 10;
+    frac.second *= static_cast<uint64_t>(10.0 / guest_time_scalar_);
+  }
+  // Keep this a rational calculation and reduce the fraction
+  reduce_fraction(frac);
+
+  std::unique_lock<std::mutex> lock(tick_mutex_);
```
CookiePLMonster:
```suggestion
  std::lock_guard lock(tick_mutex_);
```

`std::lock_guard<std::mutex>` if we don't have C++17 on all platforms.

src/xenia/base/clock.cc
```diff
@@ -10,50 +10,136 @@
...
+  if (cvars::clock_no_scaling) {
+    return;
+  }
+
   guest_time_scalar_ = scalar;
   RecomputeGuestTickScalar();
 }
 
+std::pair<uint64_t, uint64_t> Clock::guest_tick_ratio() {
+  std::unique_lock<std::mutex> lock(tick_mutex_);
```
CookiePLMonster:
```suggestion
  std::lock_guard lock(tick_mutex_);
```

src/xenia/base/clock_win.cc
```diff
@@ -13,15 +13,13 @@
 
 namespace xe {
 
-uint64_t Clock::host_tick_frequency() {
-  static LARGE_INTEGER frequency = {{0}};
-  if (!frequency.QuadPart) {
-    QueryPerformanceFrequency(&frequency);
-  }
+uint64_t Clock::host_tick_frequency_platform() {
+  LARGE_INTEGER frequency = {{0}};
```
CookiePLMonster:
```suggestion
  LARGE_INTEGER frequency;
```

Pretty sure it doesn't need to be initialized since `QueryPerformanceFrequency` annotates the argument as an output argument.

src/xenia/base/clock_win.cc
```diff
@@ -37,7 +35,7 @@ uint64_t Clock::QueryHostSystemTime() {
 }
 
 uint64_t Clock::QueryHostUptimeMillis() {
-  return QueryHostTickCount() / (host_tick_frequency() / 1000);
+  return host_tick_count_platform() / (host_tick_frequency_platform() / 1000);
```
CookiePLMonster:
```suggestion
  return (host_tick_count_platform() * 1000) / host_tick_frequency_platform();
```

As we discussed previously on Discord, I would consider arithmetic on frequency risky, as you cannot really make assumptions about its units. I'll feel much safer when it's done via multiplication instead.

CookiePLMonster: Same for any other instances of dividing frequency, whether they are touched by your PR or not.

src/xenia/base/math.h
```diff
@@ -59,6 +60,36 @@ T next_pow2(T value) {
   return value;
 }
 
+#if __cplusplus >= 201703L
```
CookiePLMonster: This is cool and all, but does it actually work? Last time I checked `__cplusplus` was not really meaningful, especially in Visual Studio.

src/xenia/base/math.h
```diff
@@ -59,6 +60,36 @@ T next_pow2(T value) {
...
+template <typename T>
+inline constexpr T greatest_common_divisor(T a, T b) {
+  return std::gcd(a, b);
+}
+#else
+template <typename T>
+constexpr T greatest_common_divisor(T a, T b) {
+  // Use the Euclid algorithm to calculate the greatest common divisor
+  while (b) {
+    T tmp = b;
```
CookiePLMonster:
```suggestion
    T tmp = std::exchange(b, a % b);
```

This will do nothing to codegen, but I find it much nicer looking than splitting such idioms in two lines like it's done currently. Up to you whether you want to go for it though, as it's just a styling change.

JoelLinn: Then
```
while(b)
{
  a = std::exchange(b, a % b);
}
```
I guess?

CookiePLMonster: Indeed! That will work, since `std::exchange` returns by value and not a reference.

src/xenia/base/math.h
JoelLinn: Two possible solutions:

1.  Use [`/Zc:__cplusplus`](https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2019) compiler flag for msvc.
2. Handle this in `platform.h` with something like:
```
#if XE_COMPILER_MSVC
#define XE_CPP_STD_VER _MSVC_LANG
#else
#define XE_CPP_STD_VER __cplusplus
#endif
```

See [this link](https://gist.github.com/ax3l/53db9fa8a4f4c21ecc5c4100c0d93c94) for an overview of compiler behavior.
[MSVC predefined macros](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019)

CookiePLMonster: I like the `/Zc` variation very much. However, with C++17 we finally have standard feature test macros: https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros
I wonder if they're respected by MSVC and other compilers though?

JoelLinn: Ok, so ditch both approaches, and use `__cpp_lib_gcd_lcm` macro.
This [macro is known](https://godbolt.org/z/vYpEfH) by Visual Studio 2017 version 15.9, gcc 7.3 and clang 5.0 or newer and is enabled when the c++17 standard is active.

CookiePLMonster: Excellent, good to know compilers have all adapted it.

src/xenia/base/clock.cc
CookiePLMonster: @JoelLinn You overlooked this one :)

JoelLinn: Thanks, they were in different commits.

JoelLinn:
Integrated those suggestions.
Hope this will do now.

jacob1218:
Fixes NASCAR 09 crashing on post-race results screen

legend800:
Fixes Trials Evo from freezing on load screen. Can get in-game now and plays fine, other than a lot of missing textures/shaders.

JoelLinn:
I changed `Clock::host_tick_frequency_raw()` to give a message box about incompatibility with the host CPU instead of just crashing. Some users reported to me that their AMD systems crash with `clock_source_raw` enabled. There is no easy way to determine the P0-state frequency outside of `ring0` and there is no intention on my side to try and fix this as this code path is only meant to debug problems if guests are behaving buggy.

illusion98:
Perfect Dark Zero Ingame for the first time with this PR

illusion98:
Raw Clock Source check confirmed working on AMD CPU

LuismaSP89:
This PR helped many games, but causes issues in others, for example:

Games booting now: Alice madness returns (Now can pass the press start screen)

Sniper Elite V2 no longer boots (Before this PR it was nearly-perfect)
Resident evil 5: Perfect in master, some graphical artifacts with this PR.
Bayonetta: Some crashes and devices lost after some gameplay. Near perfect in master.

illusion98:
> Sniper Elite V2 no longer boots (Before this PR it was nearly-perfect)

@LuismaSP89 working fine here

JoelLinn:
> Sniper Elite V2 no longer boots (Before this PR it was nearly-perfect)

Other people report no problems

> Resident evil 5: Perfect in master, some graphical artifacts with this PR.
> Bayonetta: Some crashes and devices lost after some gameplay. Near perfect in master.

I think graphical artifacts occur when the pipeline is not synced (according to △) . It happened for me on master more or less pronounced on the same build at different times. Device loss is a known issue which also can happen every minute or every 3 hours.

LuismaSP89:
@illusion98 Testing with latest canary build and Sniper Elite V2 GOTY edition. PAL version. 

Maybe yours are different version?

illusion98:
The regression is only on Canary and is not related to this Pull.

LuismaSP89:
Yes, it's on canary, and it's related with this PR because it started when this PR was merged in the canary.

Do not misunderstand me, I think this PR is glorious and helps many games, but breaks others, I only want to test and help the developer of this PR to merge it without regressions.

JoelLinn:
Try downloading the build of this PR and compare that with master.

Click on commits, on the green checkmark next to the last commit, 'details' next to appveyor.
On the new page on Configuration:Release, on Artifacts and download that zip file.
That build only differs from master by this PR.

LuismaSP89:
My apologies, sniper elite works as it should with the latest build of your PR.

Btw, playing resident evil 5 results in a "graphic device lost" and this doesn't happen in master. I don't know if is a coincidence, or if is related with this PR.

LuismaSP89:
Lost odyssey hangs while the "now loading" screen in a new game. This doesn't happened in master. I need to test too if is related to this PR.

legend800:
Don't post anything here and spam everyone with messages until you confirm the change is directly caused by this pr.

JoelLinn:
Rebased to latest master and changed copyrights to 2019.

PR #1510, opened 2019-11-26
[Kernel] Rewrote ObReferenceObjectByHandle

Gliniak:
From functional changes:
 - Implemented situation when we can receive different object type in handle and object_type_ptr. It returns X_STATUS_OBJECT_TYPE_MISMATCH; (Which is based on [windows documentation](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-obreferenceobjectbyhandle))

from other changes: 
 - Removed DRY code
 - overall code simplification

Thanks to @CookiePLMonster for help and patience.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
@@ -78,70 +78,35 @@ DECLARE_XBOXKRNL_EXPORT1(ObLookupThreadByThreadId, kNone, kImplemented);
 dword_result_t ObReferenceObjectByHandle(dword_t handle,
                                          dword_t object_type_ptr,
                                          lpdword_t out_object_ptr) {
-  X_STATUS result = X_STATUS_SUCCESS;
+  const static std::unordered_map<XObject::Type, uint32_t> obj_type_match = {
+      {XObject::kTypeEvent, 0xD00EBEEF},
+      {XObject::kTypeSemaphore, 0xD017BEEF},
+      {XObject::kTypeThread, 0xD01BBEEF}};
 
   auto object = kernel_state()->object_table()->LookupObject<XObject>(handle);
-  if (object) {
-    // TODO(benvanik): verify type with object_type_ptr
 
-    // TODO(benvanik): get native value, if supported.
-    uint32_t native_ptr;
-    switch (object_type_ptr) {
-      case 0x00000000: {  // whatever?
-        switch (object->type()) {
-          case XObject::kTypeEvent: {
-            assert(object->type() == XObject::kTypeEvent);
-            native_ptr = object->guest_object();
-            assert_not_zero(native_ptr);
-          } break;
-          case XObject::kTypeSemaphore: {
-            assert(object->type() == XObject::kTypeSemaphore);
-            native_ptr = object->guest_object();
-            assert_not_zero(native_ptr);
-          } break;
-          case XObject::kTypeThread: {
-            assert(object->type() == XObject::kTypeThread);
-            native_ptr = object->guest_object();
-            assert_not_zero(native_ptr);
-          } break;
-          default: {
-            assert_unhandled_case(object->type());
-            native_ptr = 0xDEADF00D;
-          } break;
-        }
-      } break;
-      case 0xD00EBEEF: {  // ExEventObjectType
-        assert(object->type() == XObject::kTypeEvent);
-        native_ptr = object->guest_object();
-        assert_not_zero(native_ptr);
-      } break;
-      case 0xD017BEEF: {  // ExSemaphoreObjectType
-        assert(object->type() == XObject::kTypeSemaphore);
-        native_ptr = object->guest_object();
-        assert_not_zero(native_ptr);
-      } break;
-      case 0xD01BBEEF: {  // ExThreadObjectType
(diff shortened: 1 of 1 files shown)


faha223:
I believe this causes Halo 3 to crash on startup. Can someone else confirm?

illusion98:
> I believe this causes Halo 3 to crash on startup. Can someone else confirm?

@faha223 working here

src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
```diff
@@ -78,70 +78,40 @@ DECLARE_XBOXKRNL_EXPORT1(ObLookupThreadByThreadId, kNone, kImplemented);
...
+  }
+
+  uint32_t native_ptr = object->guest_object();
+
+  auto obj_type = std::find_if(
+      std::begin(obj_type_match), std::end(obj_type_match),
+      [object](const auto& entry) { return entry.second == object->type(); });
+
+  if (obj_type != std::end(obj_type_match)) {
+    if (object_type_ptr) {
```
randprint: >    if (object_type_ptr) {
Could we just drop this null check since we would have already bailed at checking object for null above to include the object_type_ptr == 0x00000000 case?

CookiePLMonster: That's precisely the bug I mentioned above.

Gliniak:
Resolved merge conflict with: https://github.com/xenia-project/xenia/commit/728531eff7b3af41fa85685185334a0da1160d4f

src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
```diff
@@ -78,70 +78,40 @@ DECLARE_XBOXKRNL_EXPORT1(ObLookupThreadByThreadId, kNone, kImplemented);
...
-    object->RetainHandle();
-    if (out_object_ptr.guest_address()) {
-      *out_object_ptr = native_ptr;
+  if (!object) {
+    return X_STATUS_INVALID_HANDLE;
+  }
+
+  uint32_t native_ptr = object->guest_object();
+
+  auto obj_type = std::find_if(
```
DrChat: Maybe make `obj_type_match` an std::map instead to make it more obvious how this structure is intended to be used.

src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
```diff
@@ -78,70 +78,40 @@ DECLARE_XBOXKRNL_EXPORT1(ObLookupThreadByThreadId, kNone, kImplemented);
 dword_result_t ObReferenceObjectByHandle(dword_t handle,
                                          dword_t object_type_ptr,
                                          lpdword_t out_object_ptr) {
-  X_STATUS result = X_STATUS_SUCCESS;
+  const std::pair<XObject::Type, uint32_t> obj_type_match[] = {
```
DrChat: Could you make this a `const static` declaration? It isn't intended to be used by anyone else so `static` will make it invisible to other compilation units.

src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
```diff
@@ -78,70 +78,40 @@ DECLARE_XBOXKRNL_EXPORT1(ObLookupThreadByThreadId, kNone, kImplemented);
...
+  }
+
+  uint32_t native_ptr = object->guest_object();
+
+  auto obj_type = std::find_if(
+      std::begin(obj_type_match), std::end(obj_type_match),
+      [object](const auto& entry) { return entry.first == object->type(); });
+
+  if (obj_type != std::end(obj_type_match)) {
+    if (object_type_ptr) {
```
DrChat: We can probably just combine the if statement here into `if (object_type_ptr && (object_type_ptr != obj_type->second))`

Gliniak:
I adjusted 2/3 suggestions. I stayed with std::pair instead of std::map. 
That was suggestion from @CookiePLMonster to use it as more _proper_ solution. (Don't be mad bro 💃 )

src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
CookiePLMonster: I suggested against it because:
1) It has constant contents and size known at compile time
2) For such small samples binary search isn't faster than linear but has more code overhead

DrChat: AFAIK you can declare an `std::map` the same way. You're right about the search potentially being slower, though it likely does not matter here unless a game hammers on this function.
We can use `std::unordered_map` instead because order does not matter.

CookiePLMonster: Indeed it's possible - so I guess it's just a matter of preference.

src/xenia/kernel/xboxkrnl/xboxkrnl_ob.cc
DrChat: Er, looks like I mistakenly thought that this was declared _outside_ the function.
The latter part of my last statement is false, but making this static means it won't be re-initialized whenever the function is entered.

Additionally I believe `constexpr` may also achieve the same results.

PR #1514, opened 2019-12-01
[UI] Fix misc imgui things.

JoelLinn:
### Fixes these issues which mainly affect developers debugging:
- Because frame time is calculated based on `GetSystemTimeAsFileTime()` a time delta of zero is possible (Value is 100ns resolution but apparently not updated at that rate by the system).
image
- A new imgui frame is created before it is verified the context isn't lost.
image


JoelLinn:
Rebased onto latest master.
Made it compile against clock changes.

randprint:
image
nitpick: some overlays are missing the transparent background they used to have

PR #1517, opened 2019-12-03
[CPU] Change thunk layout and fix Linux thunk misalignment

Prism019:
The current guest-to-host and host-to-guest thunks do not save registers that are volatile on Linux. The current thunk layout also has the memory location for the xmm registers misaligned to 16 bytes, meaning the calls to vmovaps should throw a GP exception even on Windows (follow-up for anyone with Windows: is the struct not being properly packed? After this PR is merged, sizeof(Thunk) should be 0x100, or 256 in decimal, and before this PR gets merged, it should be 0x120, or 288 in decimal.)

This PR saves registers necessary to be saved on Linux as well as changes the layout of the Thunk struct to properly align the xmm register memory locations. The scratch and dummy elements removed from the Thunk struct have so-far been unused since the struct's introduction in 2016.

These changes have been tested on a few games on Windows to ensure emulation still works.


Prism019:
So I was right. The XEPACKEDSTRUCT isn't packing the struct correctly on Windows, and as such, didn't cause the GP exceptions.

Before this commit was merged, sizeof(Thunk) was 304 bytes, when it should have been 288. Maybe we should investigate this..

PR #1534, opened 2020-02-01
Add F2 Hotkey to open current commit ID URL

illusion98:
Allows the usage of F2 key to open current build commit in browser


illusion98:
Needs suggestions on module tags.

illusion98:
Addressed comment from Rick.

PR #1535, opened 2020-02-03
[CPU] Implemented instructions: rldclx, rldcrx

Gliniak:
These opcodes are barely used. I implemented them, by looking at different, but similar instructions and merged what I needed.

If required I might split this commit in two. (commit per instruction)

**Disclaimer:** I cannot provide tests for them as I cannot check them on console.
FYI: other opcodes from this "group" don't have tests

**Usage:**
Only known titles that uses them are _Telltale_ games to decompress data.

Impact:
It makes _Telltale Batman_ startable and probably also few other _Telltale_ titles

image

There is also Borat for your commitment
image

# Resources
[Power ISA v3.0B](https://openpowerfoundation.org/?resource_lib=power-isa-version-3-0)

### DIFF
--- a/src/xenia/cpu/ppc/ppc_emit_alu.cc
+++ b/src/xenia/cpu/ppc/ppc_emit_alu.cc
@@ -858,13 +858,53 @@ int InstrEmit_xoris(PPCHIRBuilder& f, const InstrData& i) {
 // Integer rotate (A-6)
 
 int InstrEmit_rldclx(PPCHIRBuilder& f, const InstrData& i) {
-  XEINSTRNOTIMPLEMENTED();
-  return 1;
+  // n <- rB[58:63]
+  // r <- ROTL[64](rS, n)
+  // b <- mb[5] || mb[0:4]
+  // m <- MASK(b, 63)
+  // rA <- r & m
+  Value* n = f.And(f.Truncate(f.LoadGPR(i.MDS.RB), INT8_TYPE),
+                   f.LoadConstantInt8(0x3F));
+
+  uint32_t mb = (i.MDS.MB5 << 5) | i.MDS.MB;
+  uint64_t m = XEMASK(mb, 63);
+  Value* v = f.LoadGPR(i.MDS.RT);
+
+  v = f.RotateLeft(v, n);
+  if (m != 0xFFFFFFFFFFFFFFFF) {
+    v = f.And(v, f.LoadConstantUint64(m));
+  }
+
+  f.StoreGPR(i.MDS.RA, v);
+  if (i.MDS.Rc) {
+    f.UpdateCR(0, v);
+  }
+  return 0;
 }
 
 int InstrEmit_rldcrx(PPCHIRBuilder& f, const InstrData& i) {
-  XEINSTRNOTIMPLEMENTED();
-  return 1;
+  // n <- rB[58:63]
+  // r <- ROTL[64](rS, n)
+  // b <- mb[5] || mb[0:4]
+  // m <- MASK(0, b)
+  // rA <- r & m
+  Value* n = f.And(f.Truncate(f.LoadGPR(i.MDS.RB), INT8_TYPE),
+                   f.LoadConstantInt8(0x3F));
+
+  uint32_t mb = (i.MDS.MB5 << 5) | i.MDS.MB;
+  uint64_t m = XEMASK(0, mb);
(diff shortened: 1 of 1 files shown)


DrChat:
IBM Reference:
chrome_ZxHADoJiTj
chrome_fLQEAbysD8

src/xenia/cpu/ppc/ppc_emit_alu.cc
```diff
@@ -858,13 +858,57 @@ int InstrEmit_xoris(PPCHIRBuilder& f, const InstrData& i) {
...
 int InstrEmit_rldcrx(PPCHIRBuilder& f, const InstrData& i) {
-  XEINSTRNOTIMPLEMENTED();
-  return 1;
+  // n <- rB[58:63]
+  // r <- ROTL[64](rS, n)
+  // b <- mb[5] || mb[0:4]
+  // m <- MASK(0, b)
+  // rA <- r & m
+  Value* n =
+      f.And(f.Truncate(f.LoadGPR(i.X.RB), INT8_TYPE), f.LoadConstantInt8(0x7F));
```
DrChat: I think the mask for `[58:63]` would be `0x3F` (63-58+1 = 6 bits masked)
Additionally, you'll want to stick with `i.MDS.RB` for the instruction form - using `i.MDS` across the entire implementation of this instruction.

The IBM documentation I pasted earlier shows the instruction form

src/xenia/cpu/ppc/ppc_emit_alu.cc
```diff
@@ -858,13 +858,57 @@ int InstrEmit_xoris(PPCHIRBuilder& f, const InstrData& i) {
...
 int InstrEmit_rldclx(PPCHIRBuilder& f, const InstrData& i) {
-  XEINSTRNOTIMPLEMENTED();
-  return 1;
+  // n <- rB[58:63]
+  // r <- ROTL[64](rS, n)
+  // b <- mb[5] || mb[0:4]
+  // m <- MASK(b, 63)
+  // rA <- r & m
+  Value* n =
+      f.And(f.Truncate(f.LoadGPR(i.X.RB), INT8_TYPE), f.LoadConstantInt8(0x7F));
```
DrChat: Ditto with mask (`0x3F`) and `i.MDS`

src/xenia/cpu/ppc/ppc_emit_alu.cc
```diff
@@ -858,13 +858,57 @@ int InstrEmit_xoris(PPCHIRBuilder& f, const InstrData& i) {
...
-  return 1;
+  // n <- rB[58:63]
+  // r <- ROTL[64](rS, n)
+  // b <- mb[5] || mb[0:4]
+  // m <- MASK(0, b)
+  // rA <- r & m
+  Value* n =
+      f.And(f.Truncate(f.LoadGPR(i.X.RB), INT8_TYPE), f.LoadConstantInt8(0x7F));
+
+  uint32_t mb = (i.MD.MB5 << 5) | i.MD.MB;
```
DrChat: Ditto (`i.MDS`)

src/xenia/cpu/ppc/ppc_emit_alu.cc
```diff
@@ -858,13 +858,57 @@ int InstrEmit_xoris(PPCHIRBuilder& f, const InstrData& i) {
...
+  Value* v = f.LoadGPR(i.MD.RT);
+
+  if (n) {
+    v = f.RotateLeft(v, n);
+  }
+  if (m != 0xFFFFFFFFFFFFFFFF) {
+    v = f.And(v, f.LoadConstantUint64(m));
+  }
+
+  f.StoreGPR(i.MD.RA, v);
```
DrChat: Ditto (`i.MDS`)

Gliniak:
Thanks for review I appreciate that. (I will resolve that issues asap)

Nevertheless I have a question about that mask.
I used it from sldx, which based on documentation uses exact same mask. Is it some kind of simplification for that instruction? (few others also have same mask applied)

DrChat:
If those instructions have the same mask, they might actually be wrong as well. Ping @gibbed.
IBM docs typically use an inclusive range for the mask.

Gliniak:
I updated that commit. Indeed mask was wrong (Thanks for pointing that out). 

Other instructions have proper mask (They use 0x7F instead of 0x3F for further computation).

PR #1536, opened 2020-02-06
[XAM] - correct storage device ID fixes saving and storage detection

randprint:
Correct storage device ID fixes saving and storage detection in many games

most of the games this can be evidenced in are Unreal Engine titles or games that otherwise do not work in master yet but it can be shown in Saints Row 2 that does work in the master branch
this ID was found from debugging games where storage device detection wasn't working and reverse engineering of Unreal Engine 3


randprint:
is closing and opening this the only way to restart travis?

jgoyvaerts:
You can also do a force push on the branch (for example changing the commit message slightly) to trigger a new build normally

PR #1538, opened 2020-02-07
[XAM] Correct storage device ID fixes saving and storage detection

randprint:
Correct storage device ID fixes saving and storage detection in many games

most of the games this can be evidenced in are Unreal Engine titles or games that otherwise do not work in master yet but it can be shown in Saints Row 2 that does work in the master branch
this ID was found from debugging games where storage device detection wasn't working and reverse engineering of Unreal Engine 3

### DIFF
--- a/src/xenia/kernel/xam/xam_content.cc
+++ b/src/xenia/kernel/xam/xam_content.cc
@@ -43,11 +43,14 @@ struct DeviceInfo {
 // they incorrectly only look at the lower 32-bits of free_bytes,
 // when it is a 64-bit value. Which means any size above ~4GB
 // will not be recognized properly.
+//
+// NOTE(randprint): you can use 120 GB and 42 GB 'fullness'
+// with the proper deviceID feel free to change at your discression
 #define ONE_GB (1024ull * 1024ull * 1024ull)
 static const DeviceInfo dummy_device_info_ = {
-    0xF00D0000,    1,
-    4ull * ONE_GB,  // 4GB
-    3ull * ONE_GB,  // 3GB, so it looks a little used.
+    0x00000001,    1,  // found from debugging / reversing UE3 engine titles
+    4ull * ONE_GB,     // 4GB
+    3ull * ONE_GB,     // 3GB, so it looks a little used.
     u"Dummy HDD",
 };
 #undef ONE_GB
@@ -72,7 +75,7 @@ DECLARE_XAM_EXPORT2(XamContentGetLicenseMask, kContent, kStub, kHighFrequency);
 dword_result_t XamContentGetDeviceName(dword_t device_id,
                                        lpu16string_t name_buffer,
                                        dword_t name_capacity) {
-  if ((device_id & 0xFFFF0000) != dummy_device_info_.device_id) {
+  if ((device_id & 0x0000000F) != dummy_device_info_.device_id) {
     return X_ERROR_DEVICE_NOT_CONNECTED;
   }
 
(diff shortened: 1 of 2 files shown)


Gliniak:
Promised info (I will edit this commit with more info soon)

Some games have "hardcoded" values (Not checked in code yet) for _device_id_ in _**XamContentGetDeviceData**_. 
Games check (with or without this change) 0x7E times in loop from _device_id_ 0x01 to 0x7F.
(even if device with certain id exist loop always goes to 0x7F!)

Due to that hardcoded value _**XamContentGetDeviceData**_ cannot meet condition and returns **_X_ERROR_DEVICE_NOT_CONNECTED_**

full log with that loop. It's from my local branch, but I can assure it's fine
[DeviceCheckLoop.log](https://github.com/xenia-project/xenia/files/4239281/DeviceCheckLoop.log)

< ToDo IDA code fragments >

Gliniak:
I did some research (based on GoW2).

I found function that is responsible for seeking HDD. However instead of receiving 0xFOOD0000 as param_2 (aka. device_id) it receives 0x01. (I guess it might be hardcoded?)

I'm into making changes. We need to remove that mask and just compare id if it matches provided id by game

image

src/xenia/kernel/xam/xam_content.cc
```diff
@@ -72,7 +75,7 @@ DECLARE_XAM_EXPORT2(XamContentGetLicenseMask, kContent, kStub, kHighFrequency);
 dword_result_t XamContentGetDeviceName(dword_t device_id,
                                        lpwstring_t name_buffer,
                                        dword_t name_capacity) {
-  if ((device_id & 0xFFFF0000) != dummy_device_info_.device_id) {
+  if ((device_id & 0x0000000F) != dummy_device_info_.device_id) {
```
Gliniak: Remove mask check completely

src/xenia/kernel/xam/xam_content.cc
```diff
@@ -89,7 +92,7 @@ DECLARE_XAM_EXPORT1(XamContentGetDeviceName, kContent, kImplemented);
 
 dword_result_t XamContentGetDeviceState(dword_t device_id,
                                         lpunknown_t overlapped_ptr) {
-  if ((device_id & 0xFFFF0000) != dummy_device_info_.device_id) {
+  if ((device_id & 0x0000000F) != dummy_device_info_.device_id) {
```
Gliniak: Same as above

src/xenia/kernel/xam/xam_content.cc
```diff
@@ -121,7 +124,7 @@ static_assert_size(X_CONTENT_DEVICE_DATA, 0x50);
 
 dword_result_t XamContentGetDeviceData(
     dword_t device_id, pointer_t<X_CONTENT_DEVICE_DATA> device_data) {
-  if ((device_id & 0xFFFF0000) != dummy_device_info_.device_id) {
+  if ((device_id & 0x0000000F) != dummy_device_info_.device_id) {
```
Gliniak: Same as above

src/xenia/kernel/xam/xam_content.cc
```diff
@@ -160,7 +163,7 @@ dword_result_t XamContentCreateEnumerator(dword_t user_index, dword_t device_id,
                                           lpdword_t buffer_size_ptr,
                                           lpdword_t handle_out) {
   assert_not_null(handle_out);
-  if ((device_id && (device_id & 0xFFFF0000) != dummy_device_info_.device_id) ||
+  if ((device_id && (device_id & 0x0000000F) != dummy_device_info_.device_id) ||
```
Gliniak: Same as above

PR #1539, opened 2020-02-08
[CPU] Increased maximum function count

Gliniak:
Few games have function count higher than 50k.
Ghidra and IDA reports the same number of functions.

Confirmed titles: _Watch Dogs_, _Far Cry 3_

Benefits: It allows FC3 go a little bit further, but still crashes.

Screen from IDA (FC3): https://media.discordapp.net/attachments/440280035056943104/675094290426494984/unknown.png


gibbed:
Self-note: this will add about ~1MB allocations total from places that use this constant.

PR #1545, opened 2020-02-12
Replace binary_function by plain function.

JoelLinn:
- Deprecated in C++11
- Removed in C++17


JoelLinn:
O holy divinity of clang. Please show mercy in the face of my peasant code formatting.

JoelLinn:
Didnt explicitly search all files but compiled the whole solution with c++17 and that was the only binary function it complained about, so I would go so far and make that assumption - **yes**.

PR #1550, opened 2020-02-19
[XAM/Network] Implemented NetDll___WSAFDIsSet

Gliniak:
### Overview

This is based on description from [this site](https://www.intervalzero.com/library/RTX/WebHelp/Content/PROJECTS/SDK%20Reference/WinsockRef/WSAFDIsSet.htm) as Microsoft don't provide information about return value.

~~I marked it as _Sketchy_. Documentation says that it returns integer when socket was found in _fd_set_, but it does not specifies what value (is it 1 or provided socket).~~

Update: Thanks RedMadKnight for providing info about returned value

### Benefits:
_Renegade Ops_ goes ingame
_Test Drive Unlimited 2_ might go further


RedMadKnight:
In [winehq](https://www.winehq.org/pipermail/wine-patches/2015-January/136359.html) this func is defined as 1/0 (found/not found)

PR #1562, opened 2020-02-24
Gzips log file on application exit

jpb18:
This PR adds the zlib library to the third_party folder.

It also adds to the logging.cc and logging.h the code required to deflate (compress) the .log file into a .log.gz file. Also added a macro to logging.h (ZIPLOG) to invoke the xe::ZipCurrentFile() function easily. 

That macro is now invoked on the OnClose method on the window_win.cc file. 

There's also an unvalidated attempt to trigger the zipping function on an unhandled exception by using the SetUnhandledExceptionFilter from windows.h. Future work would include adapting that part for usage on Linux.

The file generated on one of the tests was roughly 60 times smaller than the original one.


Razzile:
Nice. Could you add a flag to enable/disable log compression?

janisozaur:
May I suggest using LZ4 instead? (I think) I understand the reason for this PR: the logs are large and thus problematic to handle. But this could cause an extended shutdown time and in case there are no events being processed, the application might get marked as not responding. The reason for suggesting LZ4 instead is how fast it is, at price of slightly worse compression ratio: https://github.com/lz4/lz4#benchmarks

Couple of notes to consider otherwise:
* GitHub will accept some compressed files as attachments: https://help.github.com/en/github/managing-your-work-on-github/file-attachments-on-issues-and-pull-requests
  - `.gz` under 25MB is accepted, but `.lz4` is not. It is just the filename though, there's no inspection of content.
* Maybe it would be worthwhile to implement the compression as a filter over the logging to file? You could submit chunks to the compression library, possibly also doing a flush (sync) every so often to spread the compression cost over time.
* There are `zcat`, `lz4cat`, etc. utils that come useful when having compressed logs.

jpb18:
> May I suggest using LZ4 instead? (I think) I understand the reason for this PR: the logs are large and thus problematic to handle. But this could cause an extended shutdown time and in case there are no events being processed, the application might get marked as not responding. The reason for suggesting LZ4 instead is how fast it is, at price of slightly worse compression ratio: https://github.com/lz4/lz4#benchmarks

Hmmm I haven't noticed much difference when closing the application. However, the biggest test I had was during a 3 minute gameplay of a demo.

> 
> Couple of notes to consider otherwise:
> 
> * GitHub will accept some compressed files as attachments: https://help.github.com/en/github/managing-your-work-on-github/file-attachments-on-issues-and-pull-requests
>   
>   * `.gz` under 25MB is accepted, but `.lz4` is not. It is just the filename though, there's no inspection of content.

Besides having more familiarity with .gz that was one of the reasons why I picked it for this.

> * Maybe it would be worthwhile to implement the compression as a filter over the logging to file? You could submit chunks to the compression library, possibly also doing a flush (sync) every so often to spread the compression cost over time.

Probably as a custom implementation of an output stream right? You'd write it into the file and all would be handled in the background. Possíbly a concurrent queue that would be emptied once it got to 80% of its capacity?

> * There are `zcat`, `lz4cat`, etc. utils that come useful when having compressed logs.

I did not knew about them. Gotta check them out.

jpb18:
> Nice. Could you add a flag to enable/disable log compression?

Good idea!

PR #1565, opened 2020-02-25
[x64] NAN up vmsum3fp128/vmsum4fp128.

gibbed:
It seems vmsum3fp128/vmsum4fp128 need to output a NAN on overflow.

Do so. Probably inefficiently.

(Also, [theme song for this PR](https://youtu.be/Hci2kZ8DlJM).)


gibbed:
Oh. I should mention this fixes collisions and crashes in Forza Horizon 2. Probably affects other games too.

PR #1570, opened 2020-03-01
[CI] Add GitHub Actions

Margen67:
Example run: https://github.com/Margen67/xenia/actions/runs/106069905

This PR adds automatic Windows and Linux builds, and Linting via GitHub Actions, and adds a badge to README.md.

### Pros:
  * Builds multiple configurations at a time. Therefore, the Debug configuration can be added without additional building time.

### Cons:
  * Artifacts:
      * Doesn't have a direct link to latest artifacts.
        This is possible with releases, but is messy.
      * Can't use ~~wildcards~~/regex, or upload files without zipping them.
        As a consequence, checked builds cannot be passworded without being double-zipped.
        This is also why multiple robocopy commands have to be done instead of one 7z command.
  * Can't delete builds/runs.

---

tl;dr Add this because AppVeyor is slow.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,115 @@
+name: CI
+
+on:
+  push:
+    branches-ignore: gh-pages
+    paths-ignore:
+      - '.github/*'
+      - '.github/ISSUE_TEMPLATE/**'
+      - 'docs/**'
+      - '*.md'
+      - '*.yml'
+      - 'LICENSE'
+  pull_request:
+    branches-ignore: gh-pages
+    paths-ignore:
+      - '.github/*'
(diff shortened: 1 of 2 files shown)


DrChat:
Nice work :)
It might be worth investigating if we should start releasing builds of Xenia through the releases page. We could tag releases or just release every time a new commit is pushed to the master branch.

Margen67:
@DrChat That's what Canary does, but it leads to a mess of tags: https://github.com/xenia-canary/xenia-canary/releases

q4a:
Hi all. I like idea of using GitHub Actions. Is there any problem with merging this PR?

q4a:
Nobody will add GitHub Actions?

PR #1571, opened 2020-03-04
Fix broken lock guard

NewProggie:
This commit fixes a bug where the lock guard (for concurrently accessing
the same scope from different threads) had basically no effect, due to
being bound to a temporary only.


Prism019:
good catch!

PR #1573, opened 2020-03-06
[xenia-build] Raise the minimum python version to 3.6

Prism019:
Xenia-build uses the "encoding" parameter of the subprocess module. This was introduced in Python 3.6.


Margen67:
Maybe a note should also be added about other installations of Python cluttering `PATH`?

Prism019:
I don't think that's our problem though? If someone has multiple installed versions of python, then I'm pretty sure they know what they're doing.

Margen67:
Superseded by #1581

PR #1577, opened 2020-03-22
[Kernel/XAM] XNotifyGetNext: Check for nullptr for param_ptr

Gliniak:
Based on dissasembled **_Ghost Recon AW2 Demo_**
I attached fragment from IDA to have confirmation that such situation might occur.

IDA proof: 
image


Triang3l:
Thanks for the fix, but please next time write the explanation of the change not only in the GitHub issue, but in code comments as well, to prevent regressions in case such change gets lost in a major rework of the implementation later, and contributors will know that this is the required behavior rather than an additional safety measure added by us. I added a comment explaining this in 61bcd467a6b63ffeaafb1894384e70a024f24554.

PR #1582, opened 2020-04-01
Make starting development on Linux easier.

JoelLinn:
- `xb devenv` command on Linux to generate CodeLite workspace and start it.
- Replace premake with upstream master revision, needed to fix CodeLite related bug. (Only change to our fork is making premake build on PowerPC - do we still need this?)
- Update building documentation.

With these changes getting an IDE and starting debugging is as easy as it is on Windows (At least on current *ubuntus). CodeLite may be inferior to CLion or others but automatic project file generation is key here.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -10,7 +10,7 @@ drivers.
 
 * Windows 7 or later
 * [Visual Studio 2019 or Visual Studio 2017](https://www.visualstudio.com/downloads/)
-* [Python 3.4+](https://www.python.org/downloads/)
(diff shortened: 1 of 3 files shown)


gibbed:
PPC support is for building the native PPC test runner I think?

JoelLinn:
yes, that makes sense I guess.
Would be nice to merge upstream premake into our fork, then I could rebase this.

gibbed:
Can you open an issue on _this repo_ to remind me to rebase it?

PR #1589, opened 2020-04-09
Use std::filesystem for simple queries and operations

bwrsandman:
With C++17 filesystem implemented, some of the `xe::filesystem` functions have become obsolete platform implementations of what the standard library already provides.

This PR removes `xe::filesystem::IsFolder`,  `xe::filesystem::PathExists`,  `xe::filesystem::CreateFolder`,  `xe::filesystem::DeleteFolder` and  `xe::filesystem::DeleteFile` and replaces them with their `std::filesystem` counterparts in the place where they are called.

This has the added bonus of implementing those functionalities in Posix as well.

### DIFF
--- a/src/xenia/kernel/xam/content_manager.cc
+++ b/src/xenia/kernel/xam/content_manager.cc
@@ -243,8 +243,7 @@ X_RESULT ContentManager::DeleteContent(const XCONTENT_DATA& data) {
   auto global_lock = global_critical_region_.Acquire();
 
   auto package_path = ResolvePackagePath(data);
-  if (std::filesystem::exists(package_path)) {
-    xe::filesystem::DeleteFolder(package_path);
+  if (std::filesystem::remove_all(package_path) > 0) {
     return X_ERROR_SUCCESS;
   } else {
     return X_ERROR_FILE_NOT_FOUND;
--- a/src/xenia/kernel/xam/content_manager.cc
+++ b/src/xenia/kernel/xam/content_manager.cc
@@ -153,7 +153,7 @@ X_RESULT ContentManager::CreateContent(const std::string_view root_name,
     return X_ERROR_ALREADY_EXISTS;
   }
 
-  if (!xe::filesystem::CreateFolder(package_path)) {
+  if (!std::filesystem::create_directories(package_path)) {
     return X_ERROR_ACCESS_DENIED;
   }
 
@@ -227,7 +227,7 @@ X_RESULT ContentManager::SetContentThumbnail(const XCONTENT_DATA& data,
                                              std::vector<uint8_t> buffer) {
   auto global_lock = global_critical_region_.Acquire();
   auto package_path = ResolvePackagePath(data);
-  xe::filesystem::CreateFolder(package_path);
+  std::filesystem::create_directories(package_path);
   if (std::filesystem::exists(package_path)) {
     auto thumb_path = package_path / kThumbnailFileName;
     auto file = xe::filesystem::OpenFile(thumb_path, "wb");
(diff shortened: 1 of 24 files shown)


src/xenia/kernel/xam/content_manager.cc
```diff
@@ -243,8 +243,8 @@ X_RESULT ContentManager::DeleteContent(const XCONTENT_DATA& data) {
   auto global_lock = global_critical_region_.Acquire();
 
   auto package_path = ResolvePackagePath(data);
-  if (xe::filesystem::PathExists(package_path)) {
-    xe::filesystem::DeleteFolder(package_path);
+  if (std::filesystem::exists(package_path)) {
+    std::filesystem::remove_all(package_path);
     return X_ERROR_SUCCESS;
   } else {
     return X_ERROR_FILE_NOT_FOUND;
```
bwrsandman, line 250: This could be replaced with the following:
```c++
  if (std::filesystem::remove_all(package_path) > 0) {
    return X_ERROR_SUCCESS;
  } else {
    return X_ERROR_FILE_NOT_FOUND;
  }
```
From the documentation, `std::filesystem::remove_all`:
> Returns the number of files and directories that were deleted (which may be zero if p did not exist to begin with).

This would cover the case of the directory not existing, but there are corner cases of the OS not removing these paths for whatever reason, in which case:
> The overload that does not take a `std::error_code&` parameter throws filesystem_error on underlying OS API errors, constructed with p as the first path argument and the OS error code as the error code argument.

We may not want any exceptions thrown here, in which case, maybe we should use the version which uses `std::error_code&` and interpret the OS API error code and return that.

src/xenia/kernel/xam/content_manager.cc
```diff
@@ -243,8 +243,7 @@ X_RESULT ContentManager::DeleteContent(const XCONTENT_DATA& data) {
   auto global_lock = global_critical_region_.Acquire();
 
   auto package_path = ResolvePackagePath(data);
-  if (xe::filesystem::PathExists(package_path)) {
-    xe::filesystem::DeleteFolder(package_path);
+  if (std::filesystem::remove_all(package_path) > 0) {
     return X_ERROR_SUCCESS;
   } else {
     return X_ERROR_FILE_NOT_FOUND;
```
bwrsandman, line 249: This could be replaced with the following:
```c++
  if (std::filesystem::remove_all(package_path) > 0) {
    return X_ERROR_SUCCESS;
  } else {
    return X_ERROR_FILE_NOT_FOUND;
  }
```
From the documentation, `std::filesystem::remove_all`:
> Returns the number of files and directories that were deleted (which may be zero if p did not exist to begin with).

This would cover the case of the directory not existing, but there are corner cases of the OS not removing these paths for whatever reason, in which case:
> The overload that does not take a `std::error_code&` parameter throws filesystem_error on underlying OS API errors, constructed with p as the first path argument and the OS error code as the error code argument.

We may not want any exceptions thrown here, in which case, maybe we should use the version which uses `std::error_code&` and interpret the OS API error code and return that.

On the other hand, these would be issues already present in Xenia without this change.

src/xenia/ui/imgui_drawer.cc
```diff
@@ -135,14 +134,15 @@ void ImGuiDrawer::SetupFont() {
...
-  if (xe::filesystem::PathExists(jp_font_path)) {
+  std::filesystem::path jp_font_path = "C:\\Windows\\Fonts\\msgothic.ttc";
+  if (std::filesystem::exists(jp_font_path)) {
     ImFontConfig jp_font_config;
     jp_font_config.MergeMode = true;
     jp_font_config.OversampleH = jp_font_config.OversampleV = 1;
     jp_font_config.PixelSnapH = true;
     jp_font_config.FontNo = 0;
-    io.Fonts->AddFontFromFileTTF(jp_font_path, 12.0f, &jp_font_config,
+    io.Fonts->AddFontFromFileTTF(jp_font_path.u8string().c_str(), 12.0f,
```
gibbed, line 144: Since we can't control the path in this situation, I think it would be better to leave it as a `const char*` to avoid the dance with a `std::filesystem::path` object.

bwrsandman: Ok. That's done.

PR #1606, opened 2020-04-18
[ImGui] Fix drawing lines.

JoelLinn:
ImGui does not orientate triangles.
Culling must not be applied in immediate drawer.

Added ImGui style colors for things that weren't rendered before.

### DIFF
--- a/src/xenia/ui/vulkan/vulkan_immediate_drawer.cc
+++ b/src/xenia/ui/vulkan/vulkan_immediate_drawer.cc
@@ -607,8 +607,8 @@ VkResult VulkanImmediateDrawer::Initialize() {
   rasterization_info.depthClampEnable = VK_FALSE;
   rasterization_info.rasterizerDiscardEnable = VK_FALSE;
   rasterization_info.polygonMode = VK_POLYGON_MODE_FILL;
-  rasterization_info.cullMode = VK_CULL_MODE_BACK_BIT;
-  rasterization_info.frontFace = VK_FRONT_FACE_CLOCKWISE;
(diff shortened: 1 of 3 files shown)


src/xenia/ui/vulkan/vulkan_immediate_drawer.cc
```diff
@@ -607,8 +607,8 @@ VkResult VulkanImmediateDrawer::Initialize() {
   rasterization_info.depthClampEnable = VK_FALSE;
   rasterization_info.rasterizerDiscardEnable = VK_FALSE;
   rasterization_info.polygonMode = VK_POLYGON_MODE_FILL;
-  rasterization_info.cullMode = VK_CULL_MODE_BACK_BIT;
-  rasterization_info.frontFace = VK_FRONT_FACE_CLOCKWISE;
+  rasterization_info.cullMode = VK_CULL_MODE_NONE;
+  rasterization_info.frontFace = VK_FRONT_FACE_COUNTER_CLOCKWISE;
```
Triang3l, line 611: This is contradicting the D3D version now (though AFAIK Vulkan has Y flipped as opposed to D3D), however it is better than clockwise of course because the CCW enum value is 0. But since we don't care, we should probably use the value from some official example if there are any that contain this.

JoelLinn: https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_vulkan.cpp#L719-L720
https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_dx12.cpp#L562-L563
For what ever reason they did that.

PR #1607, opened 2020-04-20
Rooted path fixes.

gibbed:
[Base] Fix canonicalize path for rooted paths.
[Base] Join paths better, so the first iteration isn't a join of an empty string with the first path part.
[Base] Add optional argument to allow empty parts in `utf8::split`.
[Base] Make utf8 arguments consistent. Fix `sep` to `separator`.
[VFS] Allow specifying root entry to open from with `OpenFile`.
[Kernel] `NtCreateFile` now opens from root entry when available instead of needlessly building a full path and resolving from that.
[VFS] Reduce code duplication by adding `Entry::ResolvePath`.
[VFS] Remove `ResolveBasePath` to avoid multiple calls to `find_base_guest_path`.

### DIFF
--- a/src/xenia/base/utf8.cc
+++ b/src/xenia/base/utf8.cc
@@ -485,7 +485,7 @@ std::vector<std::string_view> split_path(const std::string_view path) {
 }
 
 std::string join_paths(const std::string_view left_path,
-                       const std::string_view right_path, char32_t sep) {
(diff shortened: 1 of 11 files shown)


gibbed:
I should add that this fixes recent regression in games like Daytona USA where they weren't booting at all.

PR #1610, opened 2020-04-21
SDL GetKeystroke() implementation

JoelLinn:
Implements `GetKeystroke()` API. Button states flow as following:
SDL Event -> internally accumulated controller state (latest) -> button states at last  `GetKeystroke()` call <->  `GetKeystroke()` Keystroke

No internal queue is necessary. Biggest changesets are in first and last commit, rest is cream topping.
Please test with multiple controllers, I only have one atm.


gandalfthewhite19890404:
Please, test, with Diablo 3 and Diablo 3 Reaper of souls.

JoelLinn:
> Please, test, with Diablo 3 and Diablo 3 Reaper of souls.

If you have that game could you please try it yourself, by downloading the build from appveyor?
(Green checkmark, appveyor, release, artifacts) 
Set the hid option to 'sdl'.

The only game I have right now which uses this API is the Demo of NeoGeo Battle Coliseum which worked like I would expect (up until it crashes due to another bug)

JoelLinn:
Latest commit also fixes the bug where having any sdl subsystem enabled (apu or hid) would result in overall 2 fps less ("58 FPS bug").

PR #1612, opened 2020-04-25
[Kernel/Rtl] RtlTimeToTimeFields Check for highest defined time_t value.

Gliniak:
### **Description:**

Based on [this topic](https://stackoverflow.com/a/48045971) and this [MS page](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-s-localtime32-s-localtime64-s?view=vs-2019) it seems that time_t cannot contain value higher than os dependent value. For windows it is ``0x793406FFF``.

Every time related function that requires time_t value which is higher than that value will return error and null.

I tested it and in fact adding one day to that predefined value causes gmtime to return null

### **Affected games:**
- Dark souls

Interesting info: Output X_TIME_FIELDS data from 0 should be: `0641 0001 0001 0000 0000 0000 0000 0001`

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
@@ -29,6 +29,8 @@
 #define timegm _mkgmtime
 #endif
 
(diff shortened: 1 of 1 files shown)


Gliniak:
Closed due to: #1692

PR #1622, opened 2020-06-06
[Base] Wrap to_utf functions with try catch block

Gliniak:
Right now some games due to poor or even garbage thread naming (looking at shadowrun) likes to crash on utf conversion.

I wrapped that function around try-catch block to prevent such situation.

However I'm not sure if returning empty string is good idea and what about broken paths?

Also I was thinking about adding some log message just in case

### DIFF
--- a/src/xenia/base/string.cc
+++ b/src/xenia/base/string.cc
@@ -8,6 +8,7 @@
  */
 
 #include "xenia/base/string.h"
+#include "xenia/base/logging.h"
 
 #include <algorithm>
 #include <locale>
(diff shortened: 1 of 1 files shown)


JoelLinn:
You are right we need to handle this.

We should probably use a best effort approach in the exception handler. The library supports replacing invalid characters with placeholders. Unfortunately it seems such API is only available for utf8 strings.
```c++
// catch
std::string res;
utf8::replace_invalid(source.begin(), source.end(), std::back_inserter(res));
return res;
```

In both cases we should log a Warning. Both exceptions provide the char value over which it stumbled, could also log that as hex.

Also could you provide a sample string (as hex) which it fails to convert?

gibbed:
This is generally a case where guest strings are being treated incorrectly as if they were UTF8 to begin with, so yes, filtering is the better approach to begin with.

The Shadowrun bug is due to UAF on the thread name string so garbage ends up being passed to the kernel.

Gliniak:
Thanks for the suggestion of using utf8::replace_invalid (I didn't know it existed 🐑 )

I will replace it with it as it seems to be ok and on the other one I will add warning message.

About char on which it crashes (that 0xa6):
image

Gliniak:
This PR should be resolved by: https://github.com/xenia-project/xenia/commit/8749ca22609b6dad852a9e2b57d81d956bbdd4a3

PR #1629, opened 2020-07-12
[Kernel] Reject IO for paths with invalid characters.

gibbed:
[Kernel] Reject IO for paths with invalid characters.


gibbed:
@Prism019 Can you confirm if this fixes FIFA being a shit with that invalid path?

PR #1631, opened 2020-07-20
[App/Base] Rewrite Config and CVar implementation

Razzile:
A rewrite of the config system to add support for the following:
* allow cvars to be tagged as "unrestartable" which means that any changes won't take effect until next program restart
* provide a Config::Save() implementation to save the current state of the cvars back to the config file
* querying of, and access to a config var provided the variable's name
* miscelaneous changes to the internal workings of the config and cvar implementations. Cvars are now stored inside an instance of a `Config` class.

These changes don't affect anything in xenia right now, but will be required for the Qt UI's settings panel in the future

### DIFF
--- a/src/xenia/app/xenia_main.cc
+++ b/src/xenia/app/xenia_main.cc
@@ -228,7 +228,7 @@ int xenia_main(const std::vector<std::string>& args) {
   storage_root = std::filesystem::absolute(storage_root);
   XELOGI("Storage root: {}", xe::path_to_utf8(storage_root));
 
-  config::SetupConfig(storage_root);
+  Config::Instance().SetupConfig(storage_root);
 
   std::filesystem::path content_root = cvars::content_root;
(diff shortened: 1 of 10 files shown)


Razzile:
Not sure how to convert this into a draft PR but this PR isn't finished yet

I want to rip out CommandVar and just stick with ConfigVar asa transient ConfigVar behaves the same as a CommandVar and the latter wasn't being used in the codebase.

Razzile:
This refactor is now ready for review. One thing though @gibbed is we never came up with the proper terminology for a cvar that requires an app restart to take effect.

Triang3l:
Immutable? As long as this name isn't user-faced.

Razzile:
Immutable is a possibility for sure, but I worry it will make other devs think that this cvar cannot be changed at all, rather than requiring a restart.

Issue #1445, opened 2019-08-20
Xenia silently fails with invalid launch parameters


Margen67:
Currently, xenia doesn't give any sort of error or indication that an invalid launch parameter was used, and just closes itself.
This leads to end-users asking why their outdated/invalid launch parameters don't work.

x1nixmzeng:
What launch paramters does xenia require to boot? Fallback options should be enough rather than opening showing an error message

Margen67:
Falling back to the default would only be a good idea if xenia gives you an indication that it did so.

Razzile:
should xenia just ignore invalid launch parameters or should it actually error out. or maybe a middle ground where it prints a warning but continues?

Margen67:
@Razzile If you fall back without letting the user know then they'll still get confused.

gibbed:
Should be resolved since #1634 is now merged.

Issue #1516, opened 2019-12-03
[Config] Xenia does not print help when required


Razzile:
Xenia doesn't print help to command line, for example when running `xenia.exe --help`

This was probably an oversight when switching from gflags to our own config system

Prism019:
This issue was present before gflags was removed on Windows. The intended behaviour has worked fine in Linux since I first tried it (probably some time in 2017) and still works now. This is a Windows issue.

jgoyvaerts:
It happens because the PE on windows is built with IMAGE_SUBSYSTEM_WINDOWS_GUI set, which indicates it's a GUI application.

Cmd doesn't wait for the output of a gui application, it's just windows behavior. The same can be observed when you launch calc.exe or notepad.exe

There are possible workarounds (most applications use 2 executables to work around the issue), but they're not implemented in Xenia.

If you want the help output in windows, you can pipe the output to a file, or use mingw/git bash/some other terminal application instead of cmd (they do display the output).

More info:
https://stackoverflow.com/questions/54536/win32-gui-app-that-writes-usage-text-to-stdout-when-invoked-as-app-exe-help
https://devblogs.microsoft.com/oldnewthing/20090101-00/?p=19643

Razzile:
Interesting. I might see if I can apply one of those solutions to xenia

PR #1634, opened 2020-07-26
Invalid launch options no longer cause Xenia to exit silently

Significantbits:
This pull request aims to fix issue #1445  .  The cause of this issue is that the PrintHelpAndExit function was only writing to stdout out instead of notifying the user via a Message Box or something similar. 

This function gets called under two circumstances.

1.) The user gave the --help argument to Xenia. 
2.) The user gave invalid launch options to Xenia.

In order to differentiate between these cases, a boolean was added as an input to the function. This allowed me to put in a different message box for each of these cases. The --help message box displays the usage information, and the invalid argument message box tells the user that they gave invalid launch options then displays the usage information. 

image

image

This is my first pull request to this project, so hopefully everything follows the contribution guidelines. I tried to be careful, but please let me know if I did something wrong.

### DIFF
--- a/src/xenia/base/cvar.cc
+++ b/src/xenia/base/cvar.cc
@@ -14,6 +14,9 @@
 #define UTF_CPP_CPLUSPLUS 201703L
 #include "third_party/utfcpp/source/utf8.h"
 
+#include "xenia/base/logging.h"
+#include "xenia/base/main.h"
+
 namespace utfcpp = utf8;
 
 using u8_citer = utfcpp::iterator<std::string_view::const_iterator>;
@@ -61,7 +64,12 @@ void ParseLaunchArguments(int& argc, char**& argv,
 
     auto result = options.parse(argc, argv);
     if (result.count("help")) {
-      PrintHelpAndExit();
+      if (xe::has_console_attached()) {
+        PrintHelpAndExit();
+      } else {
+        xe::ShowInfoMessageBox(options.help({""}));
+        exit(0);
+      }
     }
 
     for (auto& it : *CmdVars) {
@@ -78,8 +86,15 @@ void ParseLaunchArguments(int& argc, char**& argv,
       }
     }
   } catch (const cxxopts::OptionException& e) {
-    std::cout << e.what() << std::endl;
-    PrintHelpAndExit();
+    if (xe::has_console_attached()) {
+      std::cout << e.what() << std::endl;
+      PrintHelpAndExit();
+    } else {
+      std::string m =
+          "Invalid launch options were given.\n" + options.help({""});
+      xe::ShowErrorMessageBox(m);
+      exit(0);
+    }
   }
 }
 
--- a/src/xenia/base/main_win.cc
+++ b/src/xenia/base/main_win.cc
@@ -38,8 +38,26 @@ bool has_console_attached_ = true;
 
 bool has_console_attached() { return has_console_attached_; }
 
+bool has_shell_environment_variable() {
+  size_t size = 0;
+  // Check if SHELL exists
+  // If it doesn't, then we are in a Windows Terminal
+  auto error = getenv_s(&size, nullptr, 0, "SHELL");
+  if (error) {
+    return false;
+  }
+  return !!size;
+}
+
 void AttachConsole() {
   if (!cvars::enable_console) {
+    return;
+  }
+
+  bool has_console = ::AttachConsole(ATTACH_PARENT_PROCESS) == TRUE;
+  if (!has_console || !has_shell_environment_variable()) {
+    // We weren't launched from a console, so just return.
+    has_console_attached_ = false;
     return;
   }
 
--- a/src/xenia/base/logging.cc
+++ b/src/xenia/base/logging.cc
@@ -363,4 +363,18 @@ void FatalError(const std::string_view str) {
   std::exit(1);
 }
 
+void ShowInfoMessageBox(std::string m) {
+#if XE_PLATFORM_WIN32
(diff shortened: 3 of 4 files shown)


jgoyvaerts:
Have you read this issue https://github.com/xenia-project/xenia/issues/1516 and tested your solution in mingw/bash based terminals? Do we really want both stdout AND a message box?

jgoyvaerts:
Also, take a look at logging.cc, there's already a message box being shown there in case of a fatal error. Can you reuse that function or write a message box function there instead of jamming the messagebox code in the middle of this cvar file?

Significantbits:
Please ignore the last push. I messed up. Fixing it now.

src/xenia/base/main_win.cc
```diff
@@ -58,6 +58,16 @@ void AttachConsole() {
   fp = _fdopen(con_handle, "w");
   *stderr = *fp;
   setvbuf(stderr, nullptr, _IONBF, 0);
+
+  char* buf = nullptr;
```
jgoyvaerts, line 62: Why these changes?

src/xenia/kernel/util/shim_utils.h
```diff
@@ -252,7 +252,7 @@ class PrimitivePointerParam : public ParamBase<uint32_t> {
   }
   T value() const { return *host_ptr_; }
   operator T() const = delete;
-  operator xe::be<T>*() const { return host_ptr_; }
+  operator xe::be<T> *() const { return host_ptr_; }
```
jgoyvaerts, line 255: Any reason for this change?

src/xenia/base/cvar.cc
```diff
@@ -60,7 +79,11 @@ void ParseLaunchArguments(int& argc, char**& argv,
 
     auto result = options.parse(argc, argv);
     if (result.count("help")) {
-      PrintHelpAndExit();
+      if (xe::has_console_attached()) {
+        PrintHelpAndExit();
+      } else {
+        ExitWithNotification(true);
```
jgoyvaerts, line 85: Instead of the ExitWithNotification function, make 1 or 2 functions in logging.cc (or somewhere else if there's a better place): ShowInfoMessageBox(string) and ShowErrorMessageBox(string), and use those directly

src/xenia/kernel/util/shim_utils.h
Significantbits: I think that happened when I did xb format. I didn't change that myself. I'll push again without these changes.

src/xenia/base/main_win.cc
Significantbits: This is to make it work with a bash environment or Windows terminal environment. I looked at issue #1516 like you said to. I tried implement the solution to redirect stdout and stderr to the parent process, but this broke the bash terminal environment. So, I check if the SHELL environment variable exists. If it doesn't exists, I assume a Windows terminal environment.

MSYS2
image
image

Windows Terminal
image

Significantbits: However, if there is no console attached then you will get the message box pops up. For example, if you give invalid arguments in Visual Studio, then you will see the Message Box.

jgoyvaerts: What do you mean to make it work? You check with xe::has_console_attached if there's a console attached, so what does this specific piece of code in main_win.cc do? What happens or doesn't work if you remove it?

Significantbits: xe::has_console_attched proves true if you launch from a Windows Terminal or bash environment. However, if you don't redirect stdout and stderr you won't see the print message in the Windows terminal, and Xenia will just exit silently like before. If you redirect stdout and stderr, then you won't see the message if Xenia is launched from a bash environment.  This code checks to see which environment Xenia was launched from and either redirects stdout and stderr, or doesn't redirect it.

So, in short, if you remove it, Xenia will still exit silently if launched from a Windows Terminal. 

I thought just getting a print message in the console would be better than a Message Box if xe::has_console_attached is true.

jgoyvaerts: If I can see your screenshot correctly, you're writing to a terminal which is not expecting output, which messes up the console for no reason.

As I said in #1516 , the windows Cmd and Powershell do not wait for output when launching Xenia through the command line. Instead of finding a quirky workaround to still output to a console which isn't expecting output, isn't it easier to fix has_console_attached to return false in those cases? (See AttachConsole in main_win.cc and feel free to improve it)

Significantbits: Okay. I'll fix has_console_attached so that a Message Box will pop up if you are using the a Windows terminal.

src/xenia/base/cvar.cc
Significantbits: I did this in the latest push. Sorry for misunderstanding the previous comment on this.

src/xenia/base/cvar.cc
```diff
@@ -24,6 +27,20 @@ cxxopts::Options options("xenia", "Xbox 360 Emulator");
 std::map<std::string, ICommandVar*>* CmdVars;
 std::map<std::string, IConfigVar*>* ConfigVars;
 
+/*void ExitWithNotification(bool help) {
```
jgoyvaerts, line 30: There's commented code here, you should probably clean this up

jgoyvaerts:
If I was you I'd split up the commit in 2 or 3 commits
- One for the fixing/improving of has_console_attached
- One for the adding of the helper methods in logging.cc
- One for the changing of the behavior in cvar.cc

src/xenia/base/cvar.cc
Significantbits: Sorry about that. Should be gone now.

PR #1635, opened 2020-08-01
[Kernel/Thread] Changed default stack location

Gliniak:
<Naming still to decide, suggestions are welcome 🐑 >

### TL;DR

It seems like actual thread stack allocation location is not valid

### Nier Case
After spending a while why it fixed Nier it seems like Nier overwrites actual thread stack space with it's own data.

I spent while checking if it affects other games (around 30+ titles). It looks like it's not, but as we can see some games might be affected.

### DIFF
--- a/src/xenia/kernel/xthread.cc
+++ b/src/xenia/kernel/xthread.cc
@@ -226,7 +226,7 @@ void XThread::InitializeGuestObject() {
 }
 
 bool XThread::AllocateStack(uint32_t size) {
-  auto heap = memory()->LookupHeap(0x40000000);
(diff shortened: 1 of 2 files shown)


gibbed:
You've missed a memory address constant in the `AllocRange` call.

Gliniak:
I've updated PR a little bit:
- Changed naming and convention for constant values
- Changed static const to static constexpr
- Put static values into XThread class (as they're related/used only in that class)

Gliniak:
Update:
- I removed part of PR with changing kernel stack location. I have no clear information where it should be stored, so for now let it be where it was.

PR #1636, opened 2020-08-02
[Kernel/IO] Return error when creating directory with non-directory flag in NtCreateFile

Gliniak:
Different title suggestions will be appreciated 👍 

I found out that fifa 07 tries to do this:

```
d> 00000028 NtOpenFile(7018F590(00000000), 00100001, 7018F5B0(FFFFFFFD,D:\,00000040), 7018F5A8, 00000003)
.
.
.
d> 00000088 RtlInitAnsiString(701EF8E8, 701EF990(D:\))
d> 00000088 NtCreateFile(701EF8E0(FFFFFFFF), 80100080, 701EF8F8(FFFFFFFD,D:\,00000040), 701EF8F0, 00000000, 00000080, 00000001, 00000001, 00000060)

```

For the first case it's fine it should return 0, but for second not so fast. 
It receives only D:\ (checked also during debugging) as a target_path without any file name with flag 0x40 (FILE_NON_DIRECTORY_FILE)

In such case we should return X_STATUS_FILE_IS_A_DIRECTORY instead of 0.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
@@ -142,7 +142,8 @@ dword_result_t NtCreateFile(lpdword_t handle_out, dword_t desired_access,
   X_STATUS result = kernel_state()->file_system()->OpenFile(
       root_entry, target_path,
       vfs::FileDisposition((uint32_t)creation_disposition), desired_access,
-      (create_options & CreateOptions::FILE_DIRECTORY_FILE) != 0, &vfs_file,
+      (create_options & CreateOptions::FILE_DIRECTORY_FILE) != 0,
+      (create_options & CreateOptions::FILE_NON_DIRECTORY_FILE) != 0, &vfs_file,
       &file_action);
   object_ref<XFile> file = nullptr;
 
--- a/src/xenia/kernel/xfile.cc
+++ b/src/xenia/kernel/xfile.cc
@@ -266,7 +266,7 @@ object_ref<XFile> XFile::Restore(KernelState* kernel_state,
   vfs::FileAction action;
   auto res = kernel_state->file_system()->OpenFile(
       nullptr, abs_path, vfs::FileDisposition::kOpen, access, is_directory,
-      &vfs_file, &action);
+      false, &vfs_file, &action);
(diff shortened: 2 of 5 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
```diff
@@ -105,6 +105,11 @@ dword_result_t NtCreateFile(lpdword_t handle_out, dword_t desired_access,
     return X_STATUS_OBJECT_NAME_INVALID;
   }
 
+  if (target_path.back() == '\\' &&
```
Triang3l, line 108: Not really sure overall about this way of checking whether it's a directory, though need to read how that object structure is filled, and where does the path originate from (if from the game, it will not necessarily probably include the \, if from Xenia's file system code, then maybe). Also if the path is empty, this will cause an error.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
```diff
@@ -105,6 +105,12 @@ dword_result_t NtCreateFile(lpdword_t handle_out, dword_t desired_access,
     return X_STATUS_OBJECT_NAME_INVALID;
   }
 
+  std::error_code directory_error_code;
+  if (std::filesystem::is_directory(target_path, directory_error_code) &&
```
Triang3l, line 109: Will this work with virtual file systems? I think is_directory should ask the host OS, not just check the path.

gibbed: Yeah, it shouldn't be touching the real FS at all. Should be an invoke into the VFS.

Gliniak: I agree. I didn't factor that it touches real file system

Gliniak:
I updated this PR:
 - Changed location of ``CreateOptions`` (this I'm not exactly sure where it should be. ``xernel/xfile.h`` or ``vfs/file.h``.
 - Changed ``OpenFile`` parameter from ``bool is_directory`` to ``uint32_t createOptions`` to provide more robust implementation (allows more egde case checking like this one)

PR #1637, opened 2020-08-04
[Kernel/Memory] Restrict Nt<whatever>VirtualMemory only to virtual memory range

Gliniak:
TL;DR

NtAllocateVirtualMemory, NtFreeVirtualMemory and NtProtectVirtualMemory shouldn't be able to allocate memory over physical or xex range.

I'm making it as draft for future updates (switch from hardcoded value into some const etc)

<This PR is for testing purposes right now.>

Affected titles (positive):
- Stormrise
- London 2012
- Dead or Alive 4

Affected titles (negative):
-TBD

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
@@ -101,6 +101,9 @@ dword_result_t NtAllocateVirtualMemory(lpdword_t base_addr_ptr,
   if (*base_addr_ptr != 0) {
     // ignore specified page size when base address is specified.
     auto heap = kernel_memory()->LookupHeap(*base_addr_ptr);
+    if (heap->heap_type() != HeapType::kGuestVirtual) {
+      return X_STATUS_INVALID_PARAMETER;
+    }
     page_size = heap->page_size();
   } else {
     // Adjust size.
@@ -192,7 +195,9 @@ dword_result_t NtProtectVirtualMemory(lpdword_t base_addr_ptr,
   }
 
   auto heap = kernel_memory()->LookupHeap(*base_addr_ptr);
-
+  if (heap->heap_type() != HeapType::kGuestVirtual) {
+    return X_STATUS_INVALID_PARAMETER;
+  }
   // Adjust the base downwards to the nearest page boundary.
   uint32_t adjusted_base =
       *base_addr_ptr - (*base_addr_ptr % heap->page_size());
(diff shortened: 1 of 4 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -101,6 +101,10 @@ dword_result_t NtAllocateVirtualMemory(lpdword_t base_addr_ptr,
   if (*base_addr_ptr != 0) {
     // ignore specified page size when base address is specified.
     auto heap = kernel_memory()->LookupHeap(*base_addr_ptr);
+
```
Triang3l, line 104: Please balance the new lines, no need for new lines between getting the heap and checking if it's usable.

src/xenia/kernel/xfile.cc
```diff
@@ -122,14 +122,13 @@ X_STATUS XFile::Read(uint32_t buffer_guest_address, uint32_t buffer_length,
       const xe::BaseHeap* buffer_end_heap =
           memory()->LookupHeap(buffer_guest_high_address);
       if (!buffer_start_heap || !buffer_end_heap ||
-          buffer_start_heap->IsGuestPhysicalHeap() !=
-              buffer_end_heap->IsGuestPhysicalHeap() ||
-          (buffer_start_heap->IsGuestPhysicalHeap() &&
+          buffer_start_heap->heap_type() != buffer_end_heap->heap_type() ||
```
Triang3l, line 125: This was specifically for disallowing access across the boundary of virtual and physical heaps, as there is special logic for  invalidating GPU memory in guest physical heaps (this isn't the correct behavior anyway from virtual memory perspective, but I made an assumption that this normally would result in an access violation anyway due to writing something near physical nullptr). While probably it doesn't make sense to touch XEX pages at all, the point in the original design was specifically to handle guest physical heaps — it's better to state intentions clearly than in a way that just happens to be work.

src/xenia/memory.h
```diff
@@ -36,6 +36,13 @@ enum SystemHeapFlag : uint32_t {
   kSystemHeapDefault = kSystemHeapVirtual,
 };
 
+enum HeapTypeFlag : uint8_t {
```
Triang3l, line 39: These are mutually exclusive, so should be just an `enum class`, not bits (and `HeapType` then, not `HeapTypeFlag`).

Triang3l (changes requested):
(just a comment because I missed the "Request changes" radio button)

src/xenia/kernel/xfile.cc
```diff
@@ -122,14 +122,12 @@ X_STATUS XFile::Read(uint32_t buffer_guest_address, uint32_t buffer_length,
       const xe::BaseHeap* buffer_end_heap =
           memory()->LookupHeap(buffer_guest_high_address);
       if (!buffer_start_heap || !buffer_end_heap ||
-          buffer_start_heap->IsGuestPhysicalHeap() !=
-              buffer_end_heap->IsGuestPhysicalHeap() ||
-          (buffer_start_heap->IsGuestPhysicalHeap() &&
+          (buffer_start_heap->heap_type() == HeapType::kGuestPhysical &&
```
Triang3l, line 125: What check that does that too were you referring to in DMs? This is for cases like reading 16 bytes to 0x9FFFFFF8, where the read crosses the boundary. Without this check, this read will be treated as just reading to virtual memory.

Triang3l (changes requested):

PR #1638, opened 2020-08-07
[App] Add portable as a launch option

jgoyvaerts:
Small PR to add portable as a launch option in addition to still checking for existence of portable.txt.

Need this change for a regression build system I'm working on, and it seems cleaner this way.


Razzile:
Good idea. Don't know why this wasn't thought of before

PR #1642, opened 2020-08-22
Fix xenia.log file not always being created in the executable folder.

jgoyvaerts:
When launching xenia from the command line, and your current working directory was not the xenia directory, the log file would be created in your current working directory instead of the executable folder.

For example: 

image

This pullrequest fixes that, and creates the log file in the executable folder (unless the user manually specified a log file location of course)


Margen67:
Fixes #1949

PR #1482, opened 2019-09-25
Support Python from the Windows Store when building

Joshua-Ashton:

PR #1643, opened 2020-08-26
Add PowerShell build script as alternative to cmd

Razzile:
Provide a PowerShell script as an alternative to the cmd script.
This has not been extensively tested but on my local machine has fixed #1482

### DIFF
--- a/xb.ps1
+++ b/xb.ps1
@@ -1,0 +1,24 @@
+function Write-FatalError($message) {
+    [Console]::ForegroundColor = 'red'
+    [Console]::Error.WriteLine($message)
+    [Console]::ResetColor()
+    Exit 1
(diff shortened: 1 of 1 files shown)


jgoyvaerts:
What's the intended purpose of this? Won't we have to maintain 2 versions (the cmd version and the powershell version), without a significant benefit?

Razzile:
@jgoyvaerts currently the cmd version doesn't work correctly with some python installations. I didn't know how to fix so I made a powershell alternative. Rick can choose if he wants to maintain one, or keep them both simultaneously (neither file is very big). Personally, I think we should move to powershell as Microsoft is shifting most command line things in Windows to PowerShell now

PR #1646, opened 2020-08-29
[Kernel/Object] Added base to object handle

Gliniak:
In first versions of Burnout Paradise there is code that rely on correct handle value.

image

uVar1 is handle to the thread that is stored via Tls.

If you calculate with actual values it looks like we're off to incorrect memory space (around 0x8A, but we should be at 0x82...)

Affected Titles:
- Burnout Paradise v1.0 (including demo)

### DIFF
--- a/src/xenia/kernel/util/object_table.cc
+++ b/src/xenia/kernel/util/object_table.cc
@@ -112,8 +112,7 @@ X_STATUS ObjectTable::AddHandle(XObject* object, X_HANDLE* out_handle) {
       ObjectTableEntry& entry = table_[slot];
       entry.object = object;
       entry.handle_ref_count = 1;
-
-      handle = slot << 2;
+      handle = XObject::kHandleBase + (slot << 2);
       object->handles().push_back(handle);
 
       // Retain so long as the object is in the table.
@@ -251,7 +250,7 @@ ObjectTable::ObjectTableEntry* ObjectTable::LookupTable(X_HANDLE handle) {
   auto global_lock = global_critical_region_.Acquire();
 
   // Lower 2 bits are ignored.
-  uint32_t slot = handle >> 2;
+  uint32_t slot = GetHandleSlot(handle);
   if (slot <= table_capacity_) {
     return &table_[slot];
   }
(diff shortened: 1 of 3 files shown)


src/xenia/kernel/xthread.h
```diff
@@ -30,6 +30,7 @@ class XEvent;
 constexpr uint32_t X_CREATE_SUSPENDED = 0x00000001;
 
 constexpr uint32_t X_TLS_OUT_OF_INDEXES = UINT32_MAX;
+constexpr uint32_t X_THREAD_MASK = 0xF8000000;
```
Triang3l, line 33: Please add a note about how Burnout Paradise uses this. Also is THREAD_MASK the original constant name from the NT kernel? If not, it should be in the kName format, not X_NAME, and possibly not in this file.

Triang3l (changes requested):

src/xenia/kernel/xthread.h
```diff
@@ -105,6 +105,8 @@ static_assert_size(X_KTHREAD, 0xAB0);
 
 class XThread : public XObject, public cpu::Thread {
  public:
+  // Burnout Paradise needs proper handle value for certain calculations
```
Triang3l, line 108: This isn't very descriptive for the future, would be better to reword the PR text in the commit, with the calculation, and the 0x8A/0x82 part.

src/xenia/kernel/xthread.h
```diff
@@ -105,6 +105,12 @@ static_assert_size(X_KTHREAD, 0xAB0);
...
+  // It gets handle value from TLS (without base handle value is 0x88)
+  // and substract 0xF8000088. Without base we're receiving wrong address
+  // Instead of receiving address that starts with 0x82... we're receiving
+  // one with 0x8A... which causes crash
+  static constexpr uint32_t kHandleBase = 0xF8000000;
```
Triang3l, line 113: XObject, not XThread, would be the more appropriate place.

src/xenia/kernel/util/object_table.cc
```diff
@@ -251,7 +250,7 @@ ObjectTable::ObjectTableEntry* ObjectTable::LookupTable(X_HANDLE handle) {
   auto global_lock = global_critical_region_.Acquire();
 
   // Lower 2 bits are ignored.
-  uint32_t slot = handle >> 2;
+  uint32_t slot = (handle - XThread::kHandleBase) >> 2;
```
Triang3l, line 253: Can `(handle - XThread::kHandleBase) >> 2` be put in an inline (or even constexpr, since the implementation allows) function like ObjectTable::GetSlot?

PR #1649, opened 2020-09-14
[Kernel/XAM] Allow sharing xam_info.cc functions between many xam files

Gliniak:
This file can be used to provide support for xam functions that are used frequently between different xam functions.

I already pasted:
``XamCreateEnumeratorHandle``
``XamGetPrivateEnumStructureFromHandle``

as I know they're used between few XAM functions (related to enumerators) and almost implemented them locally.

Feel free to suggest any improvements/changes 💃

### DIFF
--- a/src/xenia/kernel/xam/xam_info.h
+++ b/src/xenia/kernel/xam/xam_info.h
@@ -1,0 +1,19 @@
+#pragma once
+#include "xenia/kernel/kernel_state.h"
+#include "xenia/kernel/util/shim_utils.h"
+#include "xenia/kernel/xam/xam_private.h"
+
+namespace xe {
+namespace kernel {
+namespace xam {
+dword_result_t XamCreateEnumeratorHandle(unknown_t unk1, unknown_t unk2,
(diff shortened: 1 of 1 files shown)


Triang3l:
What was wrong with their original placement in `xam_info.cc`?

Gliniak:
They can be exposed in xam_info if we create header file (as you said), but my idea was to store here only functions which are used in different xam functions.
About organization part. Yeah I agree, but many functions don't match any pattern to put them in other categories

Triang3l:
But it still would be a bit messy to put, like, enumerators alongside some Direct3D functions because they are both are just used in 2 files.

Gliniak:
We might overcome this issue (kinda straightforward idea), by adding header file to each xam category

Triang3l:
Yes, should do functionally the same, but be more structured. Only when you really need that header though, for really externally callable functions.

Gliniak:
Updated 🐑

Triang3l:
Until these functions are actually used externally anywhere, there's no point in exposing them, however, once they are actually used, I think this should be the right way to make them visible.

PR #1650, opened 2020-09-17
[Kernel/Xam] XNotifyGetNext - Check for null in id_ptr

Gliniak:
Really small change to XNotifyGetNext based on _Carcassonne_ game code.

image

PS. Game still gonna crash due to returning error code in _XUsbcamCreate_, but that will be in another PR, as I still don't know whole story 🦊

### DIFF
--- a/src/xenia/kernel/xam/xam_notify.cc
+++ b/src/xenia/kernel/xam/xam_notify.cc
@@ -42,15 +42,19 @@ DECLARE_XAM_EXPORT1(XamNotifyCreateListener, kNone, kImplemented);
 // https://github.com/CodeAsm/ffplay360/blob/master/Common/AtgSignIn.cpp
 dword_result_t XNotifyGetNext(dword_t handle, dword_t match_id,
                               lpdword_t id_ptr, lpdword_t param_ptr) {
-  if (!handle) {
-    return 0;
+  if (param_ptr) {
+    *param_ptr = 0;
   }
 
+  if (!id_ptr) {
+    return X_ERROR_INVALID_PARAMETER;
+  }
+  *id_ptr = 0;
   // Grab listener.
   auto listener =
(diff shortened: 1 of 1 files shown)


src/xenia/kernel/xam/xam_notify.cc
```diff
@@ -74,7 +74,10 @@ dword_result_t XNotifyGetNext(dword_t handle, dword_t match_id,
       *param_ptr = param;
     }
   } else {
-    *id_ptr = 0;
+    // id_ptr may be null - "Carcassonne" explicitly passes nullptr in code
+    if (id_ptr) {
```
Triang3l, line 78: Not handled in the `if (dequeued)` above.

Triang3l (changes requested):

Triang3l:
Also should the first two early-outs in that function write anything to id_ptr and param_ptr?

src/xenia/kernel/xam/xam_notify.cc
Gliniak: I don't know if it should be handled in that case too. Maybe it should, but we have no case of such behaviour

Triang3l: If it's a nullable parameter in any case ("maybenull" in SAL terms), null pointer should be passable in all cases, I think.

Triang3l:
Was told by @Gliniak in PMs that it returns E_INVALIDARG in the official kernel if either of those arguments is null, so probably need to recheck all the return values of the functions.

Gliniak:
After spending some time I figured out that:
E_INVALIDARG (X_ERROR_INVALID_PARAMETER) - is returned only if id_ptr is null
E_UNEXPECTED is returned only when data under handle is corrupted in some specific way, but not handle itself

PR #1651, opened 2020-09-17
[Kernel/Usbcam] XUsbcamCreate

Gliniak:
- Provided description to some parameters
- Provided description what function more/less should do
- Changed return code based on game behaviour

Comment:
Based on fragment of game code:
image

I tested few values, but only 0 seems to be reasonable.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_usbcam.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_usbcam.cc
@@ -17,11 +17,17 @@ namespace xe {
 namespace kernel {
 namespace xboxkrnl {
 
-dword_result_t XUsbcamCreate(unknown_t unk1,  // E
-                             unknown_t unk2,  // 0x4B000
+dword_result_t XUsbcamCreate(dword_t buffer,
+                             dword_t buffer_size,  // 0x4B000 640x480?
                              lpunknown_t unk3_ptr) {
-  // 0 = success.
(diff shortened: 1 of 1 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_usbcam.cc
```diff
@@ -17,11 +17,15 @@ namespace xe {
...
+dword_result_t XUsbcamCreate(dword_t buffer,
+                             dword_t buffer_size,  // 0x4B000 640x480?
                              lpunknown_t unk3_ptr) {
-  // 0 = success.
-  return X_ERROR_DEVICE_NOT_CONNECTED;
+  // This function should return success.
+  // It looks like it only allocates space for usbcam support.
+  // returning error code might cause games to initialize incorrectly.
+  // "Carcassonne" initalization function checks for result from this
+  // function. If value is different than 0 it ends function with nullptr.
```
Triang3l, line 27: "ends function with nullptr" sounds weird, more like "dereferences a null pointer and crashes" (if you can explain what that null pointer is and how it's obtained, better explain that here too).

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_usbcam.cc
Triang3l: (saw the disassembly, better explain what it does with nullsub)

PR #1654, opened 2020-09-27
[Kernel/Thread] Added missing paramteter to KeSetAffinityThread

Gliniak:
I found out that ``KeSetAffinityThread`` have optional parameter. I called it ``affinity_ptr``, but feel free to provide better name 

image

Here we can see few things.
1. Param2 must be lower than 6 (cpu_core?)
2. KeSetAffinityThread have 3 parameters (3rd is a pointer)
3. Output of that function depends on value under 3rd parameter.

These 2 fragments are usages of that function above. We can see that second argument is check on input and output and that parameter is used as affinity (to be precise: 1 << param2)
image
image

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_threading.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_threading.cc
@@ -222,13 +222,23 @@ void KeSetCurrentStackPointers(lpvoid_t stack_ptr,
 }
 DECLARE_XBOXKRNL_EXPORT1(KeSetCurrentStackPointers, kThreading, kImplemented);
 
-dword_result_t KeSetAffinityThread(lpvoid_t thread_ptr, dword_t affinity) {
-  auto thread = XObject::GetNativeObject<XThread>(kernel_state(), thread_ptr);
-  if (thread) {
-    thread->SetAffinity(affinity);
+dword_result_t KeSetAffinityThread(lpvoid_t thread_ptr, dword_t affinity,
+                                   lpdword_t previous_affinity_ptr) {
+  // Xbox 360 uses additional parameter (in comparation to NT equivalent)
+  // which is used only for returning previous thread affinity. (Based on code
+  // dissasembly)
+  if (!affinity) {
+    return X_STATUS_INVALID_PARAMETER;
   }
 
-  return (uint32_t)affinity;
+  auto thread = XObject::GetNativeObject<XThread>(kernel_state(), thread_ptr);
+  if (thread) {
+    if (previous_affinity_ptr) {
+      *previous_affinity_ptr = 1 << thread->active_cpu();
+    }
+    thread->SetAffinity(affinity);
+  }
+  return X_STATUS_SUCCESS;
 }
 DECLARE_XBOXKRNL_EXPORT1(KeSetAffinityThread, kThreading, kImplemented);
 
--- a/src/xenia/kernel/xthread.cc
+++ b/src/xenia/kernel/xthread.cc
@@ -205,6 +205,7 @@ void XThread::InitializeGuestObject() {
   // 0xA88 = APC
   // 0x18 = timer
   xe::store_and_swap<uint32_t>(p + 0x09C, 0xFDFFD7FF);
+  xe::store_and_swap<uint8_t>(p + 0xBF, 0);
   xe::store_and_swap<uint32_t>(p + 0x0D0, stack_base_);
   xe::store_and_swap<uint64_t>(p + 0x130, Clock::QueryGuestSystemTime());
   xe::store_and_swap<uint32_t>(p + 0x144, guest_object() + 0x144);
@@ -346,6 +347,9 @@ X_STATUS XThread::Create() {
   // Exports use this to get the kernel.
   thread_state_->context()->kernel_state = kernel_state_;
 
+  // Initialize the KTHREAD object.
+  InitializeGuestObject();
+
   X_KPCR* pcr = memory()->TranslateVirtual<X_KPCR*>(pcr_address_);
 
   pcr->tls_ptr = tls_static_address_;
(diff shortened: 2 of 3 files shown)


gibbed:
Given the return value is already the affinity, I would guess it's more likely the pointer argument is for the previous value, if the caller wants it.

Gliniak:
Point with returning previous affinity make sense, as it is optional parameter. I will edit PR in that way

Gliniak:
Updated 👍

JoelLinn:
Mh, is the function you posted the implementation of `XSetThreadProcessor(thread_hndl, core_id)` ?
It's briefly mentioned [here](https://docs.microsoft.com/en-us/windows/win32/dxtecharts/coding-for-multiple-cores) and found on GitHub all over the place.

Triang3l:
Can you show how r5 is used in the disassembly of KeSetAffinityThread?

Gliniak:
image

**(I cutted code to only contains informations that are related to this topic)**

So yeah, that is sorta correct, but instead of handling it via core_id (offset 0xBF) in XTHREAD we're handling it via host variable.
It would be great to change it to handle via guest side, but I'm not sure where exactly core_id should be set on guest side
I think it should be in ExCreateThread, but where else?

Triang3l:
https://docs.microsoft.com/en-us/windows/win32/dxtecharts/coding-for-multiple-cores
> On Xbox 360, you must explicitly assign software threads to a particular hardware thread by using XSetThreadProcessor. Otherwise, all child threads will stay on the same hardware thread as the parent.
At creation time, we should obtain the core ID of the thread that invoked the creation kernel call and duplicate it for the new thread. If using guest affinities on the host, we should also assign the thread to the same host core as the thread that created it. Then, games should set the core explicitly if they want to move the thread to some other code.

Gliniak:
Updated!

Switched into guest side handling and usage of core_id for affinity calculation

PS. Travis died?

src/xenia/kernel/xboxkrnl/xboxkrnl_threading.cc
```diff
@@ -222,12 +222,15 @@ void KeSetCurrentStackPointers(lpvoid_t stack_ptr,
...
+                                   lpdword_t previous_affinity_ptr) {
   auto thread = XObject::GetNativeObject<XThread>(kernel_state(), thread_ptr);
   if (thread) {
+    if (previous_affinity_ptr) {
+      *previous_affinity_ptr = 1 << thread->active_cpu();
+    }
     thread->SetAffinity(affinity);
   }
-
   return (uint32_t)affinity;
```
Triang3l, line 234: The return value should be HRESULT according to the disassembly findings, please add handling of all the result codes from the original function that aren't too complex to integrate into the emulator's architecture (and if you skip any, leave some // comments about them for the possible implementation in the future).

src/xenia/kernel/xthread.cc
```diff
@@ -729,6 +729,13 @@ void XThread::SetActiveCpu(uint32_t cpu_index) {
   assert_true(cpu_index < 6);
   uint8_t* pcr = memory()->TranslateVirtual(pcr_address_);
   xe::store_and_swap<uint8_t>(pcr + 0x10C, cpu_index);
+
+  if (is_guest_thread()) {
+    X_KTHREAD* thread_object =
+        memory()->TranslateVirtual<X_KTHREAD*>(guest_object());
+
```
Triang3l, line 736: This empty line doesn't separate any significant portions.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_threading.cc
Triang3l: Also it would be good to have comments explaining the difference between the interfaces of KeSetAffinityThread on the Xbox 360 and on Windows, why there's a new parameter that isn't there in NT documentation or Wine, referencing the disassembly as the source.

Triang3l:
`affinity_` and `affinity()` of XThread are not used anywhere, maybe would be good to remove it (the core ID is already in 2 places).

src/xenia/kernel/xthread.h
```diff
@@ -88,7 +88,8 @@ struct X_KTHREAD {
   char unk_10[0xAC];             // 0x10
   uint8_t suspend_count;         // 0xBC
   uint8_t unk_BD;                // 0xBD
-  uint16_t unk_BE;               // 0xBE
+  uint8_t unk_BE;                // 0xBE
+  uint8_t core_id;               // 0xBF
```
Triang3l, line 92: What's the source of the `core_id` name? It maybe a bit confusing that it's different than `current_cpu` if that's not the original name of the variable (Windows NT uses the word "Processor" for similar purposes also).

Triang3l, line 92: This needs to be initialized for new threads in XThread::Create.

Triang3l (changes requested):

src/xenia/kernel/xthread.cc
```diff
@@ -358,11 +359,14 @@ X_STATUS XThread::Create() {
...
       static_cast<uint8_t>(creation_params_.creation_flags >> 24);
 
-  pcr->current_cpu = GetFakeCpuNumber(proc_mask);  // Current CPU(?)
-  pcr->dpc_active = 0;                             // DPC active bool?
+  uint8_t current_cpu = GetFakeCpuNumber(proc_mask);  // Current CPU(?)
+  pcr->current_cpu = current_cpu;
+  pcr->dpc_active = 0;  // DPC active bool?
 
   // Initialize the KTHREAD object.
   InitializeGuestObject();
```
Triang3l, line 367: The new field in KTHREAD also needs to be initialized to the core index.

Triang3l (changes requested):

PR #1660, opened 2020-10-11
[Kernel] Implemented ExLoadedImageName

Gliniak:
This is based on clues inside freestyledash code (Thanks 💃)

This resolve one of reasons why custom dashes usually cannot run.

PS. I'm not sure about one thing. Does maximal path length on xbox matches defined _MAX_PATH? (and the alignment 🤷‍♂️)

image

### DIFF
--- a/src/xenia/kernel/kernel_state.cc
+++ b/src/xenia/kernel/kernel_state.cc
@@ -316,6 +316,17 @@ void KernelState::SetExecutableModule(object_ref<UserModule> module) {
     *variable_ptr = executable_module_->hmodule_ptr();
   }
 
+  // Setup the kernel's ExLoadedImageName field
+  export_entry = processor()->export_resolver()->GetExportByOrdinal(
+      "xboxkrnl.exe", ordinals::ExLoadedImageName);
+
+  if (export_entry) {
+    char* variable_ptr =
+        memory()->TranslateVirtual<char*>(export_entry->variable_ptr);
+    xe::string_util::copy_truncating(
+        variable_ptr, executable_module_->path(),
+        xboxkrnl::XboxkrnlModule::kExLoadedImageNameSize);
+  }
   // Spin up deferred dispatch worker.
   // TODO(benvanik): move someplace more appropriate (out of ctor, but around
   // here).
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
@@ -189,6 +189,16 @@ XboxkrnlModule::XboxkrnlModule(Emulator* emulator, KernelState* kernel_state)
                                        ordinals::XexExecutableModuleHandle,
                                        ppXexExecutableModuleHandle);
 
+  // ExLoadedImageName (char*)
+  // The full path to loaded image/xex including its name.
+  // Used usually in custom dashboards (Aurora)
+  // Todo(Gliniak): Confirm that official kernel always allocate space for this
+  // variable.
+  uint32_t ppExLoadedImageName =
+      memory_->SystemHeapAlloc(kExLoadedImageNameSize);
+  export_resolver_->SetVariableMapping(
+      "xboxkrnl.exe", ordinals::ExLoadedImageName, ppExLoadedImageName);
+
   // ExLoadedCommandLine (char*)
   // The name of the xex. Not sure this is ever really used on real devices.
   // Perhaps it's how swap disc/etc data is sent?
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_module.h
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_module.h
@@ -27,6 +27,8 @@ namespace xboxkrnl {
 
 class XboxkrnlModule : public KernelModule {
  public:
+  static constexpr size_t kExLoadedImageNameSize = 255 + 1;
+
   XboxkrnlModule(Emulator* emulator, KernelState* kernel_state);
   virtual ~XboxkrnlModule();
 


src/xenia/kernel/kernel_state.cc
```diff
@@ -316,6 +316,20 @@ void KernelState::SetExecutableModule(object_ref<UserModule> module) {
...
+  export_entry = processor()->export_resolver()->GetExportByOrdinal(
+      "xboxkrnl.exe", ordinals::ExLoadedImageName);
+
+  if (export_entry && module) {
+    auto variable_ptr = memory()->TranslateVirtual<xe::be<uint32_t>*>(
+        export_entry->variable_ptr);
+    std::string image_name(module->path());
+    uint32_t image_line_length =
+        xe::align(static_cast<uint32_t>(image_name.length()) + 1, 256u);
+    std::memset(variable_ptr, 0, image_line_length);
```
Triang3l, line 329: Why an entire memset rather than just writing the null terminator (strcpy)?

src/xenia/kernel/kernel_state.cc
```diff
@@ -316,6 +316,20 @@ void KernelState::SetExecutableModule(object_ref<UserModule> module) {
...
+  // Setup the kernel's ExLoadedImageName field
+  export_entry = processor()->export_resolver()->GetExportByOrdinal(
+      "xboxkrnl.exe", ordinals::ExLoadedImageName);
+
+  if (export_entry && module) {
+    auto variable_ptr = memory()->TranslateVirtual<xe::be<uint32_t>*>(
+        export_entry->variable_ptr);
+    std::string image_name(module->path());
+    uint32_t image_line_length =
+        xe::align(static_cast<uint32_t>(image_name.length()) + 1, 256u);
```
Triang3l, line 328: Why is this aligned? MAX_PATH (260), if applicable here even, is the absolute limit, not an increment.

Triang3l (changes requested):

src/xenia/kernel/kernel_state.cc
Gliniak: Copy pasting code 🤣 I will change it in a second

src/xenia/kernel/kernel_state.cc
```diff
@@ -316,6 +316,23 @@ void KernelState::SetExecutableModule(object_ref<UserModule> module) {
...
+      "xboxkrnl.exe", ordinals::ExLoadedImageName);
+
+  if (export_entry) {
+    char* variable_ptr =
+        memory()->TranslateVirtual<char*>(export_entry->variable_ptr);
+    size_t length = 0;
+
+    if (module) {
+      std::string image_name(module->path());
+      length = std::min(size_t(FILENAME_MAX - 1), image_name.size());
```
Triang3l, line 330: That should be the guest maximum file name size since this structure writes to the guest memory, not the host one.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
```diff
@@ -186,6 +186,10 @@ XboxkrnlModule::XboxkrnlModule(Emulator* emulator, KernelState* kernel_state)
                                        ordinals::XexExecutableModuleHandle,
                                        ppXexExecutableModuleHandle);
 
+  uint32_t ppExLoadedImangeName = memory_->SystemHeapAlloc(FILENAME_MAX);
```
Triang3l, line 189: Apart from the FILENAME_MAX issue (even though we allocate this memory by ourselves, games using this string likely don't expect it to be longer than MAX_PATH), there's a typo ("imange").

Triang3l (changes requested):

src/xenia/kernel/kernel_state.cc
```diff
@@ -316,6 +317,24 @@ void KernelState::SetExecutableModule(object_ref<UserModule> module) {
...
+        memory()->TranslateVirtual<char*>(export_entry->variable_ptr);
+    size_t length = 0;
+
+    if (module) {
+      std::string image_name(module->path());
+      length = std::min(size_t(XFile::kMaximalFileNameLength - 1),
+                        image_name.size());
+      std::memcpy(variable_ptr, image_name.c_str(), length);
+    }
+    variable_ptr[length] = '\0';
```
Triang3l, line 335: We have xe::string_util::copy_truncating now for this :)

Triang3l (changes requested):

src/xenia/kernel/kernel_state.cc
Gliniak: Changed ;) Also figured out a little mistake I've made last time 🐑

src/xenia/kernel/xfile.h
```diff
@@ -76,6 +76,7 @@ class X_FILE_DIRECTORY_INFORMATION {
 class XFile : public XObject {
  public:
   static const XObject::Type kObjectType = XObject::Type::File;
+  static constexpr uint16_t kMaximalFileNameLength = 260;
```
gibbed, line 79: `FileName` should probably be `Path` or changed to `Name` (as it's already in `XFile`, so you'd be accessing it like `XFile::kMaximalNameLength`).

gibbed (changes requested):

src/xenia/kernel/xfile.h
Gliniak: I was thinking about it as it wasn't making much sense, but ignored it 😸. Changed to kMaximalNameLength

gibbed: Thinking about it more, it should most certainly be `Path` since it's the maximum length of the entire thing, not individual name components.

Gliniak: Done

gibbed (approved):

src/xenia/kernel/xfile.h
```diff
@@ -76,6 +76,7 @@ class X_FILE_DIRECTORY_INFORMATION {
 class XFile : public XObject {
  public:
   static const XObject::Type kObjectType = XObject::Type::File;
+  static constexpr uint16_t kMaximalPathLength = 260;
```
Triang3l, line 79: "Maximal" is an uncommon word in coding, usually it's "maximum" or (like we usually use) just "max". Also a comment saying that it includes the null terminator would be good, or even changing this to 259 because length, unlike buffer size, usually doesn't include anything but the character count itself.

Triang3l (changes requested):

src/xenia/kernel/xfile.h
Triang3l: For lengths, size_t is a more suitable type also.

Gliniak: Done

src/xenia/kernel/kernel_state.cc
```diff
@@ -316,6 +317,17 @@ void KernelState::SetExecutableModule(object_ref<UserModule> module) {
...
 
+  // Setup the kernel's ExLoadedImageName field
+  export_entry = processor()->export_resolver()->GetExportByOrdinal(
+      "xboxkrnl.exe", ordinals::ExLoadedImageName);
+
+  if (export_entry) {
+    char* variable_ptr =
+        memory()->TranslateVirtual<char*>(export_entry->variable_ptr);
+    xe::string_util::copy_truncating(variable_ptr, executable_module_->path(),
+                                     XFile::kMaxPathLength);
```
Triang3l, line 328: Not done :)

src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
```diff
@@ -189,6 +190,11 @@ XboxkrnlModule::XboxkrnlModule(Emulator* emulator, KernelState* kernel_state)
                                        ordinals::XexExecutableModuleHandle,
                                        ppXexExecutableModuleHandle);
 
+  uint32_t ppExLoadedImageName =
+      memory_->SystemHeapAlloc(XFile::kMaxPathLength);
```
Triang3l, line 194: And here

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
Triang3l: Not sure if we need to always allocate all the 260 bytes here or just min(length, maxLength) + 1, by the way, the former is safer of course though.

Triang3l (approved):

src/xenia/kernel/xfile.h
```diff
@@ -76,6 +76,7 @@ class X_FILE_DIRECTORY_INFORMATION {
 class XFile : public XObject {
  public:
   static const XObject::Type kObjectType = XObject::Type::File;
+  static constexpr size_t kMaxPathLength = 255;
```
Triang3l, line 79: It's not the global maximum path length, seems to be specific to LoadedImageName until we have evidence to the contrary.

src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
```diff
@@ -189,6 +190,11 @@ XboxkrnlModule::XboxkrnlModule(Emulator* emulator, KernelState* kernel_state)
                                        ordinals::XexExecutableModuleHandle,
                                        ppXexExecutableModuleHandle);
 
+  uint32_t ppExLoadedImageName =
```
Triang3l, line 193: Need a comment about 256 bytes (rather than 260 MAX_PATH) being always allocated by the original kernel, in their entirety, disregarding the actual length.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
```diff
@@ -189,6 +189,14 @@ XboxkrnlModule::XboxkrnlModule(Emulator* emulator, KernelState* kernel_state)
                                        ordinals::XexExecutableModuleHandle,
                                        ppXexExecutableModuleHandle);
 
+  // ExLoadedImageName (char*)
+  // The full path to loaded image/xex including its name.
+  // Used usually in custom dashboards (Aurora)
+  uint32_t ppExLoadedImageName =
+      memory_->SystemHeapAlloc(kExLoadedImageNameSize);
```
gibbed, line 196: Minor nitpick: initialize this to zeros (`std::memset(...)`), I don't think `SystemHeapAlloc` guarantees a zeroed buffer?

gibbed (changes requested):

gibbed (approved):

src/xenia/kernel/xboxkrnl/xboxkrnl_module.cc
```diff
@@ -189,6 +189,14 @@ XboxkrnlModule::XboxkrnlModule(Emulator* emulator, KernelState* kernel_state)
                                        ordinals::XexExecutableModuleHandle,
                                        ppXexExecutableModuleHandle);
 
+  // ExLoadedImageName (char*)
+  // The full path to loaded image/xex including its name.
```
Triang3l, line 193: Probably need a comment confirming that the full kExLoadedImageNameSize (disregarding the actual length) is always allocated in the official kernel.

src/xenia/kernel/kernel_state.cc
```diff
@@ -316,6 +316,18 @@ void KernelState::SetExecutableModule(object_ref<UserModule> module) {
     *variable_ptr = executable_module_->hmodule_ptr();
   }
 
+  // Setup the kernel's ExLoadedImageName field
+  export_entry = processor()->export_resolver()->GetExportByOrdinal(
+      "xboxkrnl.exe", ordinals::ExLoadedImageName);
+
```
Triang3l, line 322: This empty line is not needed — these two parts set up one thing; while this entire function sets up a lot of others.

Triang3l (changes requested):

PR #1661, opened 2020-10-13
Add LaunchPath to log

illusion0001:


Gliniak:
Isn't it a better idea to one log for both of them? and also sanitise path by: ``utfcpp::replace_invalid`` (is it even required here?)

gibbed:
Pretty sure `generic_u8string` does not actually perform any conversion, this would cause junk in the log file on Windows?

illusion0001:
pushed with `path_to_utf8`

folder name `1r ラ κόσμε ࠀ𐀀 ����� ߿￿`

`0000D6F0 LaunchXexPath: F:\1-x360\1r ラ κόσμε ࠀ𐀀 ����� ߿￿\default.xex`

PR #1664, opened 2020-10-25
[Base] Cache improvements

Gliniak:
As we talked on Discord.

I changed default cache partition behaviour to be mounted and also added another instance of device: "cache".
I thought that little simplification will be nice to have, so I created loop that adds every entry. 🐑

### DIFF
--- a/src/xenia/app/xenia_main.cc
+++ b/src/xenia/app/xenia_main.cc
@@ -67,7 +67,7 @@ DEFINE_path(
     "Storage");
 
 DEFINE_bool(mount_scratch, false, "Enable scratch mount", "Storage");
-DEFINE_bool(mount_cache, false, "Enable cache mount", "Storage");
+DEFINE_bool(mount_cache, true, "Enable cache mount", "Storage");
 
 DEFINE_transient_path(target, "",
                       "Specifies the target .xex or .iso to execute.",
--- a/src/xenia/app/xenia_main.cc
+++ b/src/xenia/app/xenia_main.cc
@@ -278,27 +278,22 @@ int xenia_main(const std::vector<std::string>& args) {
   }
 
   if (cvars::mount_cache) {
-    auto cache0_device =
-        std::make_unique<xe::vfs::HostPathDevice>("\\CACHE0", "cache0", false);
-    if (!cache0_device->Initialize()) {
-      XELOGE("Unable to scan cache0 path");
-    } else {
-      if (!emulator->file_system()->RegisterDevice(std::move(cache0_device))) {
-        XELOGE("Unable to register cache0 path");
-      } else {
-        emulator->file_system()->RegisterSymbolicLink("cache0:", "\\CACHE0");
-      }
-    }
+    constexpr std::pair<std::string_view, std::string_view> cache_mounts[] = {
+        {"\\CACHE", "cache"}, {"\\CACHE0", "cache0"}, {"\\CACHE1", "cache1"}};
 
-    auto cache1_device =
-        std::make_unique<xe::vfs::HostPathDevice>("\\CACHE1", "cache1", false);
-    if (!cache1_device->Initialize()) {
-      XELOGE("Unable to scan cache1 path");
-    } else {
-      if (!emulator->file_system()->RegisterDevice(std::move(cache1_device))) {
-        XELOGE("Unable to register cache1 path");
+    for (auto const cache_mount : cache_mounts) {
+      auto cache_device = std::make_unique<xe::vfs::HostPathDevice>(
+          cache_mount.first, cache_mount.second, false);
+
+      if (!cache_device->Initialize()) {
+        XELOGE("Unable to scan {} path", cache_mount.second);
       } else {
-        emulator->file_system()->RegisterSymbolicLink("cache1:", "\\CACHE1");
+        if (!emulator->file_system()->RegisterDevice(std::move(cache_device))) {
+          XELOGE("Unable to register {} path", cache_mount.second);
+        } else {
+          emulator->file_system()->RegisterSymbolicLink(
+              std::string(cache_mount.second) + ':', cache_mount.first);
+        }
       }
     }
   }


src/xenia/app/xenia_main.cc
```diff
@@ -67,7 +67,7 @@ DEFINE_path(
     "Storage");
 
 DEFINE_bool(mount_scratch, false, "Enable scratch mount", "Storage");
-DEFINE_bool(mount_cache, false, "Enable cache mount", "Storage");
+DEFINE_bool(mount_cache, true, "Enable cache mount", "Storage");
```
Triang3l, line 70: This one is kinda ugh because it won't be changed for existing users since they already have the entry in the config :/

src/xenia/app/xenia_main.cc
```diff
@@ -278,27 +278,22 @@ int xenia_main(const std::vector<std::string>& args) {
...
-    if (!cache0_device->Initialize()) {
-      XELOGE("Unable to scan cache0 path");
-    } else {
-      if (!emulator->file_system()->RegisterDevice(std::move(cache0_device))) {
-        XELOGE("Unable to register cache0 path");
-      } else {
-        emulator->file_system()->RegisterSymbolicLink("cache0:", "\\CACHE0");
-      }
-    }
+    const std::pair<std::string, std::string> cache_mounts[] = {
```
Triang3l, line 281: std::string (instead of const char*) is discouraged for literals due to potential dynamic allocation (though not likely to happen in practice here specifically due to small string optimization, but in the general case it's possible).

Triang3l (changes requested):

src/xenia/app/xenia_main.cc
JoelLinn: I would go with `const std::string_view`. The used APIs expect that anyways. Playing with godbolt I found out clang 11 will call `strlen` on a `const char*` even with `-O3`. That is scary...

This isn't hot code though so we shouldnt get too crazy...

src/xenia/app/xenia_main.cc
```diff
@@ -278,27 +278,22 @@ int xenia_main(const std::vector<std::string>& args) {
...
+
+      if (!cache_device->Initialize()) {
+        XELOGE("Unable to scan {} path", cache_mount.second);
       } else {
-        emulator->file_system()->RegisterSymbolicLink("cache1:", "\\CACHE1");
+        if (!emulator->file_system()->RegisterDevice(std::move(cache_device))) {
+          XELOGE("Unable to register {} path", cache_mount.second);
+        } else {
+          emulator->file_system()->RegisterSymbolicLink(
+              cache_mount.second + ':', cache_mount.first);
```
JoelLinn, line 295: This string concatenation needs some buffer anyways...

src/xenia/app/xenia_main.cc
JoelLinn: Do you suggest changing the variable name? imho it is a rather good name right now

Razzile: I wonder if it's worth having a versioning system for cvars? would be a pain to implement though

Triang3l: Thought about something similar — something like `DEFAULT_CHANGED_bool(mount_cache, false, 2020, 10, 26, 13);` with the constructor `max`ing the global current defaults version, and a config variable containing the defaults version at the moment of config writing. Hacky, but possibly more user- and parsing-friendly (as opposed to comments, which would be pain to treat as special defaults comments for sorting purpose), and won't need to wait for the majority of users to upgrade their config to the comments+overrides format before changing defaults safely.

Triang3l: The version would be (in UTC+0 time) `hour + day * 100 + month * 10000 + year * 1000000` — fits in int32.

JoelLinn: why not simply comment out default cvars in the config on writing? If the default has changed on the next load it will automatically apply.

Gliniak: Maybe we should create issue for that and continue there as it isn't exactly related to this PR 🤷

Triang3l: @JoelLinn Oh, yes, since we're overwriting the entire file deleting all user-added comments, it should work fine without sorting issues. There are two issues though:
* Users will have to be told to remove # before the line whenever they change something — we're going to have a lot of "I changed this and it doesn't work" questions.
* If a user, for instance, has last used Xenia in September, we add the comment system in October, change some default in November, then a user updating Xenia in December will not get the new default.

gibbed: @Triang3l That's actually easier than you might think.

We could add a new general config version cvar, which Xenia checks on startup, if it's not present or has a value less than we expect, we can prompt the user that their config file may have issues due to being out of date.

I can agree with commenting out values that match the default.

Triang3l: That's what I had in mind, though instead of some centralized version number location, per-cvar history, and automatic upgrading if the value matches a previous default, rather than a warning that will make users have to drop their entire config.

gibbed: I'd be down for that as well, would just need to be in a form that's user-friendly.

Triang3l: The needed functionality has been added in 37013ee352da93e8e2374bd252d3c8a27ba71a1a — after defining `mount_cache`, add:
`UPDATE_from_bool(mount_cache, year, month, day, hour, false);`
and update the `IConfigVarUpdate::kLastCommittedUpdateDate` to the same date in `base/cvar.h`.

src/xenia/app/xenia_main.cc
```diff
@@ -278,27 +278,22 @@ int xenia_main(const std::vector<std::string>& args) {
...
 
-    auto cache1_device =
-        std::make_unique<xe::vfs::HostPathDevice>("\\CACHE1", "cache1", false);
-    if (!cache1_device->Initialize()) {
-      XELOGE("Unable to scan cache1 path");
-    } else {
-      if (!emulator->file_system()->RegisterDevice(std::move(cache1_device))) {
-        XELOGE("Unable to register cache1 path");
+    for (auto const cache_mount : cache_mounts) {
+      auto cache_device = std::make_unique<xe::vfs::HostPathDevice>(
```
Triang3l, line 285: A special cache directory (`cache_root`) has recently been added — please use it instead so it can be cleared from the system settings on Android in the future (something like `{cache_root}/guest_mounts/{title ID}/…`, though a more nice subdirectory name may be chosen, it's up for discussion).

Triang3l (changes requested):

src/xenia/app/xenia_main.cc
emoose: Making cache per-titleID is a good idea, on actual 360 it clears cache between different titles IIRC (and maybe media IDs too), and some games can do weird things if cache has data it doesn't expect (eg. Halo 3 delta has issues loading if you use cache from even a different build-type (debug/release/etc) than the one you're using, titleid wouldn't help here though, but maybe also using mediaid would?)

I guess cache-mounting code would need to be moved from xenia_main() for this though, that'd probably be helpful anyway, ran into problems with xenia only creating the cache folders on startup instead of when a game is actually loaded too many times now...

gibbed: If it's that sensitive we may want to generate a hash of the executable's code for uniqueness then.

emoose: Yeah that'd probably be best, just checked the halo 3 delta XEXs and none of them seem to use media IDs at all unfortunately. Maybe there's a hash in the XEX headers we could use instead of needing to generate it ourselves.

Gliniak: > Making cache per-titleID is a good idea, on actual 360 it clears cache between different titles IIRC

If it is cleared everytime you run different app, then maybe we should just clear it while xenia is opened? 🤔 It's only a game cache tbh and it shouldn't (?) have performance impact.

Triang3l: If we make an implementation where the cache only lives for as long as the game is running, maybe a RAM-based virtual file system could also be a good option (along with a true directory on the disk).

Gliniak:
Superseded by #1867

Issue #1296, opened 2018-12-01
Completing XOVERLAPPED before returning from export causes issues in some games


emoose:
Right now all kinds of exports that take XOVERLAPPED params are completing them before returning from the export, this works fine in most cases but unfortunately seems to cause issues in others, where I guess maybe the game isn't expecting them to be immediately completed and ends up not running correctly as a result.

Some of the worst examples are probably the XamShow*UI functions, in those we actually wait for user input before returning from the export at all, meaning we could be stalling the game for possibly minutes/hours before we return control to it. (while the stalling is maybe a separate issue, it can probably be fixed with the same solution to the overlapped problem)

An easy way I found to remedy this is by putting the main code (+ the overlapped completion code) of the export into a separate thread (or if the main code is pretty much non-time-consuming, having a small thread sleep just before completion), that way the overlapped is still in progress when the game regains execution, and the export also takes a fraction of the time it previously took to run too. (changes for this are in my branch here: https://github.com/emoose/xenia/commits/xamshowui-fixes, though in XamShowDeviceSelectorUI I use an imgui dialog to simulate a thread sleep since I didn't know how to actually make the thread sleep at the time :P)

This solved issues in the dash.xex during initial setup where it wouldn't recognise input from XamShowKeyboardUI, also fixes Sonic Unleashed not recognising the device id returned from XamShowDeviceSelectorUI, and Project Sylpheed also now goes in-game with the fix, there's probably many other games where this helps with things too.

I'm not sure if creating new threads for every export that uses XOVERLAPPED params is really the best idea though, I could see threads quickly piling up, maybe even ending up in contention with each other...
IIRC 360's Xam uses a task-queue system instead, where tasks are done sequentially rather than my current solution of starting a new thread immediately.
Something like that would probably be a lot better, and bring us closer to how the 360 actually works, though I'm not really sure where to start with something like that...

illusion98:
This fix also gets Warriors Orochi 3 in-game
![](https://cdn.discordapp.com/attachments/308207592482668545/595179441378820097/Untitled.png)

emoose:
Saw some talk on discord about Project Sylpheed, so bumping this issue since AFAIK this is what  causes it to freeze with a "Select storage device" window ingame.

My threaded-xoverlapped branch seems to fix this, you can see the changed code for XamShowDeviceSelectorUI here:
[...]

PR #1670, opened 2020-11-05
Xam content / Storage fixes

randprint:
These couple of fixes fix up all the rest of the storage detection issues I can find with the caveat of XamUIDeviceSelector needing some solution to set its pointer after a delay

The kernel commit sets the same flag to XboxHardwareInfo that the kernel does at startup after storage is initialized, found disassembling the kernel trying to debug why these last few problems with storage being detected persisted.

The commit to xam_content changes the size of the storage being reported as a hdd device type to the smallest size hdd that shipped with the 360, this fixed saving in Rapala fishing.

The changes to byteswap the deviceinfo struct in XamContentGetDeviceState fix a large number of storage detection issues in many games, with the caveat that a solution is needed in Xam_UI to XamShowDeviceSelectorUI as discussed and detailed here in issue https://github.com/xenia-project/xenia/issues/1296

### DIFF
--- a/src/xenia/kernel/xam/xam_content.cc
+++ b/src/xenia/kernel/xam/xam_content.cc
@@ -46,14 +46,12 @@ struct DeviceInfo {
 // they incorrectly only look at the lower 32-bits of free_bytes,
 // when it is a 64-bit value. Which means any size above ~4GB
 // will not be recognized properly.
-//
-// NOTE(randprint): you can use 120 GB and 42 GB 'fullness'
-// with the proper deviceID feel free to change at your discression
 #define ONE_GB (1024ull * 1024ull * 1024ull)
 static const DeviceInfo dummy_device_info_ = {
-    0x00000001,    1,  // found from debugging / reversing UE3 engine titles
-    4ull * ONE_GB,     // 4GB
-    3ull * ONE_GB,     // 3GB, so it looks a little used.
+    0x00000001,      // id
+    1,               // 1=HDD
+    20ull * ONE_GB,  // 20GB
+    3ull * ONE_GB,   // 3GB, so it looks a little used.
     u"Dummy HDD",
 };
 #undef ONE_GB
@@ -118,7 +116,7 @@ DECLARE_XAM_EXPORT1(XamContentGetDeviceState, kContent, kStub);
 
 typedef struct {
   xe::be<uint32_t> device_id;
-  xe::be<uint32_t> unknown;
+  xe::be<uint32_t> device_type;
   xe::be<uint64_t> total_bytes;
   xe::be<uint64_t> free_bytes;
   xe::be<uint16_t> name[28];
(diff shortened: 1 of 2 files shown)


src/xenia/kernel/xam/xam_content.cc
```diff
@@ -133,10 +133,11 @@ dword_result_t XamContentGetDeviceData(
 
   device_data.Zero();
   const auto& device_info = dummy_device_info_;
-  device_data->device_id = device_info.device_id;
-  device_data->unknown = device_id & 0xFFFF;  // Fake it.
-  device_data->total_bytes = device_info.total_bytes;
-  device_data->free_bytes = device_info.free_bytes;
+  // byteswap the device info
+  device_data->device_id = xe::byte_swap(device_info.device_id);
```
gibbed, line 137: This shouldn't be necessary, `X_CONTENT_DEVICE_DATA` has all of its fields wrapped in `xe::be`?

src/xenia/kernel/xam/xam_content.cc
```diff
@@ -46,13 +46,13 @@ struct DeviceInfo {
...
-// NOTE(randprint): you can use 120 GB and 42 GB 'fullness'
-// with the proper deviceID feel free to change at your discression
+// Rapala fishing only detected the hdd if the size was
+// raised to 20GB (the smallest xbox hdd)
 #define ONE_GB (1024ull * 1024ull * 1024ull)
 static const DeviceInfo dummy_device_info_ = {
     0x00000001,    1,  // found from debugging / reversing UE3 engine titles
-    4ull * ONE_GB,     // 4GB
-    3ull * ONE_GB,     // 3GB, so it looks a little used.
+    20ull * ONE_GB,     // 4GB
```
gibbed, line 54: Comments need updating.

gibbed (changes requested):

gibbed:
Additionally, please fix your commit messages. Short message should be max 50 characters. Your XAM commit message is a bit messy (and a stray -).

src/xenia/kernel/xam/xam_content.cc
randprint: testing proved otherwise? I removed the obscene amount of debug prints that led me to find this. We're swapping it in XamContentCreateDeviceEnumerator and games that used that seemed to be working, I followed suit when I noticed this function was failing to work properly and voilà.

gibbed: The code that you're referring to is copying data by pointers to struct fields, not simply assigning swapped values to the struct. Thus the swaps.

https://github.com/xenia-project/xenia/blob/4f5c313fe39350c7abd86150bf0558e23b5d98d1/src/xenia/kernel/xam/xam_content.cc#L222-L226

gibbed: I'm not saying your change is a bad one, just that the byte swap here seems out of place / wrong.

randprint: Actually, I think it is wrong to byteswap here you were correct to doubt it, I made a false correlation here and didn't test enough.

PR #1672, opened 2020-11-07
[Base] Patching System

Gliniak:
This is initial version of xex patching for Xenia.

It still requires some work, maybe a little bit of simplification and cleanup.

I created it as different project for future support for patching xex externally.

# Missing functionalities:
- CRC check or patching only when pattern matches
- Some future GUI functions?
- Saving changes to patch file directly from xenia (also for future GUI usage)

# Usage
1. There must be directory patches where xenia.exe is located (might be changed to Documents patch)
2. Patches for certain title **MUST** be  in file called <title_id>.patch (Example: 535107E4.patch)
3. To activate certain patch it must be enabled in .patch file and patching system must be enabled in xenia config
4. If title have any patch applied then in titlebar will be note ``[Patches Applied]``

Example patch:
```
title_name = "Gears Of War"
title_id = "4D5307D5"

[[patch]]
    name = "GoW - Texture rendering"
    desc = "None"
    author = "Gliniak"
    is_enabled = true
	
    [[patch.be32]]
        address = 0x822306D8
        value = 0x409AFF6C
```

# Concerns
- Having requirements of <title_id>.patch as patch filename might be problematic if there will be a lot of patches available. Title_id should be checked inside patch file (like in example) independently of its name. This will increase future readability and management.


src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,213 @@
...
+namespace patcher {
+
+PatchingSystem::PatchingSystem() {
+  isAnyPatchApplied_ = false;
+  loadPatches();
+}
+
+PatchingSystem::~PatchingSystem() {}
+
+void PatchingSystem::loadPatches() {
```
Triang3l, line 29: UpperCase is used for method and class names (except for snake_case getters).

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,213 @@
...
+#include "xenia/base/filesystem.h"
+#include "xenia/base/logging.h"
+
+DEFINE_bool(apply_patches, true, "Enables patching functionality", "General");
+
+namespace xe {
+namespace patcher {
+
+PatchingSystem::PatchingSystem() {
+  isAnyPatchApplied_ = false;
```
Triang3l, line 23: snake_case_ is used for variables.

Triang3l (changes requested):

src/xenia/app/emulator_window.cc
```diff
@@ -454,6 +454,12 @@ void EmulatorWindow::UpdateTitle() {
         ")";
   }
 
+  auto patching_system = emulator()->patching_system();
```
Triang3l, line 457: auto should generally be avoided because it makes code harder to read — only for cases when you already have a cast on the right side, or for things like iterators.

src/xenia/emulator.cc
```diff
@@ -196,6 +196,8 @@ X_STATUS Emulator::Setup(
   // Bring up the virtual filesystem used by the kernel.
   file_system_ = std::make_unique<xe::vfs::VirtualFileSystem>();
 
+  patching_system_ = std::make_unique<xe::patcher::PatchingSystem>();
```
Triang3l, line 199: The word "system" is redundant, as most large classes are systems in general.

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,213 @@
+/**
+ ******************************************************************************
+ * Xenia : Xbox 360 Emulator Research Project                                 *
+ ******************************************************************************
+ * Copyright 2016 Ben Vanik. All rights reserved.                             *
+ * Released under the BSD license - see LICENSE in the root for more details. *
+ ******************************************************************************
+ */
+#include <regex>
```
Triang3l, line 9: Standard library includes are between the class header and other Xenia includes.

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,213 @@
...
+
+// TODO(Gliniak): Somehow resolve this mess. Maybe by template?
+uint8_t PatchingSystem::getAllocSize(const std::string provided_size) {
+  uint8_t alloc_size = sizeof(uint32_t);
+
+  if (provided_size == "be64") {
+    alloc_size = sizeof(uint64_t);
+  }
+
+  if (provided_size == "be32") {
```
Triang3l, line 198: `else if`

src/xenia/patcher/patcher.h
```diff
@@ -0,0 +1,80 @@
...
+  void loadPatches();
+  void applyPatch(Memory* memory, const patchInfoEntry* patch);
+  void applyPatchesForTitle(Memory* memory, const uint32_t title_id);
+
+  patchFileEntry readPatchFile(const std::filesystem::path& file_path);
+  std::vector<patchDataEntry>* readPatchData(
+      const std::string size_type,
+      const std::shared_ptr<cpptoml::table>& patch_table);
+
+  std::vector<patchFileEntry> getAllPatches() { return loaded_patch_files; }
```
Triang3l, line 66: This would create a copy of the vector — return a reference, `std::vector<PatchFileEntry>&`, instead.

src/xenia/patcher/patcher.h
```diff
@@ -0,0 +1,80 @@
...
+  std::vector<patchFileEntry> loaded_patch_files;
+
+  bool isAnyPatchApplied_;
+
+  uint8_t getAllocSize(const std::string provided_size);
+};
+
+}  // namespace patcher
+}  // namespace xe
+#endif
```
Triang3l, line 80: No new line at the end.

Triang3l (changes requested):

Gliniak:
Updated:
- Resolved most of review notes
- Removed dumb requirement of having `<title_id>.patch` filename template. Now you can use `<title_id><whatever>.patch`
(example: `4D5307D5.patch` can be named: `4D5307D5 - Gears Of War.patch`)

src/xenia/patcher/patcher.h
```diff
@@ -0,0 +1,81 @@
...
+#include <map>
+#include <regex>
+#include <vector>
+
+#include "third_party/cpptoml/include/cpptoml.h"
+#include "third_party/xxhash/xxhash.h"
+
+#include "xenia/memory.h"
+
+struct patchDataEntry {
```
Triang3l, line 22: Apart from incorrect case of the first letter, this should be in the xe::patcher namespace too.

Triang3l (changes requested):

Gliniak:
Updated:
 - Resolved review notes
 - Removed many "auto" usages
 - Added XXH3 Hash calculation (patch file might include generated hash, but it also might be skipped)

src/xenia/patcher/premake5.lua
```diff
@@ -0,0 +1,14 @@
...
+project("xenia-patcher")
+  uuid("e1c75f76-9e7b-48f6-b17e-dbd20f7a1592")
+  kind("StaticLib")
+  language("C++")
+  links({
+    "xenia-base"
+  })
+  defines({
+  })
+  recursive_platform_files()
```
Margen67, line 14: end of file newline

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+
+PatchingSystem::~PatchingSystem() {}
+
+void PatchingSystem::LoadPatches() {
+  if (!cvars::apply_patches) {
+    return;
+  }
+
+  const std::filesystem::path patches_directory =
+      filesystem::GetExecutableFolder() / "patches";
```
Gliniak, line 33: Should we leave it as it is or maybe put it inside ``Documents\Xenia\Patches``?

src/xenia/kernel/user_module.cc
```diff
@@ -102,6 +115,8 @@ X_STATUS UserModule::LoadFromFile(const std::string_view path) {
     return result;
   }
 
+  XELOGI("Module hash: {:08X} for {}", hash_, name_);
```
gibbed, line 118: `hash_` is a 64-bit value (uint64_t), so this should be `{:016X}`.

src/xenia/kernel/user_module.h
```diff
@@ -38,6 +38,7 @@ class UserModule : public XModule {
 
   const std::string& path() const override { return path_; }
   const std::string& name() const override { return name_; }
+  const uint64_t hash() const { return hash_; }
```
gibbed, line 41: The return value doesn't need `const`-ness, just the method.

`uint64_t hash() const { return hash_; }`

src/xenia/emulator.cc
gibbed: I would agree with Triang3l that using system in the name is a bit unclear, perhaps something like `PatchDB` would make more sense?

src/xenia/kernel/user_module.h
```diff
@@ -100,6 +101,7 @@ class UserModule : public XModule {
 
   std::string name_;
   std::string path_;
+  uint64_t hash_;
```
gibbed, line 104: Should initialize this to `0` or some other way to indicate that the value is invalid.

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+      filesystem::ListFiles(patches_directory);
+  std::regex file_name_regex_match = std::regex("^[A-Fa-f0-9]{8}.*\\.patch$");
+
+  for (const xe::filesystem::FileInfo& patch_file : patch_files) {
+    // Skip files that doesn't have only title_id as name and .patch as
+    // extension
+    if (!std::regex_match(patch_file.name.u8string(), file_name_regex_match)) {
+      XELOGE(
+          "PatchingSystem: Skipped loading file {} due to incorrect filename",
+          patch_file.name.u8string());
```
gibbed, line 45: `to_utf8(...)`

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+
+    if (!is_any_patch_applied_) {
+      is_any_patch_applied_ = true;
+    }
+  }
+}
+
+PatchFileEntry PatchingSystem::ReadPatchFile(
+    const std::filesystem::path& file_path) {
+  PatchFileEntry patchFile;
```
gibbed, line 125: Inconsistent variable naming. `file_entry`?

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+  std::shared_ptr<cpptoml::table> patch_toml_fields;
```
gibbed, line 126: Variable naming. `root_table` or `file_table`?

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+
+  auto title_name = patch_toml_fields->get_as<std::string>("title_name");
+  auto title_id = patch_toml_fields->get_as<std::string>("title_id");
+  auto title_hash = patch_toml_fields->get_as<std::string>("hash");
+
+  patchFile.title_id = strtoul((*title_id).c_str(), NULL, 16);
+  patchFile.hash = strtoull((*title_hash).c_str(), NULL, 16);
+  patchFile.title_name = *title_name;
+
+  auto tarr = patch_toml_fields->get_table_array("patch");
```
gibbed, line 144: Bad variable name. `patch_table_array`?

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+
+  for (auto table : *tarr) {
```
gibbed, line 146: If going with other proposed variable name changes. `table` should probably be `patch_table`?

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+    patch.patch_name = patch_name;
+    patch.patch_desc = patch_desc;
+    patch.patch_author = patch_author;
+    patch.is_enabled = is_enabled;
+
+    const std::string data_types[] = {"be64", "be32", "be16", "be8"};
+
+    // Iterate through all available data sizes
+    for (const std::string& type : data_types) {
+      auto entries = ReadPatchData(type, table);
```
gibbed, line 163: `patch_entries`

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+    const std::string size_type,
+    const std::shared_ptr<cpptoml::table>& patch_table) {
+  std::vector<PatchDataEntry>* patch_data = new std::vector<PatchDataEntry>();
+  auto patch_data_tarr = patch_table->get_table_array(size_type);
+
+  if (!patch_data_tarr) {
+    return patch_data;
+  }
+
+  for (const auto& patch_data_table : *patch_data_tarr) {
```
gibbed, line 184: It'd probably be better if this loop were in the caller of this function and this function explicitly only cared about parsing individual patch entries.

It'd probably be OK if there was another function that did the handling of iterating the `data_types` array, I just think the iteration should be the responsibility of the outer scope.

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+    auto value = patch_data_table->get_as<std::uint64_t>("value");
+
+    PatchDataEntry patchData = {GetAllocSize(size_type), *address, *value};
+    patch_data->push_back(patchData);
+  }
+  return patch_data;
+}
+
+uint8_t PatchingSystem::GetAllocSize(const std::string provided_size) {
+  uint8_t alloc_size = sizeof(uint32_t);
```
gibbed, line 196: Needs better error handling, defaulting to `uint32_t` if an unknown size string is passed in is probably a bad idea.

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,217 @@
...
+}
+
+void PatchingSystem::ApplyPatchesForTitle(Memory* memory,
+                                          const uint32_t title_id,
+                                          const uint64_t hash) {
+  for (const PatchFileEntry& patchFile : loaded_patch_files) {
+    if (patchFile.title_id != title_id) {
+      continue;
+    }
+    if (patchFile.hash != 0 && patchFile.hash != hash) {
```
gibbed, line 66: Is it actually impossible for there to be a XXH3_64bits hash of `0`?

gibbed (changes requested):

src/xenia/patcher/patcher.cc
gibbed: This should be relative to user configuration in some fashion.

Triang3l: Yes, relative to the storage root in our terms, which is the "home" directory of Xenia.

Triang3l: Though maybe eventually we should support multiple search paths for patches.

src/xenia/patcher/patch_db.cc
```diff
@@ -0,0 +1,164 @@
...
+    XELOGW("PatchDB: Unsupported patch type!");
+    return alloc_size;
+  }
+
+  alloc_size = uint8_t(itr->second);
+  return alloc_size;
+}
+
+}  // namespace patcher
+}  // namespace xe
```
Margen67, line 164: eof newline

src/xenia/patcher/patch_db.h
```diff
@@ -0,0 +1,78 @@
...
+  const std::map<std::string, size_t> patch_data_types_size = {
+      {"f64", sizeof(uint64_t)},  {"f32", sizeof(uint32_t)},
+      {"be64", sizeof(uint64_t)}, {"be32", sizeof(uint32_t)},
+      {"be16", sizeof(uint16_t)}, {"be8", sizeof(uint8_t)}};
+};
+
+}  // namespace patcher
+}  // namespace xe
+
+#endif  // XENIA_PATCH_LOADER_H_
```
Margen67, line 78: eof newline

Margen67:
> Updated:
> 
>     * Resolved most of review notes
> 
>     * Removed dumb requirement of having `<title_id>.patch` filename template. Now you can use `<title_id><whatever>.patch`
>       (example: `4D5307D5.patch` can be named: `4D5307D5 - Gears Of War.patch`)

This should be `.patch.toml` to be consistent with the config.

src/xenia/patcher/patch_db.cc
```diff
@@ -0,0 +1,164 @@
...
+    uint32_t address = *patch_data_table->get_as<std::uint32_t>("address");
+    uint64_t value = *patch_data_table->get_as<std::uint64_t>("value");
+
+    if (size_type == "f32") {
+      float fvalue = float(*patch_data_table->get_as<double>("value"));
+      value = *(reinterpret_cast<uint32_t*>(&fvalue));
+    } else if (size_type == "f64") {
+      double dvalue = *patch_data_table->get_as<double>("value");
+      value = *(reinterpret_cast<uint64_t*>(&dvalue));
+    }
```
Margen67, line 128: Maybe this should be a switch case?

Gliniak: Sure, but I would need to add some string->enum map while I load each patch

src/xenia/patcher/patch_db.cc
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  const std::filesystem::path patches_directory = patches_root_ / "patches";
+  const std::vector<xe::filesystem::FileInfo>& patch_files =
+      filesystem::ListFiles(patches_directory);
+  const std::regex file_name_regex_match =
+      std::regex("^[A-Fa-f0-9]{8}.*\\.patch$");
+
+  for (const xe::filesystem::FileInfo& patch_file : patch_files) {
+    // Skip files that doesn't have only title_id as name and .patch as
```
Margen67, line 42: Not sure if this is right...
```suggestion
      std::regex("^[A-F0-9]{8}.*\\.patch\\.toml$");

  for (const xe::filesystem::FileInfo& patch_file : patch_files) {
    // Skip files that doesn't have only title_id as name and .patch.toml as
```

Margen67:
@Gliniak Title updates applied via xexp don't have the correct (updated) hash.

src/xenia/gpu/d3d12/premake5.lua
```diff
@@ -39,6 +39,7 @@ project("xenia-gpu-d3d12-trace-viewer")
     "xenia-ui",
     "xenia-ui-d3d12",
     "xenia-vfs",
+    "xenia-patcher",
```
Triang3l, line 42: For easier maintenance, these lists should be sorted alphabetically unless the order matters (if it does, a comment explaining that is needed).

Triang3l (changes requested):

src/xenia/patcher/patch_db.cc
```diff
@@ -0,0 +1,223 @@
...
+
+PatchDB::~PatchDB() {}
+
+void PatchDB::LoadPatches() {
+  if (!cvars::apply_patches) {
+    return;
+  }
+
+  const std::filesystem::path patches_directory = patches_root_ / "patches";
+  const std::vector<xe::filesystem::FileInfo>& patch_files =
```
Triang3l, line 35: The usage of a reference here is confusing, will this vector object exist after the assignment, and also if ListFiles is called again afterwards?

Triang3l (changes requested):

src/xenia/patcher/patch_db.cc
```diff
@@ -0,0 +1,223 @@
...
+                                    hash) != entry.hashes.cend();
+
+        return entry.title_id == title_id &&
+               (entry.hashes.empty() || hash_exist);
+      });
+
+  return title_patches;
+}
+
+void PatchDB::ReadHash(PatchFileEntry& patchEntry,
```
Triang3l, line 208: `patch_entry` *

src/xenia/kernel/user_module.h
```diff
@@ -97,9 +98,11 @@ class UserModule : public XModule {
 
  private:
   X_STATUS LoadXexContinue();
+  void CalculateHash();
 
   std::string name_;
   std::string path_;
+  uint64_t hash_ = -1;
```
Triang3l, line 105: `UINT64_MAX` or `~UINT64_C(0)` as this is unsigned. Though neither this nor even 0 is an invalid hash value — if such indication is needed, `std::optional` should be preferred.

src/xenia/kernel/user_module.cc
```diff
@@ -804,5 +808,39 @@ void UserModule::Dump() {
...
+          from_bottom ? i : security_info->page_descriptor_count - i;
+      xex2_page_descriptor page_descriptor;
+      page_descriptor.value =
+          xe::byte_swap(security_info->page_descriptors[page_index].value);
+      if (page_descriptor.info != XEX_SECTION_CODE) {
+        continue;
+      }
+      return page_index;
+    }
+    return (uint32_t)0;
```
Triang3l, line 829: The cast is not needed here. However, is 0 an invalid index (unlike something like `UINT32_MAX`) actually?

src/xenia/kernel/user_module.cc
```diff
@@ -804,5 +808,39 @@ void UserModule::Dump() {
   xe::logging::AppendLogLine(xe::LogLevel::Info, 'i', sb.to_string_view());
 }
 
+void UserModule::CalculateHash() {
+  const uint32_t page_size =
```
Triang3l, line 812: This should preferably be done via page size querying in `Memory`.

src/xenia/patcher/patch_db.h
```diff
@@ -0,0 +1,131 @@
...
+
+struct PatchDataValue {
+  const size_t alloc_size_;
+  std::vector<uint8_t> patch_data_;
+
+  template <typename T>
+  PatchDataValue(const size_t alloc_size, const T value)
+      : alloc_size_(alloc_size) {
+    patch_data_.resize(alloc_size_);
+    memcpy(patch_data_.data(), &value, alloc_size_);
```
Triang3l, line 29: `std::memcpy`, also `#include <cstring>` is needed.

src/xenia/patcher/patch_db.h
```diff
@@ -0,0 +1,131 @@
...
+
+struct PatchFileEntry {
+  uint32_t title_id;
+  std::string title_name;
+  std::vector<uint64_t> hashes;
+  std::vector<PatchInfoEntry> patch_info;
+};
+
+enum class PatchDataType {
+  be8,
```
Triang3l, line 73: kBE8…kU16String.

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,67 @@
...
+
+#include "xenia/base/logging.h"
+
+namespace xe {
+namespace patcher {
+
+Patcher::Patcher(const std::filesystem::path patches_root) {
+  is_any_patch_applied_ = false;
+  patch_db = new PatchDB(patches_root);
+}
```
Triang3l, line 19: `patch_db_` (with the underscore) *

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,67 @@
...
+
+Patcher::~Patcher() {}
```
Triang3l, line 21: Unless there are issues with unique_ptr deletion, it should be either not declared at all, or defined as = default (though = default may be needed in the .cc rather than .h in some unique_ptr usage cases, it's somewhat confusing).

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,67 @@
...
+             patchFile.title_name, patchFile.title_id, patchEntry.patch_name);
+      ApplyPatch(memory, &patchEntry);
+    }
+  }
+}
+
+void Patcher::ApplyPatch(Memory* memory, const PatchInfoEntry* patch) {
+  for (const PatchDataEntry& patch_data_entry : patch->patch_data) {
+    uint32_t old_address_protect = 0;
+    auto address = memory->TranslateVirtual(patch_data_entry.memory_address_);
```
Triang3l, line 42: `auto` unless the type is obvious from the names in the assignment (or from an explicit cast) complicates reading the code.

src/xenia/patcher/patcher.cc
```diff
@@ -0,0 +1,67 @@
...
+      continue;
+    }
+
+    heap->QueryProtect(patch_data_entry.memory_address_, &old_address_protect);
+
+    heap->Protect(patch_data_entry.memory_address_,
+                  (uint32_t)patch_data_entry.new_data_.alloc_size_,
+                  kMemoryProtectRead | kMemoryProtectWrite);
+
+    memcpy(address, patch_data_entry.new_data_.patch_data_.data(),
```
Triang3l, line 54: `std::memcpy` and `#include <cstring>`.

Triang3l (changes requested):

Margen67:
`title_id` should be removed since the title ID is part of the file name now.

PR #1673, opened 2020-11-10
[GPU] Added Stub for Packet: PM4_WAIT_FOR_IDLE

Gliniak:


src/xenia/gpu/command_processor.cc
```diff
@@ -734,6 +734,13 @@ bool CommandProcessor::ExecutePacketType3(RingBuffer* reader, uint32_t packet) {
       result = true;
       break;
     }
+    case PM4_WAIT_FOR_IDLE: {
```
Triang3l, line 737: Would be useful to have a comment about its appearance in Duke Nukem Forever and steps to trigger it in the game.

PR #1674, opened 2020-11-11
[VFS] Reset reading index while no entry was found

Gliniak:
Copy of my comment from dev channel:

> It uses NtQueryDirectoryFile few times. On the first time it provides not existing file name what sets find_index_ to amount of files. 
> Then second execution is without filename and without restart flag what causes it to autofail due to not reseting find_index_ to 0 in previous fail.


Gliniak:
I'm closing this PR. This solution is incorrect.

PR #1676, opened 2020-11-14
[Base/Kernel] Add and use truncating null-terminating string copying

Triang3l:
This supersedes and more correctly includes #1632. Ensures null termination of strings written by certain kernel functions even in case of buffer overflow. I'm not sure about the designed overflow behavior of the kernel functions when I'm using the new copying, but overall this seems safe, also fixes the issue found by @Gliniak in #1632.

### DIFF
--- a/src/xenia/base/string_util.h
+++ b/src/xenia/base/string_util.h
@@ -10,11 +10,15 @@
 #ifndef XENIA_BASE_STRING_UTIL_H_
 #define XENIA_BASE_STRING_UTIL_H_
 
+#include <algorithm>
 #include <charconv>
+#include <cstddef>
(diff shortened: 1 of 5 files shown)


src/xenia/base/string_util.h
```diff
@@ -30,6 +34,40 @@
 namespace xe {
 namespace string_util {
 
+inline size_t copy_truncating(char* dest, std::string_view source,
```
gibbed, line 37: Should this not be `const std::string_view`? (and following instances here too)

Triang3l: Didn't notice that, fixed!

PR #1690, opened 2020-11-20
[GPU] Viz query stuff

JoelLinn:
Before this viz queries (GPU side conditional drawing based on the visibility of test geometry rendered earlier) where ignored. Test geometry was rendered like any other draw call and all following conditional draw calls where rendered anyways.

With these changes it's still not implemented (conditional draws are still rendered unconditionally) but some workarounds have been added.

- Test geometry is skipped when the corresponding register discard flag is set. (though a final implementation would still need to render these but invisibly)
- The status registers that describe the result of the finished viz queries are now set to 1 (visible) in case they get read back.
- Added perfcounter register defines.

It makes the pink boxes over wheels in the Forza Horizon 2 menu go away.


src/xenia/gpu/command_processor.cc
```diff
@@ -1218,6 +1220,12 @@ bool CommandProcessor::ExecutePacketType3_DRAW_INDX(RingBuffer* reader,
     } break;
   }
 
+  auto viz_query = register_file_->Get<reg::PA_SC_VIZ_QUERY>();
+  if (viz_query.viz_query_ena && viz_query.kill_pix_post_hi_z) {
+    // TODO Handle this properly in the render backends.
```
Triang3l, line 1225: // TODO(JoelLinn): or // TODO(Triang3l):

Triang3l (changes requested):

src/xenia/gpu/registers.h
```diff
@@ -289,17 +289,35 @@ union PA_SC_MPASS_PS_CNTL {
...
+    uint32_t kill_pix_post_hi_z : 1;  // +7 discard geometry after test (but use
+                                      // for testing)
+    uint32_t kill_pix_detail_mask : 1;  // +8 not used with d3d
   };
   uint32_t value;
   static constexpr Register register_index = XE_GPU_REG_PA_SC_VIZ_QUERY;
 };
 
+// Status flags of viz queries, doesn't seem to be read back by d3d
+union PA_SC_VIZ_QUERY_STATUS_0 {
```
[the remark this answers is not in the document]
Triang3l: For comments regarding simple 32-bit values or bit arrays of 32 booleans, a comment in register_table.inc is enough.

src/xenia/gpu/command_processor.cc
```diff
@@ -1218,6 +1220,12 @@ bool CommandProcessor::ExecutePacketType3_DRAW_INDX(RingBuffer* reader,
     } break;
   }
 
+  auto viz_query = register_file_->Get<reg::PA_SC_VIZ_QUERY>();
+  if (viz_query.viz_query_ena && viz_query.kill_pix_post_hi_z) {
+    // TODO(Triang3l || JoelLinn) Handle this properly in the render backends.
+    return true;
```
Triang3l, line 1226: We can't drop draw calls this way if there's vertex shader memexport in them. Though for now that's the only edge case that needs to be handled — it's possible that depth/stencil counters may be affected somehow, though that would be a weird situation, and we don't support occlusion queries yet.

src/xenia/gpu/command_processor.cc
```diff
@@ -1179,9 +1179,11 @@ bool CommandProcessor::ExecutePacketType3_DRAW_INDX(RingBuffer* reader,
   // initiate fetch of index buffer and draw
   // if dword0 != 0, this is a conditional draw based on viz query.
   // This ID matches the one issued in PM4_VIZ_QUERY
-  // ID = dword0 & 0x3F;
-  // use = dword0 & 0x40;
   uint32_t dword0 = reader->ReadAndSwap<uint32_t>();  // viz query info
+  uint32_t viz_id = dword0 & 0x3F;
```
Triang3l, line 1183: These will be unreferenced local variables and trigger warnings on some compilers.

src/xenia/gpu/registers.h
```diff
@@ -289,12 +289,17 @@ union PA_SC_MPASS_PS_CNTL {
...
-// Scanline converter viz query
+// Scanline converter viz query, used by D3D for gpu side conditional rendering
 union PA_SC_VIZ_QUERY {
   struct {
-    uint32_t viz_query_ena : 1;          // +0
-    uint32_t viz_query_id : 6;           // +1
-    uint32_t kill_pix_post_early_z : 1;  // +7
+    // +0 the visibility of draws should be evaluated
+    uint32_t viz_query_ena : 1;
+    // +1
```
Triang3l, line 297: I meant just usage comments on separate lines, but offsets on the same lines as fields.

src/xenia/gpu/command_processor.cc
```diff
@@ -1218,6 +1220,12 @@ bool CommandProcessor::ExecutePacketType3_DRAW_INDX(RingBuffer* reader,
     } break;
   }
 
+  auto viz_query = register_file_->Get<reg::PA_SC_VIZ_QUERY>();
+  if (viz_query.viz_query_ena && viz_query.kill_pix_post_hi_z) {
+    // TODO(Triang3l || JoelLinn) Handle this properly in the render backends.
```
Triang3l, line 1225: Small nitpick again since other changes are requested — we have : after TODO(name).

Triang3l (changes requested):

PR #1698, opened 2020-11-23
[Build] Add stub command for source file generation

Razzile:
Usage:
`xb stub --class dir/class_name`
creates class_name{.h,.cc} at `src/xenia/dir/`

`xb stub --file dir/class.cc`
creates class.cc at `src/xenia/dir`

premake is automatically ran on successful file creation 

I am not very good at python so this can probably be improved and cleaned up

### DIFF
--- a/xenia-build
+++ b/xenia-build
@@ -7,7 +7,7 @@
 Run with --help or no arguments for possible commands.
 """
 from __future__ import print_function
-
+from datetime import datetime
 import argparse
 import json
 import os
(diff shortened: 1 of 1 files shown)


xenia-build
```diff
@@ -1540,6 +1570,47 @@ class TidyCommand(Command):
...
+    def execute(self, args, pass_args, cwd):
+        if args['class']:
+            path = args['class']
+
+            status = generate_source_file('{}.h'.format(path))
+            if status > 0:
+                return status
+            status = generate_source_file('{}.cc'.format(path))
+            if status > 0:
+                return status
```
gibbed, line 1601: Need to do cleanup here if the `.cc` generation fails, it'll leave a stray `.h`.

gibbed (changes requested):

xenia-build
Razzile: good point, will look into this tomorrow

gibbed (approved):

PR #1699, opened 2020-11-24
[VFS] Avoid exception on remove/remove_all.

gibbed:


src/xenia/vfs/devices/host_path_entry.cc
```diff
@@ -97,13 +97,14 @@ std::unique_ptr<Entry> HostPathEntry::CreateEntryInternal(
 
 bool HostPathEntry::DeleteEntryInternal(Entry* entry) {
   auto full_path = host_path_ / xe::to_path(entry->name());
+  std::error_code ec;  // avoid exception on remove/remove_all failure
   if (entry->attributes() & kFileAttributeDirectory) {
     // Delete entire directory and contents.
-    return std::filesystem::remove_all(full_path);
+    return std::filesystem::remove_all(full_path, ec);
```
elad335, line 103: The overload with error_code returns an uintmax_t which can be -1 on failure.

gibbed: Ah, good catch. I didn't fully check the documentation for remove_all and thought it was the same, because of how it's used here. Prior mistake in the implementation.

PR #1703, opened 2020-11-27
Defer XAM UI functions & other improvements.

gibbed:
- [Kernel] Fix global locking in kernel dispatch thread.
- [Kernel] Improve `CompleteOverlappedDeferred`/`CompleteOverlappedDeferredEx`.
- [XAM] Identify unknowns in `XamNotifyCreateListener`/`XNotifyListener`.
- [XAM] Defer `XamShowMessageBoxUI`.
- [XAM] Defer `XamShowKeyboardUI`.
- [XAM] Fix cancel handling in `XamShowKeyboardUI`.
- [XAM] Defer `XamShowDeviceSelectorUI`.
- [XAM] Defer `XamShowDirtyDiscErrorUI`.

### DIFF
--- a/src/xenia/kernel/kernel_state.cc
+++ b/src/xenia/kernel/kernel_state.cc
@@ -326,16 +326,20 @@ void KernelState::SetExecutableModule(object_ref<UserModule> module) {
           // As we run guest callbacks the debugger must be able to suspend us.
           dispatch_thread_->set_can_debugger_suspend(true);
 
+          auto global_lock = global_critical_region_.AcquireDeferred();
           while (dispatch_thread_running_) {
-            auto global_lock = global_critical_region_.Acquire();
+            global_lock.lock();
             if (dispatch_queue_.empty()) {
               dispatch_cond_.wait(global_lock);
               if (!dispatch_thread_running_) {
+                global_lock.unlock();
                 break;
               }
             }
             auto fn = std::move(dispatch_queue_.front());
             dispatch_queue_.pop_front();
+            global_lock.unlock();
+
             fn();
           }
           return 0;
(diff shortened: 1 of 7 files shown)


src/xenia/kernel/xam/xam_notify.cc
```diff
@@ -47,14 +55,15 @@ dword_result_t XNotifyGetNext(dword_t handle, dword_t match_id,
...
+    return X_E_INVALIDARG;
   }
   *id_ptr = 0;
+
   // Grab listener.
   auto listener =
       kernel_state()->object_table()->LookupObject<XNotifyListener>(handle);
   if (!listener) {
-    return X_ERROR_INVALID_HANDLE;
+    return 0;
```
Gliniak, line 66: Are you sure about it? It seems kinda strange to return success when there is no listener

gibbed: Yes, in fact I forgot to correct an earlier return value.

gibbed: This isn't success, this is a 1/0 "got notification" and "didn't get notification".

Gliniak: Ohh fact. I was looking (while making changes to it) at function inside which returns indeed error codes, but then converts it. My bad

src/xenia/kernel/kernel_state.cc
```diff
@@ -326,16 +326,22 @@ void KernelState::SetExecutableModule(object_ref<UserModule> module) {
           // As we run guest callbacks the debugger must be able to suspend us.
           dispatch_thread_->set_can_debugger_suspend(true);
 
+          auto global_lock = global_critical_region_.Acquire();
+          global_lock.unlock();
```
Triang3l, line 330: AcquireDeferred -> std::defer_lock_t?

Triang3l (changes requested):

src/xenia/kernel/kernel_state.cc
gibbed: Are you suggesting to disconnect this code from the GCR?

Triang3l: No, std::unique_lock constructor with std::defer_lock_t can create a lock object without actually locking.

gibbed: Ah, I wasn't aware. I'll take a look, thanks.

Issue #1558, opened 2020-02-23
NtAllocateVirtualMemory - Zeroing memory with X_PAGE_NOACCESS or X_PAGE_READONLY protection flag


Gliniak:
### Overview:
In current implementation of **_NtAllocateVirtualMemory_** there might be situation where we're trying to clear page with flag: **_X_PAGE_NOACCESS_** or **_X_PAGE_READONLY_**

Fragment where problem exists: [Here](https://github.com/xenia-project/xenia/blob/master/src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc#L159)

### Games with observable behaviour:
 - [Rabbids Invasion](https://marketplace.xbox.com/en-US/Product/Rabbids-Invasion-The-Interactive-TV-Show/66acd000-77fe-1000-9115-d802555308d0) (Demo) [xenia.log](https://github.com/xenia-project/xenia/files/4242005/xenia.log)
 - [Saints Row®: The Third™](https://marketplace.xbox.com/en-gb/Product/Saints-Row-The-Third/66acd000-77fe-1000-9115-d8025451086d)
 - [Saints Row IV](https://marketplace.xbox.com/en-gb/Product/Saints-Row-IV/66acd000-77fe-1000-9115-d8024b4d07f6)

**_Solution:_**
Easiest solution is to add check for protect_bits e.g.

 `  if (address && !(alloc_type & X_MEM_NOZERO)) {
    if (alloc_type & X_MEM_COMMIT &&
        !(protect_bits & (X_PAGE_NOACCESS | X_PAGE_READONLY))) {
      kernel_memory()->Zero(address, adjusted_size);
    }
  }`

< Place for more information >

Triang3l:
Since read-only pages likely need to be zeroed too (the function should return "brand new" pages — and while it's not as important for titles on the Xbox 360 as it is in a multi-process environment on Windows, it must not leak previous contents for security, since they might belong to a different app and contain sensitive data), we should probably first allocate the pages as writable, then zero them, and then change the protection level to what was asked by the game. Ideally, however, copy-on-write needs to be done, without actually clearing until the page is actually used, but I'm not sure what we can do here considering guest memory is backed by a memory-mapped file.

MarioSonic2987:
With the workaround posted in the OP, they almost reach go to ingame, but they crash in these moments,

[...]

PR #1712, opened 2020-11-29
[Kernel] Zeroing out pages without write protect flag

Gliniak:
Issue: https://github.com/xenia-project/xenia/issues/1558

We talked about it on discord, so I'm not gonna write essay here 💃 

Optional todo: Maybe remove some of the empty lines? I added them to increase readability, but maybe they're confusing 🤔


MarioSonic2987:
Saints Row: The Third and Saints Row IV goes ingame for the first time:
image
image

PR #1713, opened 2020-11-29
[XAM/Net] Implemented WSASetLastError

Gliniak:


gibbed:
`WSASetLastError` on Windows doesn't return a value, are you sure this one does?

Issue #1667, opened 2020-11-02
[Memory] VirtualAlloc behaviour while reusing allocated page


Gliniak:
I found out that actual behaviour of VirtualAlloc isn't exactly correct.

Affected methods: ``BaseHeap::AllocFixed``, ``BaseHeap::AllocRange`` (probably)

# Actual Behaviour
Right now when we're recommiting already commited page it clears that page content, which is incorrect.

# Expected Behaviour
If we have already commited page and trying to recommit or reserve (from what I understand you cannot reserve already reserved page, but you can reserve if it is commited) it then previous content should stay intact.

Microsoft Docs: https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc
Test case: https://pastebin.com/ZVhff5Ec (Based of Microsoft example)

<placeholder 💃 >

Gliniak:
Resolved in: https://github.com/xenia-project/xenia/pull/1718

Triang3l:
Not fully resolved yet possibly — still need to validate the behavior in a range with both committed and uncommitted pages, rather than the original allocation region for the beginning of the range.

Gliniak:
Sorry, my bad.

PR #1718, opened 2020-12-06
[Kernel/Memory] Do not clear range if it was previously commited

Gliniak:
Related to issue: #1667 

Most of the info is included in that issue (Including example code). I couldn't add such check when we're not providing ``*base_addr_ptr`` from obvious reason. However I wouldn't expect that to be such a big deal.

Note for testers please check if this PR causes any regression based on xenia: https://github.com/xenia-project/xenia/commit/ff5c5f01e0e31abee675be7e48c11c20642a40fb

Affected Titles:
- Serious Sam: TFE
- Perfect Dark Zero

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
@@ -136,6 +136,8 @@ dword_result_t NtAllocateVirtualMemory(lpdword_t base_addr_ptr,
   uint32_t protect = FromXdkProtectFlags(protect_bits);
   uint32_t address = 0;
   BaseHeap* heap;
+  HeapAllocationInfo prev_alloc_info = {};
+  bool was_commited = false;
 
   if (adjusted_base != 0) {
     heap = kernel_memory()->LookupHeap(adjusted_base);
@@ -143,6 +145,8 @@ dword_result_t NtAllocateVirtualMemory(lpdword_t base_addr_ptr,
       // Specified the wrong page size for the wrong heap.
       return X_STATUS_ACCESS_DENIED;
     }
+    was_commited = heap->QueryRegionInfo(adjusted_base, &prev_alloc_info) &&
+                   (prev_alloc_info.state & kMemoryAllocationCommit) != 0;
 
     if (heap->AllocFixed(adjusted_base, adjusted_size, page_size,
                          allocation_type, protect)) {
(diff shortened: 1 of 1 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -136,13 +136,15 @@ dword_result_t NtAllocateVirtualMemory(lpdword_t base_addr_ptr,
   uint32_t protect = FromXdkProtectFlags(protect_bits);
   uint32_t address = 0;
   BaseHeap* heap;
+  HeapAllocationInfo* prev_alloc_info = new HeapAllocationInfo();
```
gibbed, line 139: You are leaking memory.

`HeapAllocationInfo prev_alloc_info = {};`

`heap->QueryRegionInfo(adjusted_base, &prev_alloc_info);`

etc

gibbed (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
gibbed: I'd say it's probably best as something like:

```c
HeapAllocationInfo prev_alloc_info = {};
bool was_commit = heap->QueryRegionInfo(adjusted_base, &prev_alloc_info) && (prev_alloc_info.kMemoryAllocationCommit & kMemoryAllocationCommit) != 0;
```

Then later use `if (!was_commit)`.

Gliniak: I disagree with second post. 
I would need to make ``bool was_commit`` set it from default to false and then inside branches aka:
 ```if (adjusted_base != 0) {```
get that info and store it to that boolean, but at this point we have it already in prev_alloc_info, so imo usage of ``bool was_commit`` is quite unnecessary.

I will just add note that should resolve eventual confusion

gibbed: The point here is to never used an uninitialized `prev_alloc_info`; even if we initialize it to zero, there's no true guarantee as to what it is in `prev_alloc_info` if `QueryRegionInfo` returns `false`.

Gliniak: Resolved

PR #1721, opened 2020-12-14
Add GitHub Actions CI

Margen67:


JoelLinn:
As discussed in discord, I am not a fan of actions.
They are broken by design as they parse commands from stdout. That alone is a terrible idea already and tells much about the general design decisions taken.
It [was exploitable](https://bugs.chromium.org/p/project-zero/issues/detail?id=2070&can=2&q=&colspec=ID%20Type%20Status%20Priority%20Milestone%20Owner%20Summary&cells=ids) in the past, and may even still be depending on the action script and project. 
I just do not trust binaries from github actions. As the issue shows, privileged action execution can be invoked by third parties by creating cross branch PRs, for which they can control the Issue header and body, possibly invoking workflow commands. That issue still persists, only `set-env` [was removed](https://github.com/actions/runner/commit/a2e32170fd83851569804fa7b27d2cae235a5225), other APIs like `save-state` which can overwrite values by later build steps are still available.
If they add Workflow Commands to control things concerning the repository itself (like generating a release) we wouldn't notice and it would be exploitable immediately.

It's effectively like SQL injection into a search box. Imho it doesn't matter what the power of the available command set is. :-(

PR #1724, opened 2020-12-17
Add CI file for drone.io

JoelLinn:
Since travis-ci is buggy, slow and now wants money (if you don't want to beg every week)

[drone.io](https://drone.io) is an open source CI tool.
[cloud.drone.io](https://cloud.drone.io) offers that tool for free (for the time being...) on fast servers.
When all _free_ CIs eventually go down we could self host it (in theory a windows runner could be added then).

In about 10 minutes, the config does:
- lint
- build
  Debug/Release using premake and cmake (to verify the devenv isn't broken)
- test
  `xenia-base-tests` for premake and cmake release builds

To minimize build times I used a prebuilt docker image which has all dependencies. So no installers are run in the CI build (which is not just slower, it is a tremendous waste of energy as well!)

I suggest the [docker image repo](https://github.com/JoelLinn/docker-xenia-buildenv) is added to the xenia-project group. Also a docker-hub account would need to be created to host it (docker hub builds it automatically). I know it adds complexity but I think it's worth it.

cloud.drone.io doesn't host artifacts but as with all CIs, things can be pushed over the wire... (There is plugins that do that with minimal config)

Also there is no stupid stdout commands as with Github actions ... The yml can even be signed, so if somebody managed to edit it, the CI will refuse to run it.

To see the xenia pipeline in the CI interface, [see here](https://cloud.drone.io/JoelLinn/xenia/31).


JoelLinn:
* Rebased onto latest master
* Changed build environment to `Ubuntu 20.04 Focal Fossa` using `clang-10`

https://cloud.drone.io/JoelLinn/xenia/31

We can merge this as is - which allows us to test drone continuously.
If the decision is made to drop travis in favor of drone, we can move the Docker image (that has pre-installed packages/toolchain) to an official xenia repo.

.drone.yml
```diff
@@ -0,0 +1,206 @@
...
+  commands:
+  - |
+    ./xenia-build premake --devenv=cmake
+    cd build
+    for c in Debug Release
+    do
+      mkdir cmake-$c
+      cd cmake-$c
+      cmake -DCMAKE_BUILD_TYPE=$c ..
+      cd ..
```
Margen67, line 66:
```suggestion
      pushd cmake-$c
      cmake -DCMAKE_BUILD_TYPE=$c ..
      popd
```

JoelLinn:
@xenia-ci what seems to be your problem :/

.drone.yml
JoelLinn: I prefer it to be explicit here.
Also we only go in one directory.
It would additionally print a line, making this more verbose than required

PR #1726, opened 2020-12-25
[Kernel] Implemented NtSetIoCompletion

Gliniak:
Additionally: "Fixed" issue where executing WaitForNotification with wait_ticks 0 caused it to return timeout.

I'm still leaving that change to debate, as it seems to be hacky-like fixed (that is why I quoted it).
It seems like it is timing issue between ``NtSetIoCompletion`` and ``NtRemoveIoCompletion`` if they're executed too close to each other sometimes it might return timeout even if it shouldn't

# Affected titles:
- [Naughty Bear: Gold Edition](https://github.com/xenia-project/game-compatibility/issues/680)
- [Wet](https://github.com/xenia-project/game-compatibility/issues/671)


gibbed:
Could you split the timeout change to a separate commit (in the same PR is fine)? Otherwise looks good.

Gliniak:
Done

PR #1727, opened 2020-12-28
[Kernel] Implement NtReadFileScatter

emoose:
This implements the NtReadFileScatter kernel export, the export code is pretty much identical to NtReadFile besides the buffer parameter being a pointer to an array of buffer pointers (seems these are meant to be of union FILE_SEGMENT_ELEMENT, but they can be treated as uint32 fine)

This buffers parameter is passed along to a newly added XFile::ReadScatter, which handles looping over the remaining bytes to read and using XFile::Read to read into each buffer in turn.

(AFAIK the buffers are each 4096 bytes, most info on the net about NtReadFileScatter says that the buffers are memory-page-sized, not sure if that means the buffer size could change depending on page-size of the buffers memory area though - the only game I've seen using this always seems to use 4096 bytes for each buffer)

In order to use XFile::Read I needed to add a new parameter to stop it from sending an XIOCompletion notification, else it'd put the wrong number of read bytes into the notification (the game only seemed to work properly when this notification had the full size of the ReadScatter call)

This XIOCompletion/notification stuff is also the reason I had to add XFile::ReadScatter instead of handling it all inside NtReadFileScatter, as the notifications were protected to that class, adding a new function to the class seemed the cleanest solution for it.

Affected titles:
- [Halo 3 Delta Build (March 2007)](https://github.com/xenia-project/game-compatibility/issues/178#issuecomment-751601895)

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
@@ -265,6 +265,93 @@ dword_result_t NtReadFile(dword_t file_handle, dword_t event_handle,
 }
 DECLARE_XBOXKRNL_EXPORT2(NtReadFile, kFileSystem, kImplemented, kHighFrequency);
 
+dword_result_t NtReadFileScatter(dword_t file_handle, dword_t event_handle,
+                                 lpvoid_t apc_routine_ptr, lpvoid_t apc_context,
+                                 pointer_t<X_IO_STATUS_BLOCK> io_status_block,
+                                 lpdword_t segment_array, dword_t length,
+                                 lpqword_t byte_offset_ptr) {
+  X_STATUS result = X_STATUS_SUCCESS;
+
+  bool signal_event = false;
+  auto ev = kernel_state()->object_table()->LookupObject<XEvent>(event_handle);
+  if (event_handle && !ev) {
+    result = X_STATUS_INVALID_HANDLE;
+  }
+
+  auto file = kernel_state()->object_table()->LookupObject<XFile>(file_handle);
+  if (!file) {
+    result = X_STATUS_INVALID_HANDLE;
+  }
+
+  if (XSUCCEEDED(result)) {
+    if (true || file->is_synchronous()) {
+      // Synchronous.
+      uint32_t bytes_read = 0;
+      result = file->ReadScatter(
+          segment_array.guest_address(), length,
+          byte_offset_ptr ? static_cast<uint64_t>(*byte_offset_ptr) : -1,
+          &bytes_read, apc_context);
+      if (io_status_block) {
+        io_status_block->status = result;
+        io_status_block->information = bytes_read;
+      }
+
+      // Queue the APC callback. It must be delivered via the APC mechanism even
+      // though were are completing immediately.
+      // Low bit probably means do not queue to IO ports.
+      if ((uint32_t)apc_routine_ptr & ~1) {
+        if (apc_context) {
+          auto thread = XThread::GetCurrentThread();
+          thread->EnqueueApc(static_cast<uint32_t>(apc_routine_ptr) & ~1u,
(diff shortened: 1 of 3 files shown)


src/xenia/kernel/xfile.cc
```diff
@@ -160,18 +160,76 @@ X_STATUS XFile::Read(uint32_t buffer_guest_address, uint32_t buffer_length,
...
+  // "FILE_SEGMENT_ELEMENT", but they can just be treated as normal pointers
+  xe::be<uint32_t>* buffers = reinterpret_cast<xe::be<uint32_t>*>(
+      memory()->TranslateVirtual(buffers_guest_address));
+
+  // TODO: not sure if this is meant to change depending on buffer address?
+  uint32_t page_size = 4096;
+
+  uint32_t read_total = 0;
+  uint32_t read_remain = length;
+  while (read_remain) {
```
Triang3l, line 196: Can ReadFileScatter be used on Windows instead of the loop?

src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
```diff
@@ -265,6 +265,90 @@ dword_result_t NtReadFile(dword_t file_handle, dword_t event_handle,
 }
 DECLARE_XBOXKRNL_EXPORT2(NtReadFile, kFileSystem, kImplemented, kHighFrequency);
 
+dword_result_t NtReadFileScatter(dword_t file_handle, dword_t event_handle,
+                                 lpvoid_t apc_routine_ptr, lpvoid_t apc_context,
+                                 pointer_t<X_IO_STATUS_BLOCK> io_status_block,
+                                 lpdword_t buffer_ptr, dword_t length,
```
Triang3l, line 271: `segment_array` would be easier to understand in the future, since "buffer" in file reading usually implies just one area where to read the data to.

src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
```diff
@@ -265,6 +265,90 @@ dword_result_t NtReadFile(dword_t file_handle, dword_t event_handle,
...
+  auto file = kernel_state()->object_table()->LookupObject<XFile>(file_handle);
+  if (!file) {
+    result = X_STATUS_INVALID_HANDLE;
+  }
+
+  if (XSUCCEEDED(result)) {
+    if (true || file->is_synchronous()) {
+      // Synchronous.
+      uint32_t bytes_read = 0;
+      result = file->ReadScatter(
```
Triang3l, line 290: NtReadFileScatter may be used to load into physical memory, memory write callbacks should be triggered similar to how it's done in Xenia for non-scattered file reads.

Triang3l (changes requested):

src/xenia/kernel/xfile.cc
emoose: Looks like ReadFileScatter only works asynchronously, but AFAIK all file IO is done synchronous atm without the FILE_FLAG_OVERLAPPED flag that ReadFileScatter needs.. If async support is ever added it might be worth looking into though, I'll add a note to the unused async part of NtReadFileScatter about it.

src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
emoose: Agreed, will change that now

src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
emoose: ReadScatter calls into XFile::Read for the actual read into the buffer, that should be triggering the memory callbacks for us I think.

Triang3l (approved):

gibbed (approved):

PR #1729, opened 2020-12-29
[Kernel] Implement XeCryptBnQwNeRsaPubCrypt via BCrypt (win32 only)

emoose:
(E: saw this was linked at a few places, if you came here to get the Halo 3 debug build running you can see what's needed at https://github.com/xenia-project/game-compatibility/issues/178#issuecomment-751601895)

This implements the XeCryptBnQwNeRsaPubCrypt export via Windows BCrypt provider, this seemed like the easiest option for now to get this export working, since I wasn't too sure about trying to add in any other third-party crypto suites (it could be worth looking into though, since we're now including custom SHA1/SHA256/AES/RC4/DES & partial RSA, maybe there's a library somewhere that already implements all those for us?)

Getting this working with BCrypt was a pain, not a whole lot of documentation about using it with RSA on the net.. probably wasted an hour trying things with BCryptDecrypt (you'd think the act of taking gibberish data & turning it to clear-text would be decryption, but for some reason this only works with public keys when using BCryptEncrypt instead), in the end I even had to dig into [SymCrypt's sources](https://github.com/microsoft/SymCrypt) to find out that decryption only works with a private key...

Luckily people [already figured out](https://github.com/mtolly/X360/blob/master/X360/X360/STFS/STFSStuff.cs#L327) how the XeCrypt modulus/cipher-text needed to be bit-fiddled for Win32 APIs to accept it, took me a while to figure out the right order for BCrypt but eventually got it working, if this doesn't get merged hopefully the method used here might still come in useful with some other cipher suite.

Also implemented XeCryptBnQw_SwapDwQwLeBe too as Halo 3 Delta also uses that, seemed like kernel was swapping every 4 bytes but trying that just seemed to stop the game from accepting RsaPubCrypt output, swapping each 8 bytes instead seems to let it work fine.

Affected titles:
- [Halo 3 Delta Build (March 2007)](https://github.com/xenia-project/game-compatibility/issues/178#issuecomment-751601895) - can now load maps fine without needing code patches

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
@@ -7,11 +7,18 @@
 ******************************************************************************
 */
 
+#include <algorithm>
+
 #include "xenia/base/logging.h"
+#include "xenia/base/platform.h"
 #include "xenia/kernel/kernel_state.h"
 #include "xenia/kernel/util/shim_utils.h"
 #include "xenia/kernel/xboxkrnl/xboxkrnl_private.h"
 #include "xenia/xbox.h"
+
+#ifdef XE_PLATFORM_WIN32
+#include "xenia/base/platform_win.h"  // for bcrypt.h
+#endif
 
 #include "third_party/crypto/TinySHA1.hpp"
 #include "third_party/crypto/des/des.cpp"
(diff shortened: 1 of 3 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
```diff
@@ -216,19 +222,137 @@ void XeCryptSha256Final(pointer_t<XECRYPT_SHA256_STATE> sha_state,
...
+  }
 }
-DECLARE_XBOXKRNL_EXPORT1(XeCryptBnQw_SwapDwQwLeBe, kNone, kStub);
+DECLARE_XBOXKRNL_EXPORT1(XeCryptBnQw_SwapDwQwLeBe, kNone, kImplemented);
 
-dword_result_t XeCryptBnQwNeRsaPubCrypt(lpqword_t qw_a, lpqword_t qw_b,
-                                        lpvoid_t rsa) {
+typedef struct {
+  xe::be<uint32_t> size;  // size of modulus in 8 byte units
+  uint32_t public_exponent;
```
Triang3l, line 236: Should probably be `xe::be<uint32_t>` for consistency (especially when other implementations are added), with byte swap explicitly bypassed (or done twice). The rest looks pretty good.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
emoose: Yeah true, looks like I can use .value to get the unswapped value too, updated it with that instead

Triang3l (approved):

emoose:
Fixed a small issue with the format strings

emoose:
Updated to use `std::reverse_copy` instead of a loop for the modulus reversing

src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
```diff
@@ -7,12 +7,18 @@
...
 
+#include <algorithm>
+
 #include "xenia/base/logging.h"
 #include "xenia/kernel/kernel_state.h"
 #include "xenia/kernel/util/shim_utils.h"
 #include "xenia/kernel/xboxkrnl/xboxkrnl_private.h"
 #include "xenia/xbox.h"
 
+#ifdef XE_PLATFORM_WIN32
```
Triang3l, line 18: xenia/base/platform.h should be included for this definition, as there are no other includes in this file that clearly imply it being included.

src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
```diff
@@ -216,19 +222,131 @@ void XeCryptSha256Final(pointer_t<XECRYPT_SHA256_STATE> sha_state,
...
 
-// Byteswap?
-dword_result_t XeCryptBnQw_SwapDwQwLeBe(lpqword_t qw_inp, lpqword_t qw_out,
-                                        dword_t size) {
-  return 0;
+// Byteswaps each 8 bytes
+void XeCryptBnQw_SwapDwQwLeBe(pointer_t<uint64_t> qw_inp,
+                              pointer_t<uint64_t> qw_out, dword_t size) {
+  for (uint32_t i = 0; i < size; i++) {
+    qw_out[i] = xe::byte_swap(qw_inp[i]);
```
Triang3l, line 229: xe::copy_and_swap is preferable.

Triang3l (changes requested):

emoose:
Changed as requested, also moved the input/output byteswap to happen closer to when those are actually being used, and changed the two calls to XeCryptBnQw_SwapDwQwLeBe to use xe::copy_and_swap directly instead.

YzyDog:
nice work any chance i can test it?

gibbed (changes requested):
Looks good aside the missing include as Triangle mentioned.

Triang3l (approved):

gibbed (approved):

PR #1730, opened 2021-01-03
[Kernel/XAM] XamContentCreateEx - Extended Error Support

Gliniak:
Based on disassembly of Silent Hill: Homecoming

I figured out that it checks for very specific extended error codes. 
On screenshot you can see two of them: 
0x80070003 - X_ERROR_PATH_NOT_FOUND
0x80070570

and there is also one more check for:
0x80070026

We're not returning the first one (correct path) or any other error code and games goes a little insane and runs second instance of main menu in the background which causes issues.

image

# What does it fix
- Silent Hill: Homecoming - It fixes issue when pressing Start Game/Load Game causes window to pop and then returns to main menu & saving
- Ridge Racer 6 - Fixes saving
- Ridge Racer Unbounded - Fixes initial save (second save still crashes due to used handle to savefile)
- Wartech: Senko no Ronde - Resolves no device found error (before main menu) & saving issues

### DIFF
--- a/src/xenia/kernel/xam/xam_content.cc
+++ b/src/xenia/kernel/xam/xam_content.cc
@@ -202,8 +202,12 @@ dword_result_t XamContentCreateEx(dword_t user_index, lpstring_t root_name,
   }
 
   if (overlapped_ptr) {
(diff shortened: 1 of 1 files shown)


Gliniak:
Updated!

Now it will be used for every function that uses overlap

PR #1732, opened 2021-01-04
[CPU/XEX] Fix ApplyPatch not updating security_info_, causing issues with some patches

emoose:
- Move security info conversion code into ReadSecurityInfo
- Update ApplyPatch to call ReadSecurityInfo after patching headers

Right now ApplyPatch doesn't update security_info_ after applying a patch, making image_size() & other code that relies on security_info_ return outdated/incorrect data.

This can be seen when applying the Saints Row title update, before applying patch image_size() returns a normal 0x02160000, but after patching it returns 0x6d060000, which ends up breaking the memory alloc code inside ApplyPatch.
Changing ApplyPatch to update security info after patching it returns a more normal 0x02040000 image size instead (and can be seen that security_info_.page_descriptor_count is lower after the patch, so I'm guessing the incorrect image size was likely due to the image_size() function using the older count, making it read past the page descriptor bounds)

Affected titles:
- Saints Row 1 - TU1 can now be applied without issue

### DIFF
--- a/src/xenia/cpu/xex_module.cc
+++ b/src/xenia/cpu/xex_module.cc
@@ -299,6 +299,9 @@ int XexModule::ApplyPatch(XexModule* module) {
     module->xex_header_mem_.resize(header_target_size);
   }
 
+  // Update security info context with latest security info data
+  module->ReadSecurityInfo();
+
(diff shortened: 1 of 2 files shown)


gibbed:
Great catch!

PR #1733, opened 2021-01-04
[XAM] XamContentGetCreator check if package exists

Gliniak:
This change is quite simple and self explanatory.

# Affected Titles
 - Tenchu Z - Working Saves


gibbed:
> `[Kernel/Xam] Changed behaviour of setting.from element`

This is pretty bad commit message, please include the why so people don't have to look up the related PR to know.

PR #1737, opened 2021-01-15
[Kernel/XAM] CloseContent: Close files related to closed content

Gliniak:
I found that few games doesn't close explicitly handles to savefiles. They do not use ``NtClose`` to close these files (even on console)

However they use ``XamContentClose`` and seems like this function should be also responsible for closing opened files within content.

< space for more info >

If you know how to simplify this even more feel free to tell me 🐑 

# Impacted Games:
- Ridge Racer Unbounded
- MindJack
- Spyro
- Saint's Row

### DIFF
--- a/src/xenia/kernel/xam/content_manager.cc
+++ b/src/xenia/kernel/xam/content_manager.cc
@@ -15,6 +15,7 @@
 #include "xenia/base/filesystem.h"
 #include "xenia/base/string.h"
 #include "xenia/kernel/kernel_state.h"
+#include "xenia/kernel/xfile.h"
 #include "xenia/kernel/xobject.h"
 #include "xenia/vfs/devices/host_path_device.h"
 
@@ -197,6 +198,7 @@ X_RESULT ContentManager::CloseContent(const std::string_view root_name) {
   if (it == open_packages_.end()) {
     return X_ERROR_FILE_NOT_FOUND;
   }
+  CloseOpenedFilesFromContent(root_name);
 
   auto package = it->second;
   open_packages_.erase(it);
@@ -270,6 +272,28 @@ bool ContentManager::IsContentOpen(const ContentData& data) const {
                      [data](std::pair<string_key, ContentPackage*> content) {
                        return data == content.second->GetPackageContentData();
                      });
+}
+
+void ContentManager::CloseOpenedFilesFromContent(
+    const std::string_view root_name) {
+  // TODO(Gliniak): Cleanup this code to care only about handles
+  // related to provided content
+  const std::vector<object_ref<XFile>> all_files_handles =
+      kernel_state_->object_table()->GetObjectsByType<XFile>(
+          XObject::Type::File);
+
+  std::string resolved_path = "";
+  kernel_state_->file_system()->FindSymbolicLink(std::string(root_name) + ':',
+                                                 resolved_path);
+
+  for (const object_ref<XFile>& file : all_files_handles) {
+    std::string file_path = file->entry()->absolute_path();
+    bool is_file_inside_content = utf8::starts_with(file_path, resolved_path);
+
+    if (is_file_inside_content) {
+      file->ReleaseHandle();
+    }
+  }
 }
 
 }  // namespace xam
--- a/src/xenia/kernel/xam/content_manager.h
+++ b/src/xenia/kernel/xam/content_manager.h
@@ -159,6 +159,7 @@ class ContentManager {
   X_RESULT DeleteContent(const ContentData& data);
   std::filesystem::path ResolveGameUserContentPath();
   bool IsContentOpen(const ContentData& data) const;
+  void CloseOpenedFilesFromContent(const std::string_view root_name);
 
  private:
   std::filesystem::path ResolvePackageRoot(uint32_t content_type);


src/xenia/kernel/xam/content_manager.cc
```diff
@@ -259,6 +264,29 @@ std::filesystem::path ContentManager::ResolveGameUserContentPath() {
...
+
+  std::string resolved_path = "";
+  kernel_state_->file_system()->FindSymbolicLink(std::string(root_name) + ':',
+                                                 resolved_path);
+
+  for (const object_ref<XFile>& file : all_files_handles) {
+    std::filesystem::path file_path =
+        std::filesystem::path(file->entry()->absolute_path());
+
+    auto symlinked_path = file_path.parent_path().string() + '\\';
```
Gliniak, line 282: auto -> std::string

src/xenia/kernel/xam/content_manager.cc
```diff
@@ -259,6 +264,29 @@ std::filesystem::path ContentManager::ResolveGameUserContentPath() {
...
+    if (symlinked_path == resolved_path) {
+      opened_files.push_back(file);
```
Triang3l, line 284: It's not very desirable to create another vector overall, since that involves dynamic memory allocation, and even less to do that to just filter another vector. Currently, as there's only one usage of this function, it would be enough to just put that loop in CloseContent with ReleaseHandle done immediately. Later, I'm not sure, though a vector looks pretty clean from the readability perspective, something like a callback would have less unavoidable (on the usage side) overhead. Though it's just my personal opinion, other contributors may have a different preference.

src/xenia/kernel/xam/content_manager.cc
Triang3l, line 282: Would this `\` separator work on Linux, and is it possible that `resolved_path` won't have one? I'd avoid making any assumptions about the string form of a path, and path == string comparisons, since one path can be represented by many different strings, with the trailing `\` or `/` present or not present (while still being a valid directory path), with `.` and `..`.

Triang3l (changes requested):

src/xenia/kernel/xam/content_manager.cc
Gliniak: Implementation directly inside ContentClose was my scratch one, but it was quite ugly imo. 
Yes it adds overhead, but in this case should we even care? This vector shouldn't be freed after successful execution? 🤔

Also this method is used like max 20 times per game run (usually 2-3 times), so it is not that problematic in case of memory usage

src/xenia/kernel/xam/content_manager.cc
Gliniak: Paths here are paths from VFS, so they're independent from platform. (like ``\\Device\\content\\1``)

src/xenia/kernel/xam/content_manager.cc
Triang3l: Heap allocations take a somewhat long time, though of course it's tiny relatively to file system calls, but still it's better to do avoid them where possible (though we have lots of std::vectors and long enough std::strings not to work with small string optimization in the emulator overall, so… ugh).

src/xenia/kernel/xam/content_manager.cc
Triang3l: That sounds like an even worse case for std::filesystem::path usage, if I understand correctly, it can do normalization for the host OS internally; also comparison is case-sensitive here in this place, is that good?

Triang3l: std::filesystem::path overall should be avoided for guest paths, as they have Xbox 360's path format, not the host OS format. The VFS code of Xenia uses ResolvePath (accepting a raw string path) in many places, it's preferable to perform all operations with guest paths in terms of VFS objects, not std::filesystem::path or strings.

Gliniak: I was worried about both of these cases too.
``std::filesystem::path`` seems to work correctly for these cases. I haven't seen anything that would do some trickery to tinker with host os, but would be nice to be 100% sure. 
Case sensitive isn't an issue here, because both paths are based on the same data (initial symlink path) and they must match exactly.

Triang3l: `file_path.parent_path()` may become `/Device/content/1` instead of `\Device\content\1`.

Gliniak: ohhh right... Ok I will change it

Gliniak: Changed, but I made quite bold assumption about existing of ``\`` character in string. I guess we shouldn't check it as w/o one symlink wouldn't even exist nor file 🤔

src/xenia/kernel/xam/content_manager.cc
```diff
@@ -259,6 +261,26 @@ std::filesystem::path ContentManager::ResolveGameUserContentPath() {
...
+      kernel_state_->object_table()->GetObjectsByType<XFile>(
+          XObject::Type::File);
+
+  std::string resolved_path = "";
+  kernel_state_->file_system()->FindSymbolicLink(std::string(root_name) + ':',
+                                                 resolved_path);
+
+  for (const object_ref<XFile>& file : all_files_handles) {
+    std::string file_path = file->entry()->absolute_path();
+    file_path = file_path.substr(0, file_path.find_last_of('\\') + 1);
```
gibbed, line 276: `utf8::starts_with`

src/xenia/kernel/xam/content_manager.cc
```diff
@@ -259,6 +261,26 @@ std::filesystem::path ContentManager::ResolveGameUserContentPath() {
   return root_path_ / title_id / kGameUserContentDirName / user_name;
 }
 
+void ContentManager::CloseOpenedFilesFromContent(
```
gibbed, line 264: This needs a `TODO: clean this up eventually` or so. Iterating all open file handles to do this seems like a bad idea in the long run.

gibbed (changes requested):

Gliniak:
Updated
- Review changes

PR #1739, opened 2021-01-19
[XAM/Content] Implemented Custom Content Header Handling

Gliniak:
It is still initial release and still might require a little bit of work.

The location I choose seems to make sense as we're already creating custom profile directory on the same level.

# Things to debate/change:
- I still think about switching from ``XCONTENT_DATA`` to ``ContentData`` (if possible), but I like to store exactly how we receive them.
- Maybe a little bit of cleanup might be possible?
- Remove usage of ``X_RESULT`` and just check if function returned success/failure

# Affected Games
- Rockstar Titles

# Disclaimer 
I checked few games, but I'm not responsible if your savefile will get corrupted after this for some strange reason 🤷


src/xenia/kernel/xam/content_manager.cc
```diff
@@ -138,6 +146,53 @@ bool ContentManager::ContentExists(const ContentData& data) {
...
+  auto title_id = fmt::format("{:8X}", kernel_state_->title_id());
+  auto header_file_path =
+      root_path_ / title_id / kGameContentHeaderDirName / file_name;
+
+  if (std::filesystem::exists(header_file_path)) {
+    auto file = xe::filesystem::OpenFile(header_file_path, "rb");
+
+    std::vector<uint8_t> buffer;
+
+    fseek(file, 0, SEEK_END);
```
Triang3l, line 183: Error handling (just ifs and returns) after every stdio call would be good for safety.

Triang3l (changes requested):

MarioSonic2987:
It's now possible to save and load saves in GTA V.
image
image

Gliniak:
Updated:
- Switched from ``vector`` to ``array`` (We know exact size we need to read so why not)
- Switched from ``fread`` to ``std::filesystem::file_size``
- Added some error handling in case of corrupted headers

src/xenia/kernel/xam/content_manager.cc
```diff
@@ -138,6 +147,60 @@ bool ContentManager::ContentExists(const ContentData& data) {
...
+    std::array<uint8_t, header_size> buffer;
+
+    auto fize_size = std::filesystem::file_size(header_file_path);
+    if (fize_size < header_size) {
+      fclose(file);
+      return X_STATUS_END_OF_FILE;
+    }
+
+    size_t result =
+        fread(const_cast<uint8_t*>(buffer.data()), 1, header_size, file);
```
JoelLinn, line 192: why the const_cast?

src/xenia/kernel/xam/content_manager.cc
```diff
@@ -138,6 +147,59 @@ bool ContentManager::ContentExists(const ContentData& data) {
   return std::filesystem::exists(path);
 }
 
+X_RESULT ContentManager::CreateContentHeaderFile(const std::string file_name,
+                                                 const XCONTENT_DATA* data) {
+  auto title_id = fmt::format("{:8X}", kernel_state_->title_id());
+  auto header_path = root_path_ / title_id / kGameContentHeaderDirName;
```
emoose, line 153: So this would put the save headers at eg. `content\{title_id}\Headers\{file_name}.header`, while the data itself would be in `content\{title_id}\{content_type}\{package_name}`?

Could see some problems coming up with that, eg. two contents with the same name but different content type would have header files collide, as unlikely as that is (doubtful games even use content types other than the SaveGame content type, but there's always a chance, maybe themes/avatar awards I guess)

Not really sure if storing header in a different location to the data itself is a great idea neither, would make save transfers/X360 conversions a bit more complicated... people have been used to the `title_id\content_type\file_name` format for years too.

I've always been a fan of putting header in parent folder of the data myself: https://github.com/xenia-canary/xenia-canary/blob/dea24e2fcb099c181f9617221205952880c048ee/src/xenia/kernel/xam/content_manager.cc#L100
Actually thinking about it, X360 already does do something similar with GOD files, header is stored at the normal package path while data files are put inside `{path}.data` folder next to it (also think GFWL also does something like that for DLC packages too, but it's been a while since I even used GFWL so I could be wrong about that :P), maybe it's worth using the same format for consistency.

Maybe this was already discussed and decided on though, just some thoughts on it.

Gliniak:
Updated:
 - Splitted headers into content_type categories (Thanks @emoose for suggestion)

src/xenia/kernel/xam/content_manager.cc
```diff
@@ -138,6 +148,63 @@ bool ContentManager::ContentExists(const ContentData& data) {
...
+  if (std::filesystem::exists(header_path / header_filename)) {
+    auto file = xe::filesystem::OpenFile(header_path / header_filename, "wb");
+    fwrite(data, 1, sizeof(XCONTENT_DATA), file);
+    fclose(file);
+    return X_STATUS_SUCCESS;
+  }
+  return X_STATUS_NO_SUCH_FILE;
+}
+
+X_RESULT ContentManager::OpenContentHeaderFile(const std::string_view file_name,
```
gibbed, line 176: Isn't this misnamed? It reads the contents of a given path.

src/xenia/kernel/xam/content_manager.cc
```diff
@@ -138,6 +148,63 @@ bool ContentManager::ContentExists(const ContentData& data) {
   return std::filesystem::exists(path);
 }
 
+X_RESULT ContentManager::CreateContentHeaderFile(const std::string file_name,
```
gibbed, line 151: Isn't this misnamed? It writes contents to a given path.

src/xenia/kernel/xam/content_manager.cc
```diff
@@ -138,6 +148,63 @@ bool ContentManager::ContentExists(const ContentData& data) {
...
+                          content_type_directory / file_name;
+  constexpr uint32_t header_size = sizeof(XCONTENT_DATA);
+
+  if (std::filesystem::exists(header_file_path)) {
+    auto file = xe::filesystem::OpenFile(header_file_path, "rb");
+
+    std::array<uint8_t, header_size> buffer;
+
+    auto fize_size = std::filesystem::file_size(header_file_path);
+    if (fize_size < header_size) {
```
gibbed, line 191: Is there an expectation that the file size could ever not be the size of `XCONTENT_DATA`?

gibbed (changes requested):

src/xenia/kernel/xam/content_manager.cc
Gliniak: Tbh I should just check if it is different than ``XCONTENT_DATA`` size. I added this check in case someone will manually poke with these files.

Gliniak:
Updated:
- Switched naming from Create/Open to Write/Read
- Changed one check to match exactly size of header structure

halotroop2288 (approved):

PR #1740, opened 2021-01-21
[Kernel] Workarounds to allow XMountUtilityDrive to be successful

emoose:
Right now it seems even with `mount_cache` enabled, when games call XMountUtilityDrive (compiled into the XEX, from some SDK lib I guess) it'll fail due to checking some things we don't handle atm.

In best case the game will ignore the result and still access cache anyway, but in worst case the game may use the result to determine if HDD is even available at all (eg. pretty much all Halo games will only recognize HDD if this succeeds)

Fortunately the function isn't too complex to work around, mostly just needs some NtOpenFile/NtWriteFile calls to the `Partition1/Cache0/Cache1` devices to succeed, I've handled this in here by creating a new NullDevice class that just returns success to every call made to it. If mount_cache is set then `\Device\Harddisk0` will be created as a NullDevice, which then handles the `Partition1/Cache0/Cache1` paths for us.

The harder bit is what happens after those checks, seems the game sets something up with XamTaskSchedule which handles the actual cache filesystem stuff for it.
(In the past I had to try getting around this part since we didn't have any impl for XamTaskSchedule, but thankfully the one added recently mostly satisfies it)

~~The code ran by XamTaskSchedule seems to fail for some reason though (maybe because of missing StfsCreateDevice kernel export or something related to that, couldn't really track down the cause), but it's not a huge issue since we can just change the result from XamTaskSchedule if we see that it's a cache-partition-related task (the code passes along an STFC header to the task, so we can check for the 0x4A6F7368 magic to know if it's a cache task, offset of the header can change depending on the game though.. probably different revisions of the cache code)~~

~~With that the XMountUtilityDrive func should return success, however there's still an issue with AFAIK XFlushUtilityDrive (from strings in Halo 3 Epsilon), which seems to set some event and then get stuck waiting on another event. It looks like the XamTask is supposed to handle these via a KeWaitForSingleObject loop (which I guess means XamTaskSchedule should be running the task in it's own thread), but since the task proc failed before it started that loop, there's nothing to set the event that Flush is waiting on...~~

~~Luckily this function will skip over all the event code if `XexCheckExecutablePrivilege(0xB)` is set, so I've just added a check to that func to return true if it's checking 0xB & the game has previously set up cache with XamTaskSchedule.
(XMountUtilityDrive also has a check for the same privilege and skips the XamTask stuff if set, but unfortunately some games can go down bad code paths if this privilege is always set, eg Sega Rally Revo... but making sure the game has actually tried mounting the cache partition beforehand seems to let it work fine)~~

EDIT: above can all be solved by making the XamTaskSchedule callback run in its own thread, making that happen is easier said than done though, my other replies go into why.

---

Of course all this is a workaround (that I've tried to make as un-hacky as I could), not a proper fix... I guess the alternative would be to implement all the cache/STFC stuff properly (which would then probably turn cache into 4GiB of inaccessible blob files...) though if anyone has any other ideas how it could be handled I'd be happy to hear them.

I don't really know if we want such multi-part workarounds to be part of main Xenia tree though so I'll leave this as a draft for now, hopefully it'll come in useful for others that want to look into cache partition stuff, or at least can be pointed to if anyone ever needs support for cache. (would appreciate any review of the code though!)

(btw if anyone wants to look into the XMountUtilityDrive func at all, searching for usages of XamTaskSchedule import should find it, usually the only func that uses that)

Affected titles:
- Halo 3: HDD is now detected, allowing access to Theatre mode, and map cache files will get written into the cache folders
- Halo 3 Epsilon: similar to Halo 3, except some reason this needs NtWriteFile to support APC routines else it'll hang after writing to cache (#1410)
- Halo 3 Beta: will now load without error & is playable, as long as code is updated to also mount cache:\\, and support for import libraries is added (https://github.com/xenia-canary/xenia-canary/commit/e9eca0a4c95c8d3e8cbc7f94dbdf3af5ce9db2df)

### DIFF
--- a/src/xenia/app/xenia_main.cc
+++ b/src/xenia/app/xenia_main.cc
@@ -75,7 +75,6 @@ DEFINE_path(
     "Storage");
 
 DEFINE_bool(mount_scratch, false, "Enable scratch mount", "Storage");
-DEFINE_bool(mount_cache, false, "Enable cache mount", "Storage");
 
 DEFINE_transient_path(target, "",
                       "Specifies the target .xex or .iso to execute.",
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -42,6 +42,7 @@
 #include "xenia/ui/imgui_dialog.h"
 #include "xenia/vfs/devices/disc_image_device.h"
 #include "xenia/vfs/devices/host_path_device.h"
+#include "xenia/vfs/devices/null_device.h"
 #include "xenia/vfs/devices/stfs_container_device.h"
 #include "xenia/vfs/virtual_file_system.h"
 
@@ -644,6 +645,19 @@ std::string Emulator::FindLaunchModule() {
 
 X_STATUS Emulator::CompleteLaunch(const std::filesystem::path& path,
                                   const std::string_view module_path) {
+  if (cvars::mount_cache) {
+    // Below are accessed directly by cache/STFC code baked into games
+    // By using a NullDevice that just returns success to all IO requests it
+    // should allow games to continue believe cache was accessed successfully
+    auto null_paths = {std::string("\\Partition0"), std::string("\\Cache0"),
+                       std::string("\\Cache1")};
+    auto null_device =
+        std::make_unique<vfs::NullDevice>("\\Device\\Harddisk0", null_paths);
+    if (null_device->Initialize()) {
+      file_system_->RegisterDevice(std::move(null_device));
+    }
+  }
+
   // Reset state.
   title_id_ = 0;
   game_title_ = "";
--- a/src/xenia/kernel/xam/xam_task.cc
+++ b/src/xenia/kernel/xam/xam_task.cc
@@ -27,6 +27,8 @@ namespace xe {
 namespace kernel {
 namespace xam {
 
+constexpr uint32_t kCachePartitionMagic = 0x4A6F7368;  // 'Josh'
+
 struct XTASK_MESSAGE {
   be<uint32_t> unknown_00;
   be<uint32_t> unknown_04;
@@ -43,8 +45,6 @@ bool cache_task_scheduled_ = false;
 dword_result_t XamTaskSchedule(lpvoid_t callback,
                                pointer_t<XTASK_MESSAGE> message,
                                lpdword_t unknown, lpdword_t handle_ptr) {
-  assert_zero(unknown);
-
   // TODO(gibbed): figure out what this is for
   *handle_ptr = 12345;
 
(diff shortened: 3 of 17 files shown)


emoose:
Oops, opened this too early

emoose:
Think I've figured a way to remove the hacky bits from XamTaskSchedule & XexCheckExecutablePrivilege, just needs the XamTaskSchedule proc to run in it's own thread, though for that proc to get to its main loop without failing I had to write a quick stub for IoCreateDevice, and add some stub code for XFileAlignmentInformation to NtQueryInformationFile.

Unfortunately making the XamTaskSchedule proc run in its own thread doesn't seem that easy though - the game calls KeEnterCriticalRegion before running it which locks the xe::global_critical_region mutex, when the thread is then created for the XamTaskSchedule proc XThread::DeliverAPCs gets called during the thread init, which uses XThread::LockApc before handling the APCs, causing a deadlock since LockApc tries locking the same xe::global_critical_region that the game already locked earlier...

Looking at the [MSDN docs](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntddk/nf-ntddk-keentercriticalregion) for KeEnterCriticalRegion it does mention about APCs there, that "the execution of normal kernel APCs is _deferred_ until this driver exits the critical region by calling the KeLeaveCriticalRegion routine", so it seems maybe this APC deadlocking is a bug with our impl? I'm not really sure how to try making it defer the APC execution instead of deadlocking yet though.

For now I've tested this with a quick hack by making KeEnter/LeaveCriticalRegion use a different mutex to global_critical_region, and looks like all the XamTask stuff is working fine - when game tries using XFlushUtilityDrive the task seems to handle/set the right events properly since the game doesn't freeze up, and it's recognizing the HDD without any of that STFC header checking/XEX privilege stuff now too.

emoose:
Pushed the changes to KeEnter/LeaveCriticalRegion to make it use a per-thread mutex instead of the global_critical_region, and other small things needed by the XamTaskSchedule cache proc.

This gets rid of the nasty XamTaskSchedule header-checking & the XexCheckExecutablePrivilege hack, but I'm not really sure if a per-thread local_critical_region mutex is the way to handle this, since there would only ever be a single thread using that mutex (KeEnterCrit/Leave would only use it thru GetCurrentThread), so there wouldn't be any opportunity for it to block any other threads, pretty much making a mutex pointless.

It also currently doesn't really help with what critical regions are meant to be used for (seems they're meant for deferring KAPCs/preventing the thread from being suspended, which this per-thread mutex solution doesn't really change anything with right now)

The way CXBX handles it seems a little better by inc/decrementing a per-thread counter, I didn't see anything in there that actually checks that counter when handling APCs though, so maybe their impl. isn't fully complete neither...
When I have some free time I'll make a new issue about this critical region stuff, since I'm not sure what the best way to handle it is atm.

Issue #1735, opened 2021-01-10
Linux compile crash with errors


Mte90:
[//]: # (Describe what's going wrong:)

```
./xb build --config=release 
Building release...

- running premake...
Building configurations...
    Baking aes_128...
    Baking capstone...
    Baking dxbc...
    Baking discord-rpc...
    Baking cxxopts...
    Baking cpptoml...
    Baking fmt...
    Baking glslang-spirv...
    Baking imgui...
    Baking libavcodec...
    Baking libavutil...
    Baking mspack...
    Baking snappy...
    Baking spirv-tools...
    Baking volk...
    Baking xxhash...
    Baking xenia-core...
    Baking xenia-app-discord...
    Baking xenia-apu...
    Baking xenia-apu-nop...
    Baking xenia-base...
    Baking xenia-base-tests...
    Baking xenia-cpu...
    Baking xenia-cpu-tests...
    Baking xenia-cpu-ppc-tests...
    Baking xenia-cpu-backend-x64...
    Baking xenia-debug-ui...
    Baking xenia-gpu...
    Baking xenia-gpu-shader-compiler...
    Baking xenia-gpu-null...
    Baking xenia-gpu-vulkan...
    Baking xenia-gpu-vulkan-trace-viewer...
    Baking xenia-gpu-vulkan-trace-dump...
    Baking xenia-hid...
    Baking xenia-hid-demo...
    Baking xenia-hid-nop...
    Baking xenia-kernel...
    Baking xenia-ui...
    Baking xenia-ui-spirv...
    Baking xenia-ui-vulkan...
    Baking xenia-ui-window-vulkan-demo...
    Baking xenia-vfs...
    Baking xenia-vfs-dump...
    Baking xenia-apu-sdl...
    Baking xenia-helper-sdl...
    Baking xenia-hid-sdl...
    Baking xenia-app...
Running action 'gmake2'...
Done (2088ms).

- building (all):release...
make: ingresso nella directory «/home/mte90/Desktop/xenia/build»
==== Building aes_128 (release_linux) ====
==== Building capstone (release_linux) ====
==== Building dxbc (release_linux) ====
==== Building discord-rpc (release_linux) ====
Linking dxbc
Linking aes_128
Linking capstone
Linking discord-rpc
==== Building cpptoml (release_linux) ====
==== Building cxxopts (release_linux) ====
==== Building fmt (release_linux) ====
==== Building glslang-spirv (release_linux) ====
[...]

PR #1741, opened 2021-01-23
Some small build fixes and changes

sephiroth99:
The four commits are as follows

- Two are build fix for Linux on recent installations (I tested with clang-11 and libstdc++10) these should help #1735. The only issues left are with libav for me.
- One is a small cleanup to remove linking to unused libraries on Linux
- The last one is a missing header when using VS2017

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -136,10 +136,6 @@ filter({"platforms:Linux", "toolset:gcc"})
   end
 
 filter({"platforms:Linux", "language:C++", "toolset:clang"})
-  links({
-    "c++",
-    "c++abi"
-  })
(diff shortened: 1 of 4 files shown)


premake5.lua
```diff
@@ -137,8 +137,6 @@ filter({"platforms:Linux", "toolset:gcc"})
 
 filter({"platforms:Linux", "language:C++", "toolset:clang"})
   links({
-    "c++",
```
JoelLinn, line 140: we can remove the complete links() command here to keep it clean, the main premake is a mess already

JoelLinn (changes requested):

PR #1742, opened 2021-01-24
[Kernel/IO] Implemented XFileRenameInformation

Gliniak:
TLDR: Some games uses ``XFileRenameInformation`` to rename savefiles adhoc.

I'm pulling it as draft. I'm not sure if everything works as intended and have proper names.

There is for sure missing path validation, but I had no idea where to put that function to be available where it is requred.

# Affected Titles:
[Nascar: Inside Line](https://github.com/xenia-project/game-compatibility/issues/1498)

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_io_info.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_io_info.cc
@@ -237,6 +237,30 @@ dword_result_t NtSetInformationFile(
       }
       break;
     }
+    case XFileRenameInformation: {
+      auto info = info_ptr.as<X_FILE_RENAME_INFORMATION*>();
+      auto root_handle = uint32_t(info->root_dir_handle);
+
+      const std::string target_path =
+          std::string(kernel_memory()->TranslateVirtual<const char*>(
+                          info->ansi_string.pointer),
+                      info->ansi_string.length);
+
+      // Place IsValidPath in path from where it can be accessed everywhere
+      // if (!IsValidPath(target_path, false)) {
+      //  return X_STATUS_OBJECT_NAME_INVALID;
+      //}
+
+      const std::string file_name = target_path.substr(
+          target_path.find_last_of('\\') + 1, target_path.length());
+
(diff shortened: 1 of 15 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_io_info.cc
```diff
@@ -237,6 +237,30 @@ dword_result_t NtSetInformationFile(
...
+      // Place IsValidPath in path from where it can be accessed everywhere
+      // if (!IsValidPath(target_path, false)) {
+      //  return X_STATUS_OBJECT_NAME_INVALID;
+      //}
+
+      const std::string file_name = target_path.substr(
+          target_path.find_last_of('\\') + 1, target_path.length());
+
+      auto path = kernel_state()->file_system()->ResolvePath(
+          target_path.substr(0, target_path.find_first_of('\\')));
```
Gliniak, line 258: That's the big assumption that each renamed file will be stored in root directory I forgot to resolve

Regunavailable:
still trynna find fixes for this gamee

Gliniak:
(for future reference, when I will manage to write it even better) 
Added to canary:
https://github.com/xenia-canary/xenia-canary/commit/74b52711a36f5749ed17ac36678944a6459b7c4d

A bit simplified and works in more cases, but still isn't perfect.

Issues:
- Providing only filename as a path might cause it to crash. I haven't seen such case in games that use it, but it might be required in the future.
- Exposing host_path (package root) to host entry

Regunavailable:
Welp, thanks for the work! I can enjoy NASCAR now but problem is I get an Error "Lost GPU Driver (Due to internal error) I switch to Vulkan but the game looks Deep fried and vulkan just completely breaks the graphics. This is why we cannot have nice things.

Regunavailable:
*Update* Game works fine and you CAN race about 1 time until you start getting Graphical bugs and it will just Crash because it "lost" the Graphics driver. I can choose Vulkan but I have to sacrifice the game looking like Sh*t.. I'll continue waiting for a fix *If Possible* That is.

PR #1744, opened 2021-01-25
[Kernel/IO] IoCompletion: Removed !wait_ticks check

Gliniak:
By merging https://github.com/xenia-project/xenia/pull/1726 we introduced regression in at least 2 titles.

The case is that ``NtRemoveIoCompletion`` and ``NtSetIoCompletion`` are executed so frequently between 2 threads (iirc) that they run apart at some point what causes subthread that execute ``NtRemoveIoCompletion`` to run before main thread does ``NtSetIoCompletion`` and that causes lock.

I haven't found proper solution, however returning X_STATUS_SUCCESS instead of X_STATUS_TIMEOUT did a really nice trick.
I added note describing wth is going on just in case.

I also added these functions to high frequency one as in Monster High it can create 10gb log in around a minute. Which seems quite sketchy, but might be understanable. In WET they're using it to provide player movement (LOL)

What it resolved: (Thanks to https://github.com/Etokapa for testing)
- Crash in: Monster High & Monster Jam (Now they work even better than before)
- Huge memleak in WET during cutscenes

## Affected titles
- WET
- Monster Jam
- Monster High
- Naughty Bear
- <some more?>

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc
@@ -458,7 +458,8 @@ dword_result_t NtSetIoCompletion(dword_t handle, dword_t key_context,
   port->QueueNotification(notification);
   return X_STATUS_SUCCESS;
 }
-DECLARE_XBOXKRNL_EXPORT1(NtSetIoCompletion, kFileSystem, kImplemented);
+DECLARE_XBOXKRNL_EXPORT2(NtSetIoCompletion, kFileSystem, kImplemented,
+                         kHighFrequency);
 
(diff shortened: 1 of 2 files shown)


Gliniak:
I cannot set it to draft right now, but let it stay as draft change. 
Maybe I will figure out something better (proper?) soon 🤞

gibbed:
I've converted the PR to a draft PR.

Gliniak:
After spending some time analysing game behaviours I came to conclusion (which seems reasonable).

In case when: ``NtRemoveIoCompletion`` doesn't receive timeout it should set it to infinite instead of 0.
0 causes it to bail-out instantly and returning timeout instead of waiting for anything.

PR #1745, opened 2021-01-25
[Kernel/XAM] Dummy X_ACHIEVEMENT_DETAILS support

Gliniak:
This is mostly based on old canary achievement support.
I simplified it as much as possible to only provide basic implementation, which shouldn't crash titles.

# Affected Titles
- PGR4 - Fixes menu crash
- South Park XBLA - Fixes info popup about device being full (requires protect_zero)
- Some more?

### DIFF
--- a/src/xenia/kernel/xam/xam_user.cc
+++ b/src/xenia/kernel/xam/xam_user.cc
@@ -2,7 +2,7 @@
  ******************************************************************************
  * Xenia : Xbox 360 Emulator Research Project                                 *
  ******************************************************************************
- * Copyright 2013 Ben Vanik. All rights reserved.                             *
+ * Copyright 2021 Ben Vanik. All rights reserved.                             *
  * Released under the BSD license - see LICENSE in the root for more details. *
  ******************************************************************************
  */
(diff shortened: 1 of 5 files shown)


Razzile:
this'll be useful for new UI 👀

src/xenia/kernel/xam/xam_user.cc
```diff
@@ -552,23 +552,62 @@ dword_result_t XamShowSigninUI(dword_t unk, dword_t unk_mask) {
...
   auto e = object_ref<XStaticEnumerator>(
-      new XStaticEnumerator(kernel_state(), count, 500));
+      new XStaticEnumerator(kernel_state(), count, entry_size));
   auto result = e->Initialize(user_index, 0xFB, 0xB000A, 0xB000B, 0);
   if (XFAILED(result)) {
     return result;
   }
 
+  // Titles can provide incorrect achievements count
+  // For example "Way of the Dogg" provides 0x80 of them
```
gibbed, line 601: What do you mean by "incorrect"? It should match the allocated memory size, not the number of achievements the game has.

Gliniak: Fair point. It's not amount of achievements that game have, but amount of entries for which we should allocate memory.

So, even it game have let's say 10 achievements, it can request space for 20 items.

gibbed:
> I found an interesting case.
> `Read Dead Redemption` with dummy filled structure will crash on loading (due to nullptr). (skipable with protect_zero = false)
> 
> Idea how to resolve it is quite simple. Allocate space in heap for some dummy string and write that address to every _ptr in achievement structure. Is it acceptable?

Is it requesting the "basic" achievement data? I'd guess the 500 bytes minus `sizeof(XACHIEVEMENT_DATA)` is used for storing character data. So for dummy it should point to after the structure there.

src/xenia/kernel/xam/xam_user.cc
```diff
@@ -552,23 +552,62 @@ dword_result_t XamShowSigninUI(dword_t unk, dword_t unk_mask) {
 }
 DECLARE_XAM_EXPORT1(XamShowSigninUI, kUserProfiles, kStub);
 
+#pragma pack(push, 1)
+struct X_XACHIEVEMENT_DETAILS {
```
gibbed, line 556: `X_ACHIEVEMENT_DETAILS`?

gibbed (changes requested):

Gliniak:
> > I found an interesting case.
> > `Read Dead Redemption` with dummy filled structure will crash on loading (due to nullptr). (skipable with protect_zero = false)
> > Idea how to resolve it is quite simple. Allocate space in heap for some dummy string and write that address to every _ptr in achievement structure. Is it acceptable?
> 
> Is it requesting the "basic" achievement data? If not I'd guess the 500 bytes minus `sizeof(X_ACHIEVEMENT_DETAILS)` is used for storing character data. So for dummy it should point to after the structure there.

Even better. It provides flags as -1 (I guess it uses every existing flag 🤔). That is really interesting case to analyze

Gliniak:
Updated:
 - Typo Fix
 - Added code that creates pointer to "dummy text" to prevent games from crashing by accessing nullptr

src/xenia/kernel/xam/xam_user.cc
```diff
@@ -552,23 +552,70 @@ dword_result_t XamShowSigninUI(dword_t unk, dword_t unk_mask) {
...
-      new XStaticEnumerator(kernel_state(), count, 500));
+      new XStaticEnumerator(kernel_state(), count, entry_size));
   auto result = e->Initialize(user_index, 0xFB, 0xB000A, 0xB000B, 0);
   if (XFAILED(result)) {
     return result;
   }
 
+  std::u16string dummy_achievement_text = to_utf16("Dummy Text");
+  uint32_t alloc_ptr =
+      kernel_memory()->SystemHeapAlloc(sizeof(dummy_achievement_text));
```
gibbed, line 602: You're allocating heap memory here, any dummy text should be within the the buffer provided to the function, and guest pointers should point to that.

gibbed (changes requested):

src/xenia/kernel/xam/xam_user.cc
Gliniak: I kindly disagree.
It depends on flags provided to this function.

If you only provides flag 0x20 then you should get pointer from SPA to certain label. (that is why I had to add this allocation)

In other cases (iirc flags 0x2, 0x4, 0x8) yes, there we have space in provided buffer to store these strings and they're stored right after all X_ACHIEVEMENTS_DETAILS structures.

gibbed: What's "SPA"? If the flags control strings being present, why would you not have space for them in the user-provided buffer?

gibbed: If it really needs empty space outside of the user-provided buffer for some reason (why wouldn't the pointers just be NULL if there's nothing?), then the allocation should be moved to when the module is created, not for every call to creating the enumerator.

Gliniak: SPA contains information about achievements, stats etc. and this is loaded into memory while game starts. 
Emoose knows more about how everything works around it as they wrote code for it long time ago.

I can agree that we can remove it from it just in case, when game abuses this function. 
The pointer cannot be null as games assume that this is always loaded correctly (for example Red Dead Redemption does that)

gibbed: Ah ok. You previously said RDR passes every flag possible (-1). Shouldn't that mean it is also providing buffer space for strings? Or is it not?

Gliniak: Yeah it does. I haven't checked how it is handled, but I guess it contains space to store these strings after achievements structure. 
However header still expect that we will pass pointer to location where it is stored instead of null.

Game doesn't have direct access to these strings. It takes pointer from this header.

gibbed: Yeah that's what I mean, the expectation would be that the strings would get copied after the achievement structs, and the structures would be filled out to pointers to these locations in the buffer as normal?

Gliniak: Yes exactly, but only if you provide these specific flags (0x2, 0x4, 0x8 or 0x4, 0x8, 0x10)

gibbed (approved):

Dabur00:
sorry, but how do I get this fix for my version of Xenia?

PR #1746, opened 2021-01-25
[Kernel/XAM] XamEnumerate: Ignore clearing provided buffer

Gliniak:
It looks like XamEnumerate shouldn't care about provided buffer_size.

Disassembling xam.xex provide info that it only clears 0x1C bytes
image

Out of curiosity I patched one of the games that I have to provide custom value.
I checked 3 values:
- 0x10000
- 0x100000
- 0x20000000

For each of these values game went further just fine, which might suggest that XamEnumerate doesn't care about provided size


gibbed:
The clearing of the buffer is more of a sanity check thing. Is it negatively affecting anything?

Gliniak:
Yes it does. Sometimes games can write arbitrary value (and that's what I did, but on bigger scale). In such case Xenia crashes, but on console game works normally.

Gliniak:
Closed!
Reason: It will be replaced soon, by different PR

PR #1747, opened 2021-01-28
[Kernel] Implemented: XeKeysHmacShaUsingKey

Gliniak:
It was quite trivial to implement so no description here. 🐑 

# Affected Titles:
- Forza Motorsport 3 - Now doesn't return tampered error

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
@@ -660,6 +660,28 @@ dword_result_t XeKeysObscureKey(lpvoid_t input, lpvoid_t output) {
 }
 DECLARE_XBOXKRNL_EXPORT1(XeKeysObscureKey, kNone, kImplemented);
 
+dword_result_t XeKeysHmacShaUsingKey(lpvoid_t obscured_key, lpvoid_t inp_1,
+                                     dword_t inp_1_size, lpvoid_t inp_2,
+                                     dword_t inp_2_size, lpvoid_t inp_3,
+                                     dword_t inp_3_size, lpvoid_t out,
+                                     dword_t out_size) {
+  if (!obscured_key) {
+    return X_STATUS_INVALID_PARAMETER;
+  }
+
+  uint8_t key[16];
+
+  // Deobscure key
+  XECRYPT_AES_STATE aes;
+  XeCryptAesKey(&aes, (uint8_t*)xe_key_obfuscation_key);
+  XeCryptAesEcb(&aes, obscured_key, key, 0);
+
+  XeCryptHmacSha(key, 0x10, inp_1, inp_1_size, inp_2, inp_2_size, inp_3,
+                 inp_3_size, out, out_size);
+  return X_STATUS_SUCCESS;
+}
+DECLARE_XBOXKRNL_EXPORT1(XeKeysHmacShaUsingKey, kNone, kImplemented);
+
 void RegisterCryptExports(xe::cpu::ExportResolver* export_resolver,
                           KernelState* kernel_state) {}
 


emoose:
Pretty weird that they didn't use XeCryptHmacSha instead, XeKeys stuff usually involves keyvault things afaik.

Has this been tried with Forza saves from an actual 360? I did see HV was using keyvaults KEY_OBFUSCATION_KEY inside HvxKeysHmacSha under some condition, not sure if that actually applies here or not though, but maybe worth looking into if there's any issues with actual 360 data.

Gliniak:
```d> F8000044 XeKeysObscureKey(7022F500, 7022F530)
d> F8000044 XeKeysObscureKey(7022F4F0, 7022F540)
d> F8000044 XeKeysHmacShaUsingKey(7022F530, 7022F4D4, 00000014, 405FE8B0, 000015D0, 00000000, 00000000, 7022F510, 00000014)
d> F8000044 XeKeysAesCbcUsingKey(7022F540, 405FE8B0, 000015D0, 405FE8B0, 7022F4D8, 00000000)```

That's from where they're taking key for that function, however there is no visible usecase of input.
I can dump saves from console, but I think that they will not work due to some checks. Maybe additional profile GUID check or something 😞

**Comment by Triang3l** [comment]
XeKeysAesCbcUsingKey deobscures the key, shouldn't XeKeysHmacShaUsingKey do that too?

**Comment by emoose** [comment]
Yeah seeing as it's named similarly to the AesCbc func and both are using ObscureKey beforehand it's likely that it should, kinda hard to trace the call from kernel -> HV to make sure though, but there's definitely code in HvxKeysHmacSha that de-obscures like the AesCbc func.

Maybe try adding the same de-obscure code from AesCbc and see if Forza still works fine with it, hopefully it'll help get actual 360 saves to work once more profile/XUID stuff is added.

**Review by Triang3l — CHANGES_REQUESTED** [review]
Adding a change request for the comments about deobsuring.

**Comment by Gliniak** [comment]
> Yeah seeing as it's named similarly to the AesCbc func and both are using ObscureKey beforehand it's likely that it should, kinda hard to trace the call from kernel -> HV to make sure though, but there's definitely code in HvxKeysHmacSha that de-obscures like the AesCbc func.
> 
> Maybe try adding the same de-obscure code from AesCbc and see if Forza still works fine with it? Hopefully it'll help get actual 360 saves to work once more profile/XUID stuff is added.

I added code from from ``XeKeysAesCbcUsingKey`` that deobscured key and it seems that it still works. Nice. I was curious what's going on with console saves, but they're not usable (as expected). I will push this in few minutes

File: src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc [hunk 1]
```diff
@@ -660,6 +660,27 @@ dword_result_t XeKeysObscureKey(lpvoid_t input, lpvoid_t output) {
...
 
+dword_result_t XeKeysHmacShaUsingKey(lpvoid_t obscured_key, lpvoid_t inp_1,
+                                     dword_t inp_1_size, lpvoid_t inp_2,
+                                     dword_t inp_2_size, lpvoid_t inp_3,
+                                     dword_t inp_3_size, lpvoid_t out,
+                                     dword_t out_size) {
+  if (obscured_key) {
+    uint8_t key[16];
+
+    // Deobscure key
```
Triang3l, line 671: Since this may be needed in other XeKeys functions if there are any, would be nice to move this to a `static` or `inline` function, but not necessary for this pull request.

src/xenia/kernel/xboxkrnl/xboxkrnl_crypt.cc
```diff
@@ -660,6 +660,27 @@ dword_result_t XeKeysObscureKey(lpvoid_t input, lpvoid_t output) {
...
+    XECRYPT_AES_STATE aes;
+    XeCryptAesKey(&aes, (uint8_t*)xe_key_obfuscation_key);
+    XeCryptAesEcb(&aes, obscured_key, key, 0);
+
+    XeCryptHmacSha(key, 0x10, inp_1, inp_1_size, inp_2, inp_2_size, inp_3,
+                   inp_3_size, out, out_size);
+    return X_STATUS_SUCCESS;
+  }
+  return X_STATUS_INVALID_PARAMETER;
```
Triang3l, line 680: It's better to handle the `!obscured_key` case in a small `if` in the beginning, and not to nest the normal behavior of the function.

Triang3l (changes requested):

PR #1751, opened 2021-02-01
[Kernel/XAM] Stub xam party functions

Gliniak:
After digging a while into code of: Sonic & All-Stars Racing Transformed
I figured out that game have very specific checks.
It checks if every required Xam function exists, if it doesn't then stores 0 instead of address causing game to crash.

Second crash is related to: ``XamPartyGetUserList``. It does very straightforward check if function returns exactly that code: 0x807D0003 (we don't have party support, so the best idea is to return this code)

image

Inside that dword is stored XAM function address and if it isn't at least mocked then it will fail fabulously
image

# Affected Titles:
 - Sonic & All-Stars Racing Transformed (still crashes before menu, but not on canary with PRs. Probably one of unmerged PRs resolves that too)


halotroop2288 (approved):

src/xenia/kernel/xam/xam_module_export_groups.inc
```diff
@@ -27,3 +27,4 @@ XE_MODULE_EXPORT_GROUP(xam, UI)
 XE_MODULE_EXPORT_GROUP(xam, User)
 XE_MODULE_EXPORT_GROUP(xam, Video)
 XE_MODULE_EXPORT_GROUP(xam, Voice)
+XE_MODULE_EXPORT_GROUP(xam, Party)
```
Triang3l, line 30: Did you mean Warty, Xarty, Yarty or Zarty?

Triang3l (changes requested):

src/xenia/kernel/xam/xam_module_export_groups.inc
Gliniak: Done 🦊

PR #1771, opened 2021-03-28
[XAM/Content] Prevent deletion of open packages

Gliniak:
Based on ``Top Spin 4`` behaviour.

Seems like game tries to delete package without closing it. Which causes issue as game tries to open files within that package later and cannot find files.

I tried to poke with it on console with edited .xex and found out that ``XamContentDelete`` in this case it is not executed sucessfully and save stay where it was.

# To summary: 
**Actual Behavior:** 
Open Package->Delete->Open File Fail

**New Behaviour:**
Open Package->Delete Fail->Open File Sucessful

### DIFF
--- a/src/xenia/kernel/xam/content_manager.h
+++ b/src/xenia/kernel/xam/content_manager.h
@@ -64,6 +64,11 @@ struct ContentData {
 
   ContentData() = default;
 
+  bool operator==(const ContentData& rhs) const {
+    return device_id == rhs.device_id && content_type == rhs.content_type &&
+           file_name == rhs.file_name;
+  }
+
   explicit ContentData(const XCONTENT_DATA& data) {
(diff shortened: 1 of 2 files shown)


src/xenia/kernel/xam/content_manager.h
```diff
@@ -150,6 +158,7 @@ class ContentManager {
                                std::vector<uint8_t> buffer);
   X_RESULT DeleteContent(const ContentData& data);
   std::filesystem::path ResolveGameUserContentPath();
+  bool IsContentOpen(const ContentData& data);
```
gibbed, line 161: Needs `const`ness.

gibbed (changes requested):

gibbed (approved):

PR #1792, opened 2021-05-03
Do not treat warnings as errors for third party projects

JoelLinn:
Enables us to get rid of the disable warnings flags for the third party projects.
We usually don't "own" that code so we shouldn't  be too picky about the warnings it emits.


JoelLinn:
Disable warning flags aren't removed from https://github.com/xenia-project/FFmpeg yet, so building with clang version < 11 is still broken (#1735) though.

Issue #1407, opened 2019-07-23
XContent package Display Name used to identify DLC in some games


emoose:
In Halo 4 it seems to check the packages Display Name from the package headers to identify certain DLCs, at least for the War Games/Spartan Ops DLCs that are installed from Disc 2.

For example you could have the Spartan Ops DLC extracted into content\4D530919\00000002\E25A9996DBF25D25C7C6235FFD30BAF4E5D543044D based on the filename of the package, but as the display name gets taken from the folder name the game will fail to recognise it.

Renaming the folder to "War Games" / "Spartan Ops Season 1.0" does let the game recognise them, but since it symlinks based on the folder name it'll then create a "Spartan Ops Season 1.0:" symlink, eg. accessing files like "Spartan Ops Season 1.0:\ff81_courtyard.map", this appears to work fine for now, but IMO isn't that optimal.

We should probably allow reading the display name from the StfsHeader struct, and maybe let extracted packages contain a headers.bin file or similar with that header inside (or a <foldername>.headers file in the parent dir, so that games don't accidentally try reading from it)

Supporting full non-extracted packages is also an option, Xenia already has pretty good support for STFS etc, and one of my older PRs did add support for loading from non-extracted packages too (I should probably split that away into a seperate PR...)

gibbed:
Question, is this display name actually required to be these values (the display name), or is this perhaps an issue where the content ID is simply too long for the game?

IE: for `War Games` would naming the directory `wg` work or any other nonsense (that isn't `War Games`) work?

emoose:
Looks like some games save systems also check against the display name of the save package, eg. Kameo seems to require the display name to equal the character name inside the save, and Halo 3 requires the display name to equal the signed in profiles gamertag.

Unfortunately doesn't seem that renaming folders works for these, so I guess they probably rely on the package file name remaining the same too.

@gibbed Need to double check, but I believe it checks the whole string against ones stored inside the XEX, though I'm not sure how localised display names would come into play like this, guess the XEX must also store those somewhere too.

x1nixmzeng:
Take look at `src/xenia/kernel/util/xdbf_utils.h` which is used for branding the game

PR #1793, opened 2021-05-05
[VFS/STFS] STFS structure improvements & CON support

emoose:
This brings some improvements to the STFS structures and fixes some STFS handling issues, mostly related to CON profile/gamesave packages:

- STFS structures have been reorganized & improved to use `xe::be<T>`/bitfields/accessors, and remove the need for `read()` functions.
- `BlockToOffsetSTFS` has been fixed to work with CON/non-read-only-flagged packages.
- `GetBlockHash` now checks whether each levels secondary hash-block should be used or not in packages that have them, and updated the algo used for finding hash block for a given data block.
- `XdbfLocale` moved to xbox.h as `XLanguage`, so that `XContentMetadata` accessors can make use of it.

The structure improvements are pretty minor, existing structs already mapped out the headers pretty well, just reorganized them a little and let them use `xe::be<T>`/accessors instead of needing to use a `read()` function, so now data can be read as/casted to `StfsHeader` etc directly, and changed some flags fields to use bitfield structs instead.
(also added structs for the license entries at 0x22C since it might be worth linking them up with `XamContentGetLicenseMask` / `XamContentCreateEx`'s license_mask_ptr param eventually, and added accessor funcs to access the per-language big-endian-unicode strings)

`BlockToOffsetSTFS` had a bug that seems to break all CON packages as the `block_shift` used by CONs would get pretty much cancelled out by mistake, I've fixed the `block_shift` and also renamed it/changed it to a multiply to make it more obvious what the purpose is.

`GetBlockHash` didn't really check for each levels secondary-block flag previously (there was some code that would make the caller use the secondary-block if a volume-descriptor flag was set, but that'd only be valid if there was just a single hash-level used, ie. 170 data blocks or less), which would prevent certain badly-formatted (but still valid) packages from loading properly.
Updated that code to walk up each hash level and check each secondary-block flag instead, but not sure if it's the most efficient way to handle it, so also added some code to cache each hash-table we read from to try to help reduce how many reads are needed.

Also updated the code for finding the hash block # for a given data block # too, old code seemed to work fine for LIVE/PIRS packages but didn't really handle CON files well, also only found the level0 hash table for a given block with no code for finding level1/2 hash blocks, changed this to use an algo close to what the kernel uses, with some changes to try make it a bit more clear about how it works.

Lastly `XdbfLocale` was renamed to `XLanguage` and moved to xbox.h so the accessors in `XContentMetadata` could make use of it (renamed because AFAIK locale is a different thing to languages, eg. XAM has a partner function for `XGetLanguage` named `XamGetLocale`, which seems to be country related going off the values in xam_locale.cc)

---

So far mostly all the CONs I've tested with seem to work fine now (luckily still have a bunch of badly-formatted packages from some old work on a profile-editing tool, heavily fragmented/corrupt/etc, many STFS tools would fail with these while X360 itself never had any trouble, looks like xenia-vfs-dump now also works fine with them)

Haven't been able to test with many XBLA packages yet, but the few I've tried (~700MB one + some smaller TU packages) still worked fine at least, only have a single SVOD/GoD package to test with too but didn't seem to cause any issues there, would appreciate if someone with access to more packages could test it out.

---

This doesn't help with letting Xenia/XamContent* actually see CON savegame/profile files though, even if Xenia could load from CONs there's still issues with writing back to them too (eg. a save loaded from CON package would fail when game tries to update it...)

In xenia-canary this was kinda worked around for profiles by making the profile-loader code extract the package automatically and read/write to the extracted folder instead, maybe something similar could be done with `StfsContainerDevice` itself too.

I've made a proof-of-concept [stfs-folders](https://github.com/emoose/xenia/commits/stfs-folders) branch that mostly works like that and it seems to work fine, just changes `StfsContainerDevice` so that it can extract the files next to the package & create/override itself with a `HostPathDevice`, with `ContentPackage` class updated to use `StfsContainerDevice` instead so that STFS packages can be seen by the XamContent* funcs.

With that any packages loaded via XamContent* should get the correct display-name/content-type/etc passed to the game, and if the package isn't read-only (such as gamesaves/profiles) it'll extract any contents and read/write to the extraction folder instead, packages created by Xenia will also have STFS headers written out for them automatically, so would probably help with #1407 & #1739 too (and makes #1408 no longer needed)

(package headers are read/written to "[package-path]" file, with "[package-path].data" folder used for the extraction folder/HostPathDevice, this follows SVODs convention of storing data files inside a "[package].data" folder next to the SVOD package/header file)

Didn't need as many changes as I thought to get working, but not really sure if this is the best way to handle this, would welcome any thoughts about it (maybe should make a separate draft PR for that after this one?)

### DIFF
--- a/src/xenia/vfs/devices/stfs_container_device.h
+++ b/src/xenia/vfs/devices/stfs_container_device.h
@@ -268,11 +268,13 @@ XEPACKEDSTRUCT(XContentMetadata, {
     char16_t description_ex_chars[kNumLanguagesV2 - kNumLanguagesV1][128];
   };
 
-  std::u16string display_name(uint32_t lang_id) const {
-    lang_id--;
+  std::u16string display_name(XLanguage language) const {
+    uint32_t lang_id = uint32_t(language) - 1;
+
     if (lang_id >= kNumLanguagesV2) {
       assert_always();
-      lang_id = 0;  // no room for this lang, read from english slot..
+      // no room for this lang, read from english slot..
+      lang_id = uint32_t(XLanguage::kEnglish) - 1;
     }
 
     const be<uint16_t>* str = 0;
@@ -284,17 +286,21 @@ XEPACKEDSTRUCT(XContentMetadata, {
     }
 
     if (!str) {
+      // Invalid language ID?
+      assert_always();
       return u"";
     }
 
     return load_and_swap<std::u16string>(str);
   }
 
-  std::u16string description(uint32_t lang_id) const {
-    lang_id--;
+  std::u16string description(XLanguage language) const {
+    uint32_t lang_id = uint32_t(language) - 1;
+
     if (lang_id >= kNumLanguagesV2) {
       assert_always();
-      lang_id = 0;  // no room for this lang, read from english slot..
+      // no room for this lang, read from english slot..
+      lang_id = uint32_t(XLanguage::kEnglish) - 1;
     }
 
     const be<uint16_t>* str = 0;
@@ -306,6 +312,8 @@ XEPACKEDSTRUCT(XContentMetadata, {
     }
 
     if (!str) {
+      // Invalid language ID?
+      assert_always();
       return u"";
     }
 
@@ -320,11 +328,13 @@ XEPACKEDSTRUCT(XContentMetadata, {
     return load_and_swap<std::u16string>(title_name_raw);
   }
 
-  bool set_display_name(uint32_t lang_id, const std::u16string_view value) {
-    lang_id--;
+  bool set_display_name(XLanguage language, const std::u16string_view value) {
+    uint32_t lang_id = uint32_t(language) - 1;
+
     if (lang_id >= kNumLanguagesV2) {
       assert_always();
-      lang_id = 0;  // no room for this lang, store in english slot..
+      // no room for this lang, store in english slot..
+      lang_id = uint32_t(XLanguage::kEnglish) - 1;
     }
 
     char16_t* str = 0;
@@ -336,6 +346,8 @@ XEPACKEDSTRUCT(XContentMetadata, {
     }
 
     if (!str) {
+      // Invalid language ID?
+      assert_always();
       return false;
     }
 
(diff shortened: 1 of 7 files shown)


src/xenia/vfs/devices/stfs_container_device.h
```diff
@@ -24,144 +27,390 @@ namespace vfs {
...
+
+  std::u16string publisher() const {
+    return load_and_swap<std::u16string>(publisher_raw);
+  }
+
+  std::u16string title_name() const {
+    return load_and_swap<std::u16string>(title_name_raw);
+  }
+
+  bool display_name(XLanguage language, const std::u16string_view value) {
```
gibbed, line 325: If I'm understanding the context here, if this is updating values shouldn't they be ie `set_display_name()` rather than just `display_name()`?

gibbed (changes requested):

src/xenia/vfs/devices/stfs_container_device.h
```diff
@@ -24,144 +27,390 @@ namespace vfs {
...
+    if (language >= XLanguage::kMaxLanguages) {
+      assert_always();
+      // no room for this lang, read from english slot..
+      language = XLanguage::kEnglish;
+    }
+
+    uint32_t lang_id = uint32_t(language) - 1;
+
+    const be<uint16_t>* str = 0;
+    if (lang_id >= 0 && lang_id < 9) {
```
Gliniak, line 277: I'm not a big fan of magic numbers over there.
Same thing with few lines below

src/xenia/vfs/devices/stfs_container_device.h
emoose: Sure, just named them like that to make them sorta feel like actual fields, but I can rename them, would probably help make code that uses them a little clearer too.

src/xenia/vfs/devices/stfs_container_device.h
emoose: Ah I did mean to do something about that but forgot to go over it, I guess since they're `metadata_version` related I could change them to `kMaxLanguagesV1`/`kMaxLanguagesV2` instead.

emoose:
Updated with requested changes

gibbed (approved):

gibbed:
@emoose Draft PR for your folders idea is fine, though I'm not a fan of the folders thing. Want to eventually get non-mmap version of stfs stuff in so we can do mounts of content from non-host file sources, for things like #1247.

gibbed:
These changes fix mounting of YouTube's container which was breaking badly before. I know of at least one other title that was exploding previously as well, will test eventually.

PR #1797, opened 2021-05-05
Improve GCC/Std compatibility and Linux dev experience

JoelLinn:
This fixes a bunch of issues when compiling with GCC (10).

One issue still left:

* `xe::gpu::RenderTargetCache::RenderTargetKey` initializes its key value to `0` and therefore has an implicit constructor. It is used in a union and thus throws a warning. According to @Triang3l there will be changes to it "soon", so keeping this untouched for now.
```
../src/xenia/gpu/render_target_cache.h:549:25: error: member ‘xe::gpu::RenderTargetCache::RenderTargetKey xe::gpu::RenderTargetCache::OwnershipRange::<unnamed union>::<unnamed struct>::host_depth_render_target_float24’ with constructor not allowed in anonymous aggregate
  549 |         RenderTargetKey host_depth_render_target_float24;
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

* ~~GCC throws a warning during LTO (does not show if LTO is disabled). It is triggered by the xma context struct load/store methods. Bug filed with GCC~~ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100801 Added workaround
```
../src/xenia/base/memory.h: In function ‘copy_and_swap.constprop’:
../src/xenia/base/memory.cc:105: warning: iteration 4611686018427387903 invokes undefined behavior [-Waggressive-loop-optimizations]
  105 |     dest[i] = byte_swap(src[i]);
      | 
../src/xenia/base/memory.cc:104: note: within this loop
  104 |   for (; i < count; ++i) {  // handle residual elements
      | 
../src/xenia/base/memory.cc:125: warning: iteration 4611686018427387903 invokes undefined behavior [-Waggressive-loop-optimizations]
  125 |     dest[i] = byte_swap(src[i]);
      | 
../src/xenia/base/memory.cc:124: note: within this loop
  124 |   for (; i < count; ++i) {  // handle residual elements
      | 

```


src/xenia/gpu/render_target_cache.h
```diff
@@ -230,8 +230,8 @@ class RenderTargetCache {
     };
     uint32_t key = 0;
     struct Hasher {
-      size_t operator()(const RenderTargetKey& key) const {
-        return std::hash<uint32_t>{}(key.key);
+      size_t operator()(const RenderTargetKey& renderTargetKey) const {
```
Triang3l, line 233: We are using `render_target_key` naming, not `renderTargetKey`.

src/xenia/gpu/texture_dump.cc
```diff
@@ -59,17 +59,17 @@ void TextureDump(const TextureInfo& src, void* buffer, size_t length) {
   switch (src.format) {
     case xenos::TextureFormat::k_DXT1: {
       dds_header.pixel_format.flags = 0x4u;
-      dds_header.pixel_format.fourcc = '1TXD';
+      dds_header.pixel_format.fourcc = multi_char('1', 'T', 'X', 'D');
```
Triang3l, line 62: I think it would be more convenient to have little-endian and big-endian versions of `multi_char`. This is supposed to mean "DXT1", not "1TXD".

src/xenia/base/memory.h
```diff
@@ -441,6 +441,10 @@ inline void store_and_swap<std::u16string>(void* mem,
   return store_and_swap<std::u16string_view>(mem, value);
 }
 
+inline constexpr uint32_t multi_char(char a, char b, char c, char d) {
```
Triang3l, line 444: This name seems kinda ambiguous to me — specifically, this generates a four-character code, aka FourCC.

Triang3l, line 444: `char` is signed on some platforms — if a value of over 128 was passed here, would it be treated as 0x0000008… in the formula below, or as 0xFFFFFF8…? I'm not sure how promotion is going to work here. But I'd cast them explicitly to `uint32_t` below.

src/xenia/base/testing/threading_test.cc
```diff
@@ -154,7 +154,7 @@ TEST_CASE("HighResolutionTimer") {
   // Time the actual sleep duration
   {
     const auto interval = 50ms;
-    std::atomic<uint64_t> counter;
+    std::atomic<uint64_t> counter = 0;
```
Triang3l, line 157: `counter(0)` should be enough — the initialization doesn't have to be atomic (the assignment would be, I think?)

src/xenia/gpu/ucode.h
```diff
@@ -145,7 +145,7 @@ enum class AllocType : uint32_t {
 };
 
 // Instruction data for ControlFlowOpcode::kExec and kExecEnd.
-struct ControlFlowExecInstruction {
+XEPACKEDSTRUCT(ControlFlowExecInstruction, {
```
Triang3l, line 148: Alignment of 4, not 1, is safe in these structures — no need to make them `XEPACKEDSTRUCT`. In most cases, `XEPACKEDSTRUCT` can be removed from bitfields overall.

src/xenia/gpu/xenos.h
```diff
@@ -1070,7 +1073,7 @@ XEPACKEDUNION(xe_gpu_texture_fetch_t, {
     uint32_t pitch : 9;                                  // +22 byte_pitch >> 5
     uint32_t tiled : 1;                                  // +31
 
-    TextureFormat format : 6;           // +0 dword_1
+    uint32_t format : 6;                // +0 dword_1, enum TextureFormat
```
Triang3l, line 1076: When a field has a type, it's preferred to use it instead for cleanliness — we're using types like Endian, MsaaSamples, TextureFormat wherever possible. It's already `: uint32_t`, it was fine that way.

src/xenia/gpu/texture_info.cc
```diff
@@ -32,7 +32,7 @@ bool TextureInfo::Prepare(const xe_gpu_texture_fetch_t& fetch,
 
   auto& info = *out_info;
 
-  info.format = fetch.format;
+  info.format = static_cast<TextureFormat>(fetch.format);
```
Triang3l, line 35: Not needed if the change to `xe_gpu_texture_fetch_t` is reverted.

Triang3l (changes requested):

src/xenia/gpu/ucode.h
JoelLinn: The struct is used in a packed union and therefore GCC complains:
`ignoring packed attribute because of unpacked non-POD field`

So I made them packed but encapsulated the bit field in another struct. I assumed the packed attribute is not inherited by anonymous structs, is this wrong?

src/xenia/gpu/xenos.h
Triang3l: It's better to remove kUnknown from TextureFormat — it's only used in 4 places, 2 deprecated, 1 where it statically can't be obtained (and can be replaced with `? D24S8 : D24FS8` at the location of the call).

src/xenia/gpu/ucode.h
Triang3l: It's safe to remove `XEPACKEDSTRUCT` and `XEPACKEDUNION` from everywhere in `ucode.h`. I'm not sure about the inner structs, probably yes, but it's weird that it's used at all in this case then.

src/xenia/gpu/texture_dump.cc
JoelLinn: mh. we have be template to convert to host endian. Likewise bit shifting always happens in host endian.
how exactly would these endian versions come into play?

Triang3l: There probably needs to be a function for generating a FourCC ('A', 'B', 'C', 'D') that would match bytes `ABCD` if stored to a file as uint32_t on the host. With guest file structures it may be a bit confusing though.

JoelLinn: I think that is what the multi_char function does right now, isnt it?

JoelLinn: I added a new struct `fourcc_t`
It has methods to retrieve `host`, `be` and `le` variants (`host==le` on x64 of course)
Magic constant now are of that type. Some constants are used in different endians so that makes it easier.
Internally, fourccs are stored as `be`. This way compiler should hopefully be able to short circuit assignments/comparisons with `be<uint32_t>` values.

Triang3l: The idea was to internally accept the characters in the natural order (DXT1 rather than 1TXD), but to be able to do something like
```
DDSHeader dds_header;
dds_header.four_cc = make_fourcc_le('D', 'X', 'T', '1');
```
where `make_fourcc_le` would be doing `'D'|('X'<<8)|('T'<<16)|('1'<<24)` on a little-endian host and `'1'|('T'<<8)|('X'<<24)|('D'<<24)` on a big-endian one, and then `fwrite` the whole `dds_header`. Using a separate FourCC would make such `fwrite` impossible if `fourcc_t` is in `DDSHeader`, or require an unneeded trip through the big-endian representation if it's `uint32_t`.

gibbed:
`xe::kernel::RegisterExport` is largely used in a macro so renaming to avoid ambiguity would be fine.

JoelLinn:
> `xe::kernel::RegisterExport` is largely used in a macro so renaming to avoid ambiguity would be fine.

But how would the macro differentiate them? (Without having two differently named macros)

gibbed:
@JoelLinn We already have different macros per argument list.

gibbed:
Actually I misunderstood the problem now that I look at it more closely.

JoelLinn:
`pkg-config` change broke `clang` linking with `cmake`.
Will update `premake-cmake` submodule when https://github.com/Enhex/premake-cmake/pull/8 is merged

JoelLinn:
* Waiting for another regression in upstream premake-cmake https://github.com/Enhex/premake-cmake/pull/9
* There is an issue with GCC LTO going crazy in constant propagation https://gcc.gnu.org/pipermail/gcc-help/2021-May/140325.html still hoping for an answer
* Fixed some typos in early commits, hence big rebase

PR #1807, opened 2021-05-15
[VFS/STFS] Use xe::filesystem::OpenFile instead of MappedMemory, other small improvements

emoose:
Changes StfsContainerDevice to use `xe::filesystem::OpenFile` / `FILE*` instead of MappedMemory, maybe can help with nested packages (guess we might need to go a little further & add support for loading from vfs::Entry or something for that though?)

There's a few other minor improvements that the commit msgs go into, adding names to some anonymous structs, making the vfs::Device functions like `available_allocation_units()` return something more useful, adding `StfsDirectoryEntry/StfsDirectoryBlock` structs so we can read the data in one go without needing to read each seperate field, etc.

Moved the x360-specific structs to stfs_xbox.h so they aren't filling stfs_container_device.h any more, should make it more obvious these are structs X360 itself uses too, not something that we defined.

---

Like my previous STFS PR I haven't been able to test with that many packages, the ones I've tried didn't seem to have any issues, but if anyone is able to test with more packages/SVODs it'd be appreciated.

(also I wonder if changing MappedMemory out for OpenFile might affect load times at all? Didn't notice anything myself, but I didn't really look into it much)

---

I made these changes while working on adding support for writing out packages to StfsContainerDevice, more info about that here: https://github.com/xenia-project/xenia/pull/1793#issuecomment-837699003 / https://github.com/emoose/xenia/commits/stfs-writer

Some details about that:

> Hash-tables are simplified down into chains (vectors of block-numbers) that the StfsContainerEntry uses to know where to seek for reads/writes, main StfsContainerDevice code handles converting those chains to/from hash-table entries, allocating new blocks to add to chains, freeing unused blocks, etc.

> Hashes of L0 hash table and above are checked when the hash entries are read from them (STFSGetDataHashTable/STFSGetHashTable), so we know whether to trust the next-block info or not (some STFS packages out there sadly have corrupt tables, with LIVE/PIRS this is easy to salvage though since blocks are almost always contiguous there), maybe in future this can be expanded to check the actual data hashes too.

> Hash entry writing works differently though, instead of updating hashes whenever we update a hash-entry/block we can instead just mark whatever data block is being modified as dirty, then when the package is closed STFSFlush() gets called, which writes out the STFS headers/directories and recalculates hashes for any dirty blocks (L0 -> L1 -> L2)

> Not sure if it's worth changing this or not, making it similar to reading would likely mean a lot of wasted time calculating hashes that are quickly overwritten, but maybe changing it so that STFSFlush() is called whenever a packaged files handle is closed would be a good start.

Might be a while before I have that ready though so figured this might be worth PRing seperately, but if anyone has any thoughts about that STFS writing code I'd be happy to hear them!


gibbed:
Are we losing any significant performance benefit by moving to typical file reads?

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -207,76 +220,99 @@ StfsContainerDevice::Error StfsContainerDevice::ReadSVOD() {
...
 
   // Parse the root directory
-  uint8_t* magic_block = data + magic_offset_;
-  uint32_t root_block = xe::load<uint32_t>(magic_block + 0x14);
-  uint32_t root_size = xe::load<uint32_t>(magic_block + 0x18);
-  uint32_t root_creation_date = xe::load<uint32_t>(magic_block + 0x1C);
-  uint32_t root_creation_time = xe::load<uint32_t>(magic_block + 0x20);
+  xe::filesystem::Seek(svod_header, magic_offset + 0x14, SEEK_SET);
+
+  uint32_t root_block;
```
gibbed, line 303: Perhaps a locally defined struct to read into, as opposed to a batch set of variables?

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -182,23 +184,34 @@ Entry* StfsContainerDevice::ResolvePath(const std::string_view path) {
...
+  xe::filesystem::Seek(header_file, 0L, SEEK_END);
+  files_total_size_ = xe::filesystem::Tell(header_file);
+  xe::filesystem::Seek(header_file, 0L, SEEK_SET);
+
+  if (sizeof(StfsHeader) > files_total_size_) {
+    return Error::kErrorTooSmall;
+  }
+
+  // Read header & check signature
+  fread(&header_, sizeof(StfsHeader), 1, header_file);
```
gibbed, line 198: `fread`s (not just this one) could probably use return value checking to ensure the count matches what we expect.

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -301,16 +337,27 @@ StfsContainerDevice::Error StfsContainerDevice::ReadEntrySVOD(
...
-  uint32_t data_block = xe::load<uint32_t>(data + 0x04);
-  uint32_t length = xe::load<uint32_t>(data + 0x08);
-  uint8_t attributes = xe::load<uint8_t>(data + 0x0C);
-  uint8_t name_length = xe::load<uint8_t>(data + 0x0D);
-  auto name_buffer = reinterpret_cast<const char*>(data + 0x0E);
-  auto name = std::string(name_buffer, name_length);
+  auto file = files_.at(entry_file);
+  xe::filesystem::Seek(file, entry_address, SEEK_SET);
+
+  uint16_t node_l;
```
gibbed, line 344: Perhaps a locally defined struct to read into, as opposed to a batch set of variables?

gibbed (changes requested):

emoose:
Will update with requested changes soon, need to test with some packages from my box first.

BTW does anyone know if `#pragma pack(push, 1)` is available on GCC/Clang, and would that be better than using `XEPACKEDSTRUCT`?

Googling around shows they apparently should support it, but I don't know if that's something that has to be specially enabled or something. I'm assuming there's a reason we have `XEPACKEDSTRUCT` instead of using pragma?

JoelLinn:
> BTW does anyone know if `#pragma pack(push, 1)` is available on GCC/Clang, and would that be better than using `XEPACKEDSTRUCT`? Seems VS has some Intellisense issues with `XEPACKEDSTRUCT`, and sounds like it might be breaking GCC pretty badly too...

Very good point, I actually didn't know that existed. I have been working on (re-)introducing GCC compatibility (#1797) but gcc doesn't like the packed attribute when non POD types are involved (was able to fix some but your last PR basically made gcc unfixable because of that)
Looks like the `#pragma pack(push, 1)` is indeed [supported by all three compilers ](https://godbolt.org/z/bfoazq3dE) and for some reason GCC now works better.

@gibbed  I think we might just as well phase out our macros(?)

Didn't check yet if that also fixes the issues with GCC for `unions` but it's at least something

emoose:
Updated with req'd changes, also fixed some parts to use references instead of copying (`auto file = files_.at(0)` -> `auto& file`), and swapped out `XEPACKEDSTRUCT`s for `#pragma pack` instead.

Hopefully might help with the issues GCC was having, but I think the `be<uint16_t>`/`char16_t` unions could still have problems there, in my stfs-writer branch I noticed sometimes those unions would stop copy-constructor being generated for some reason, maybe something similar is happening there too?

gibbed:
@JoelLinn Instead of phasing out the macro, wouldn't it be wiser to flip it to doing what GCC can support?

(this wouldn't be an expected change for this PR)

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -207,84 +222,128 @@ StfsContainerDevice::Error StfsContainerDevice::ReadSVOD() {
...
-  uint32_t root_creation_date = xe::load<uint32_t>(magic_block + 0x1C);
-  uint32_t root_creation_time = xe::load<uint32_t>(magic_block + 0x20);
-  uint64_t root_creation_timestamp =
-      decode_fat_timestamp(root_creation_date, root_creation_time);
-
-  auto root_entry = new StfsContainerEntry(this, nullptr, "", &mmap_);
+  xe::filesystem::Seek(svod_header, magic_offset + 0x14, SEEK_SET);
+
+  struct {
+    uint32_t root_block;
```
gibbed, line 322: Naming. Why retain `root_` prefix here within the structure?

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -152,22 +144,32 @@ StfsContainerDevice::Error StfsContainerDevice::MapFiles() {
   }
 
   for (size_t i = 0; i < fragment_files.size(); i++) {
-    auto file = fragment_files.at(i);
-    auto path = file.path / file.name;
-    auto data = MappedMemory::Open(path, MappedMemory::Mode::kRead);
-    if (!data) {
+    auto& fragment = fragment_files.at(i);
```
gibbed, line 147: Unless I'm mistaken, `const`-ness is possible here?

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -207,84 +222,128 @@ StfsContainerDevice::Error StfsContainerDevice::ReadSVOD() {
...
+      magic_offset = 0x12000;
+
+      // Check for XSF Header
+      const char* XSF_MAGIC = "XSF";
+      xe::filesystem::Seek(svod_header, 0x2000, SEEK_SET);
+      if (fread(magic_buf, 1, 3, svod_header) != 3) {
+        XELOGE("ReadSVOD failed to read SVOD XSF magic at 0x2000");
+        return Error::kErrorReadError;
+      }
+      if (memcmp(magic_buf, XSF_MAGIC, 3) == 0) {
```
gibbed, line 276: `std::` prefix.

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -207,84 +222,128 @@ StfsContainerDevice::Error StfsContainerDevice::ReadSVOD() {
...
-      base_offset_ = 0x0000;
-      magic_offset_ = 0x2000;
+    xe::filesystem::Seek(svod_header, 0x2000, SEEK_SET);
+    if (fread(magic_buf, 1, countof(magic_buf), svod_header) !=
+        countof(magic_buf)) {
+      XELOGE("ReadSVOD failed to read SVOD magic at 0x2000");
+      return Error::kErrorReadError;
+    }
+
+    if (memcmp(magic_buf, MEDIA_MAGIC, 20) == 0) {
```
gibbed, line 245: `std::` prefix.

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -207,84 +222,128 @@ StfsContainerDevice::Error StfsContainerDevice::ReadSVOD() {
...
-      XELOGE(
-          "SVOD is not a single file, but the magic block was found at "
-          "0xD000.");
+      xe::filesystem::Seek(svod_header, 0xD000, SEEK_SET);
+      if (fread(magic_buf, 1, countof(magic_buf), svod_header) !=
+          countof(magic_buf)) {
+        XELOGE("ReadSVOD failed to read SVOD magic at 0xD000");
+        return Error::kErrorReadError;
+      }
+      if (memcmp(magic_buf, MEDIA_MAGIC, 20) == 0) {
```
gibbed, line 292: `std::` prefix.

src/xenia/vfs/devices/stfs_container_device.cc
```diff
@@ -207,84 +222,128 @@ StfsContainerDevice::Error StfsContainerDevice::ReadSVOD() {
...
-    // 0x2000 is from hash tables. In most cases, these will be STFS, not SVOD.
-
-    base_offset_ = 0xB000;
-    magic_offset_ = 0xD000;
-
-    // Check for single file system
-    if (header_.metadata.data_file_count == 1) {
-      svod_layout_ = SvodLayoutType::kSingleFile;
-      XELOGI("SVOD is a single file. Magic block present at 0xD000.");
+    if (memcmp(magic_buf, MEDIA_MAGIC, 20) == 0) {
```
gibbed, line 261: `std::` prefix.

gibbed (changes requested):

JoelLinn:
> @JoelLinn Instead of phasing out the macro, wouldn't it be wiser to flip it to doing what GCC can support?
> 
> (this wouldn't be an expected change for this PR)

I thought we couldn't use `#pragma` in the macro.
I was today years old when I learned that `_Pragma("...")` is available with GCC/Clang similar to MSVCs `__pragma( .... )`
I will address this in the GCC PR. 
Note that `#pragma pack()` is already used a number of times in master right now

emoose:
Updated with requested changes

gibbed (approved):

PR #1809, opened 2021-05-15
[Build] Support building with the standalone VS Build Tools

amine4567:
It has been possible for some time now to install the VS Build tools for C++ without installing the VS IDE. However, the current xenia build script doesn't support this because vswhere only searches for an installation of the VS IDE by default. This is easily fixable by adding `-products Microsoft.VisualStudio.Product.BuildTools` to the vswhere command and would work whether the VS IDE is installed or not.

### DIFF
--- a/xenia-build
+++ b/xenia-build
@@ -74,7 +74,16 @@ def import_vs_environment():
     install_path = None
     env_tool_args = None
 
-    vswhere = subprocess.check_output('third_party/vswhere/vswhere.exe -version "[15,)" -latest -prerelease -format json -utf8 -products Microsoft.VisualStudio.Product.BuildTools', shell=False, universal_newlines=True, encoding="utf-8")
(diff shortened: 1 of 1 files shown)


gibbed:
Your change is causing `vswhere` to fail to locate VS on AppVeyor (tripping a bug in old logic that needs to be fixed, unrelated to this PR), so it's not sufficient to simply look for `Microsoft.VisualStudio.Product.BuildTools`.

amine4567:
I misunderstood the products parameter of vswhere, I thought the BuildTools product is a subset of the VS IDEs installations. I made the necessary change to target all the possible 3 versions of VS IDE plus the standalone BuildTools. I took the list of product ids from here: https://aka.ms/vs/workloads

gibbed:
👍🏻

PR #1812, opened 2021-05-18
[XAM] Implemented: XStaticContentEnumerator

Gliniak:
Based on Rick's ``AchievementEnumerator``.

I had to split ``ContentEnumerator`` into separate entity because: 
1. It will allow to implement custom logic for that specific enumerator
2. Sometimes games instead of providing _buffer_size_ in bytes they provide some arbitrary value (probably max amount of files).

I probably forgot about something or something can be simplified, so please point that out. Thanks

Affected games:
- Splatterhouse
- ¯\_(ツ)_/¯ (We will see soon)

### DIFF
--- a/src/xenia/kernel/xam/xam_content.cc
+++ b/src/xenia/kernel/xam/xam_content.cc
@@ -31,6 +31,58 @@ namespace xe {
 namespace kernel {
 namespace xam {
 
+class XStaticContentEnumerator : public XEnumerator {
+ public:
+  XStaticContentEnumerator(KernelState* kernel_state,
+                           size_t items_per_enumerate)
+      : XEnumerator(kernel_state, items_per_enumerate, sizeof(XCONTENT_DATA)) {}
+
+  void AppendItem(ContentData item) { items_.push_back(item); }
+  size_t item_count() const { return items_.size(); }
+
+  uint32_t WriteItems(uint32_t buffer_ptr, uint8_t* buffer_data,
+                      uint32_t buffer_size, uint32_t* written_count) override {
+    size_t count =
+        std::min(items_.size() - current_item_, items_per_enumerate());
+
+    if (!count) {
+      return X_ERROR_NO_MORE_FILES;
+    }
+
+    size_t items_size = count * item_size();
+    size_t actual_buffer_size = buffer_size;
+
+    if (actual_buffer_size < item_size()) {
+      // Game provided max amount of items instead of used space
+      if (count * items_per_enumerate() > actual_buffer_size) {
+        return X_ERROR_INSUFFICIENT_BUFFER;
+      }
+      actual_buffer_size = items_size * items_per_enumerate();
+    }
+
+    if (items_size > actual_buffer_size) {
+      return X_ERROR_INSUFFICIENT_BUFFER;
+    }
+    memset(buffer_data, 0, items_size);
+
+    auto details = reinterpret_cast<XCONTENT_DATA*>(buffer_data);
+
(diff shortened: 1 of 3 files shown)


src/xenia/kernel/xam/xam_enum.cc
```diff
@@ -41,27 +41,8 @@ uint32_t xeXamEnumerate(uint32_t handle, uint32_t flags, lpvoid_t buffer_ptr,
...
-      // Known culprits:
-      //   Final Fight: Double Impact (saves)
-      XELOGW(
-          "Broken usage of XamEnumerate! buffer size={:X} vs actual "
-          "size={:X} (item size={:X}, items per enumerate={})",
-          buffer_size, actual_buffer_size, e->item_size(),
-          e->items_per_enumerate());
-    }
-
-    result = X_ERROR_INSUFFICIENT_BUFFER;
```
gibbed, line 58: You've removed this check, which I would prefer to keep.

src/xenia/kernel/xam/xam_content.cc
```diff
@@ -31,6 +31,64 @@ namespace xe {
...
+    if (items_size > actual_buffer_size) {
+      return X_ERROR_INSUFFICIENT_BUFFER;
+    }
+    memset(buffer_data, 0, items_size);
+
+    auto details = reinterpret_cast<XCONTENT_DATA*>(buffer_data);
+
+    for (size_t i = 0, o = current_item_; i < count; ++i, ++current_item_) {
+      const auto& item = items_[current_item_];
+      details[i].content_type = item.content_type;
```
gibbed, line 72: The `ContentData` type already has a helper function for (correctly) writing out to guest memory. You don't need to do it yourself here.

```cpp
item.Write(&details[i]);
```

src/xenia/kernel/xam/xam_enum.cc
```diff
@@ -41,27 +41,8 @@ uint32_t xeXamEnumerate(uint32_t handle, uint32_t flags, lpvoid_t buffer_ptr,
   } else if (!buffer_ptr) {
     result = X_ERROR_INVALID_PARAMETER;
   } else {
-    size_t needed_buffer_size = e->item_size() * e->items_per_enumerate();
-
-    uint32_t actual_buffer_size = buffer_size;
-    if (buffer_size == e->items_per_enumerate()) {
-      actual_buffer_size = static_cast<uint32_t>(needed_buffer_size);
-      // Known culprits:
-      //   Final Fight: Double Impact (saves)
```
gibbed, line 50: Are saves "content" here that your enumerator would correct Final Fight's behavior? I don't recall.

gibbed (changes requested):

Gliniak:
Thanks for the review. 
I see you have that "problematic" game. Would you mind to check if saving works correctly.

src/xenia/kernel/xam/xam_content.cc
```diff
@@ -31,6 +31,58 @@ namespace xe {
...
+
+    size_t items_size = count * item_size();
+    size_t actual_buffer_size = buffer_size;
+
+    if (actual_buffer_size < item_size()) {
+      // Game provided max amount of items instead of used space
+      if (count * items_per_enumerate() > actual_buffer_size) {
+        return X_ERROR_INSUFFICIENT_BUFFER;
+      }
+      actual_buffer_size = items_size * items_per_enumerate();
```
Gliniak, line 60: I'm not exactly sure if that logic is correct here. In terms of savefiles probably, but I don't know how it will behave with other content types

Gliniak:
Superseded by #1862 
I will check if my test case works with new version

PR #1820, opened 2021-05-24
[HID] Added option to turn off vibration

Gliniak:
Some games in x360 lib doesn't allow to turn off vibrations directly from the game. That option is for these cases. 
I tried it in two more ways.

1. To set profile setting related to vibration to 0
2. Tell that controller doesn't support vibrations

But they didn't worked 😢


src/xenia/kernel/xam/xam_input.cc
```diff
@@ -118,6 +120,10 @@ dword_result_t XamInputSetState(dword_t user_index, dword_t unk,
     actual_user_index = 0;
   }
 
+  if (!cvars::enable_vibration) {
+    vibration->left_motor_speed = 0;
+    vibration->right_motor_speed = 0;
+  }
```
Triang3l, line 126: Unbalanced blank lines.

src/xenia/kernel/xam/xam_input.cc
```diff
@@ -118,6 +120,10 @@ dword_result_t XamInputSetState(dword_t user_index, dword_t unk,
     actual_user_index = 0;
   }
 
+  if (!cvars::enable_vibration) {
+    vibration->left_motor_speed = 0;
```
Triang3l, line 124: This must not be modifying the original structure in the guest memory, it should be creating a copy in this case.

Triang3l (changes requested):

src/xenia/kernel/xam/xam_input.cc
```diff
@@ -118,6 +120,10 @@ dword_result_t XamInputSetState(dword_t user_index, dword_t unk,
...
+    vibration->right_motor_speed = 0;
+  }
   auto input_system = kernel_state()->emulator()->input_system();
   return input_system->SetState(user_index, vibration);
```
Triang3l, line 128: Ideally this should probably be handled by a proxy function in the HID subsystem, I think, also with the possibility of toggling it at runtime, zeroing the current speed when vibration is disabled mid-game from the UI, and remembering the speeds from the latest XamInputSetState call and applying it when vibration is re-enabled.

src/xenia/kernel/xam/xam_input.cc
```diff
@@ -20,6 +20,8 @@ namespace xe {
 namespace kernel {
 namespace xam {
 
+DEFINE_bool(enable_vibration, true, "Enable controller vibration", "HID");
```
Triang3l, line 23: `bool` implies "enable", there's no need for the "enable" word. Most config variable descriptions also end with a period.

src/xenia/kernel/xam/xam_input.cc
JoelLinn: I am unsure if it is necessary to have the rumble speed remembered. What guest do you now that has active rumble (at the same level) for more than a second? (while there is active user input)

src/xenia/kernel/xam/xam_input.cc
JoelLinn: +1, just "allocate" a new one on the stack (in the function scope)
You could also have the memcpy and pointer assignment inside the if block

Also, i suggest moving this code to `InputSystem::SetState`

JoelLinn (changes requested):
I don't think kernel is the right place to do this. hid sub-system makes more sense to me (see comments)

src/xenia/hid/input_system.cc
```diff
@@ -62,10 +64,14 @@ X_RESULT InputSystem::GetState(uint32_t user_index, X_INPUT_STATE* out_state) {
 X_RESULT InputSystem::SetState(uint32_t user_index,
                                X_INPUT_VIBRATION* vibration) {
   SCOPE_profile_cpu_f("hid");
-
+  X_INPUT_VIBRATION* modified_vibration = new X_INPUT_VIBRATION();
```
JoelLinn, line 67: This is a memory leak 🥲
Just do `X_INPUT_VIBRATION modified_vibration;` to allocate it on the stack and pass `&modified_vibration`

JoelLinn (changes requested):

src/xenia/hid/input_system.cc
Triang3l: Unless you're already adding the scale for the vibration, you can just `return` immediately from the function, by the way — you don't need to set vibration to 0 if you never set it to anything other than 0 anyway.

Gliniak: Smack me with a hammer for my stupidity... I dunno what I was thinking and why I thought it will be fine

JoelLinn: Well, if this var can someday be changed from gui and controller is currently rumbling, itll never stop iIrc

Triang3l: The GUI button can explicitly stop the current vibration.

src/xenia/kernel/xam/xam_input.cc
Triang3l: Environmental effects such as thunder possibly, big explosions in scripted scenes.

Triang3l: Overall, statically, in the XInput API, there are no requirements of how frequently the function needs to be called so that we can treat a specific maximum length as "negligible". By only dropping calls, we're introducing arbitrary behavior not present in the API specification, as well as unspecified constraints on arbitrarily chosen dynamic state (when the game performs the XInputSetState calls, and the point in time when the user has toggled vibration). A properly isolated implementation should store the desired vibration level, and perform the toggling with the event flow similar to that of a new keystroke arriving from the UI thread in the `winkey` HID backend.

halotroop2288 (approved):

PR #1821, opened 2021-05-25
[XAM] XamShowSigninUI: Fix issue with loading profile(s)

madmodder123:
This fixes an issue with loading profiles when selecting certain game modes (even if playing alone).

Specifically in Perfect Dark XBLA when you would try to load up the Combat Simulator, Co-operative, or Counter-operative game modes the title would display a loading message, and then just boot the user back to the main menu. 
https://github.com/xenia-project/game-compatibility/issues/316#issuecomment-379186790

The system expects the 4 player sign in UI to show up and then be submitted, but Xenia did not send an XNotify confirming that had happened, so the title behaves as if the UI had been closed. 

Example of expected behavior on a real console (see 3:35):
https://www.youtube.com/watch?v=nHkzTpqnNRM

I am not sure if any other titles are impacted.

### DIFF
--- a/src/xenia/kernel/xam/xam_user.cc
+++ b/src/xenia/kernel/xam/xam_user.cc
@@ -546,7 +546,12 @@ DECLARE_XAM_EXPORT1(XamUserAreUsersFriends, kUserProfiles, kStub);
 
 dword_result_t XamShowSigninUI(dword_t unk, dword_t unk_mask) {
   // Mask values vary. Probably matching user types? Local/remote?
-  // Games seem to sit and loop until we trigger this notification.
+
+  // To fix game modes that display a 4 profile signin UI (even if playing alone):
+  // XN_SYS_SIGNINCHANGED
+  kernel_state()->BroadcastNotification(0x0000000A, 1);
+  // Games seem to sit and loop until we trigger this notification:
+  // XN_SYS_UI (off)
   kernel_state()->BroadcastNotification(0x00000009, 0);
   return X_ERROR_SUCCESS;
 }


src/xenia/kernel/xam/xam_user.cc
```diff
@@ -546,7 +546,12 @@ DECLARE_XAM_EXPORT1(XamUserAreUsersFriends, kUserProfiles, kStub);
 
 dword_result_t XamShowSigninUI(dword_t unk, dword_t unk_mask) {
   // Mask values vary. Probably matching user types? Local/remote?
-  // Games seem to sit and loop until we trigger this notification.
+
+  // To fix game modes that display a 4 profile signin UI (even if playing alone):
+  // XN_SYS_SIGNINCHANGED
```
Triang3l, line 551: Ideally these (for now, only these two, with more added if/when found in an acceptable source) should be in xbox.h, there's already `typedef uint32_t XNotificationID;` in it. We still need to discuss the exact naming convention though as different apps (such as `XmpApp`) may use their own notification IDs, and probably the app's code files themselves would be the better place for them rather than one huge `enum class`.

Triang3l: Copying what I wrote on Discord in #dev:

There are system notifications as well are per-app notifications, so one question is whether we should be using one global `enum class XNotificationID : uint32_t` with notification IDs for all apps, or keep that `typedef uint32_t XNotificationID` and store app notification IDs in the headers of the respective apps (like we currently do for XmpApp).

- In favor of the former: more type safety, more unified naming and shorter declaration of notification IDs, the set is fixed anyway since we're emulating an existing console, and currently running apps can't have colliding notification IDs.
- In favor of the latter: declaration closer to actual usage, each app's interface is fully declared within its header.

In case of the latter: what prefix should we be using for notification IDs? We're using private (though private doesn't make much sense here IMO since it's for inter-"process" communication) `kMsg` within app classes currently, and no naming at all (hardcoded values everywhere) for system notifications. Though Msg doesn't imply Notification or vice versa, I don't like this name personally.

`kNotificationID…` for app-local, `kSystemNotificationID…` for global? Or `kXNotificationID` / `kSystemXNotificationID`? A global `enum class` would have something more consistent: `XNotificationID::kSystem…`, `XNotificationID::kXMP…`

PR #1822, opened 2021-05-30
[Kernel] Improvements to MmQueryStatistics

Gliniak:
## Summary
- Fixed incorrect calculation of available pages
- Changed amount of total virtual bytes
- Added real amount of reserved virtual bytes

## Detailed
- ``total_virtual_memory_bytes`` was taken by some shenanigans directly from console (I taken it from one title, but it seems to be const within titles 🤔 )
- ``title.available_pages`` was probably calculated incorrectly. We forgot to remove pages used by kernel (0x300). On console they're used meanwhile in Xenia they're don't, but still we should be aware of them. (This fixes ``Prototype 2``)

## Doubts:
- I'm not sure if ``GetReservedPageCount`` works as it should


src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -495,33 +495,24 @@ dword_result_t MmQueryStatistics(
...
-#undef GET_USED_PAGE_COUNT
+  uint32_t reserved_pages_bytes = 0;
 
-  assert_true(used_pages < stats_ptr->total_physical_pages);
+  for (uint8_t i = 0; i < std::size(heap_page_sizes); i++) {
+    xe::BaseHeap* heap =
+        kernel_memory()->LookupHeapByType(true, heap_page_sizes[i]);
+    assert_not_null(heap);
+
+    used_pages += (heap->GetUsedPagesCount() * heap->page_size()) / 4096;
```
Triang3l, line 507: Here the global critical region lock is taken twice, not once, so there may be an allocation between `GetUsedPagesCount()` and `GetReservedPageCount()`, and in this case they may be inconsistent. But overall, you don't even need separate GetReserved and GetUnreserved, as they are (total minus each other).

src/xenia/memory.cc
```diff
@@ -716,6 +716,10 @@ void BaseHeap::DumpMap() {
   }
 }
 
+uint32_t BaseHeap::GetUsedPagesCount() {
```
Triang3l, line 719: Inconsistent usage of singular/plural words. We use "PageCount", not "PagesCount", in the code base.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -495,33 +495,24 @@ dword_result_t MmQueryStatistics(
...
   uint32_t used_pages = 0;
-  used_pages += GET_USED_PAGE_SIZE(heap_a);
-  used_pages += GET_USED_PAGE_SIZE(heap_c);
-  used_pages += GET_USED_PAGE_SIZE(heap_e);
-#undef GET_USED_PAGE_SIZE
-#undef GET_USED_PAGE_COUNT
+  uint32_t reserved_pages_bytes = 0;
 
-  assert_true(used_pages < stats_ptr->total_physical_pages);
+  for (uint8_t i = 0; i < std::size(heap_page_sizes); i++) {
```
gibbed, line 502: `std::size` returns a `size_t`, why `uint8_t` for the iterator?

gibbed (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
Gliniak: Thanks!
Last thing I did before pushing was change from 3 to ``std::size(heap_page_sizes)`` and I forget about switching from uint8_t.

src/xenia/memory.cc
Gliniak: Good catch 🐱 Thanks

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -495,33 +495,24 @@ dword_result_t MmQueryStatistics(
...
+  }
 
+  assert_true(used_pages < stats_ptr->total_physical_pages);
   stats_ptr->title.available_pages =
-      stats_ptr->total_physical_pages - used_pages;
-  stats_ptr->title.total_virtual_memory_bytes =
-      0x2FFF0000;  // TODO(gibbed): FIXME
-  stats_ptr->title.reserved_virtual_memory_bytes =
-      0x00160000;                                // TODO(gibbed): FIXME
+      stats_ptr->total_physical_pages - stats_ptr->kernel_pages - used_pages;
```
Triang3l, line 513: Nice catch on Discord — this should be accumulated in the global critical region so it can't be, for instance, above 512 MB if freeing in one heap and allocation in another heap happens between iterations.

Triang3l (changes requested):

src/xenia/memory.cc
```diff
@@ -353,6 +353,25 @@ BaseHeap* Memory::LookupHeapByType(bool physical, uint32_t page_size) {
 
 VirtualHeap* Memory::GetPhysicalHeap() { return &heaps_.physical; }
 
+void Memory::GetHeapsPageStatsSummary(std::vector<BaseHeap*> provided_heaps,
```
Triang3l, line 356: The vector should not be passed as value as it causes unneeded allocations, it should be `const std::vector&` if it's read-only or `std::vector&` if writable. For read-only, what's even more preferred is a raw pointer to the first element + count. for greater flexibility

Triang3l (changes requested):

src/xenia/memory.cc
```diff
@@ -353,6 +353,25 @@ BaseHeap* Memory::LookupHeapByType(bool physical, uint32_t page_size) {
...
+  auto lock = global_critical_region_.Acquire();
+  for (const BaseHeap* heap : provided_heaps) {
+    uint32_t heap_unreserved_pages = heap->ureserved_page_count();
+    uint32_t heap_reserved_pages = heap->reserved_page_count();
+
+    unreserved_pages += heap_unreserved_pages;
+    reserved_pages += heap_reserved_pages;
+    used_pages += ((heap->total_page_count() - heap_unreserved_pages) *
+                   heap->page_size()) /
+                  4096;
```
Triang3l, line 370: Unless there's reverse engineering data proving otherwise for MmQueryStatistics (that it needs to report the physical memory 4 KB portion count — but then I wouldn't call them "pages"), the raw physical memory doesn't have a concept of paging at all, while virtual memory (0xA… are virtual too, but mapped directly 1:1 to physical) has pages with different sizes (4 KB, 64 KB, 16 MB), therefore you can't call an arbitrary 4 KB region a "page".

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -495,33 +495,26 @@ dword_result_t MmQueryStatistics(
...
+  uint32_t reserved_pages = 0;
+  uint32_t unreserved_pages = 0;
   uint32_t used_pages = 0;
-  used_pages += GET_USED_PAGE_SIZE(heap_a);
-  used_pages += GET_USED_PAGE_SIZE(heap_c);
-  used_pages += GET_USED_PAGE_SIZE(heap_e);
-#undef GET_USED_PAGE_SIZE
-#undef GET_USED_PAGE_COUNT
+  uint32_t reserved_pages_bytes = 0;
+  std::vector<BaseHeap*> virtual_heaps;
```
Triang3l, line 503: Allocating a vector on every call when you need only [3] heaps is overkill.

Triang3l (changes requested):

src/xenia/memory.h
```diff
@@ -113,6 +113,17 @@ class BaseHeap {
   // Size of each page within the heap range in bytes.
   uint32_t page_size() const { return page_size_; }
 
+  // Amount of pages assigned to heap
+  uint32_t total_page_count() const { return uint32_t(page_table_.size()); }
+
+  // Sum of ureserved pages in heap
+  uint32_t ureserved_page_count() const { return unreserved_page_count_; }
```
Triang3l, line 120: "ureserved" instead of "unreserved" (twice, in the comment and in the name).

Triang3l (changes requested):

src/xenia/memory.cc
```diff
@@ -909,6 +900,7 @@ bool BaseHeap::AllocFixed(uint32_t base_address, uint32_t size,
     page_entry.allocation_protect = protect;
     page_entry.current_protect = protect;
     page_entry.state = kMemoryAllocationReserve | allocation_type;
+    unreserved_page_count_--;
```
Triang3l, line 903: This shouldn't be done if the page is already reserved (and the game is just committing it, as a common usage case).

Triang3l (changes requested):

src/xenia/memory.cc
```diff
@@ -353,6 +353,27 @@ BaseHeap* Memory::LookupHeapByType(bool physical, uint32_t page_size) {
 
 VirtualHeap* Memory::GetPhysicalHeap() { return &heaps_.physical; }
 
+void Memory::GetHeapsPageStatsSummary(const BaseHeap* provided_heaps[],
```
Triang3l, line 356: `const BaseHeap* const* provided_heaps`, as both the heaps themselves and the array of the heaps are not modified by this function.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -495,33 +495,24 @@ dword_result_t MmQueryStatistics(
...
+  uint32_t reserved_pages = 0;
+  uint32_t unreserved_pages = 0;
   uint32_t used_pages = 0;
-  used_pages += GET_USED_PAGE_SIZE(heap_a);
-  used_pages += GET_USED_PAGE_SIZE(heap_c);
-  used_pages += GET_USED_PAGE_SIZE(heap_e);
-#undef GET_USED_PAGE_SIZE
-#undef GET_USED_PAGE_COUNT
+  uint32_t reserved_pages_bytes = 0;
+  static const BaseHeap* physical_heaps[3] = {
```
Triang3l, line 502: This shouldn't be `static`, otherwise this will be initialized only once (will be broken if heaps — or the Memory object — are recreated, will still keep the values from the previous call), and only from one thread.

Triang3l (changes requested):

Gliniak:
For unknown (yet) reason. This change causes our favourite wild west/horse simulator and probably some other games to flicker in certain spots, but only with resolution scaling enabled.

PR #1824, opened 2021-05-31
[Base] Fixed edgecase in fnfp related to single char dirs

Gliniak:
Im not sure if test case: `TEST_PATH(utf8::find_name_from_path, "/C/", "/C");` is even valid, but I can remove it at anytime.


gibbed:
The bugfix was superseded by dcffc1a4d5207bba49fba68b60c6218efc23a600 and removed from this PR, but we can keep the new tests.

Issue #1579, opened 2020-03-28
Add a disclaimer about Vulkan being buggy


Margen67:
People seem to use Vulkan despite the d3d12 backend being superior, so add a message so people stop asking which backend is better.

gandalfthewhite19890404:
Here is similar issue - https://github.com/xenia-project/xenia/issues/1484

LittleStinky123:
Would be really nice if they added it, my PC crashed with a black screen when using vulkan.

PR #1833, opened 2021-06-10
Added a warning in the config file about Vulkan being obsolete and buggy.

AnnoyingRain5:
This should resolve https://github.com/xenia-project/xenia/issues/1579


Triang3l:
There are some typos, as well as line break formatting different than what is required by our clang-format (run `xb format` after `git add` if you have outstanding changes or checked out a previous commit if you want to replace via force-pushing, or `clang-format -i src/xenia/….cc` directly if you want to format non-staged files). Still, this is a very temporary change, I don't think having such a message is important, considering it's already written on our wiki.

src/xenia/app/xenia_main.cc
```diff
@@ -47,7 +47,10 @@
 #include "third_party/xbyak/xbyak/xbyak_util.h"
 
 DEFINE_string(apu, "any", "Audio system. Use: [any, nop, sdl, xaudio2]", "APU");
-DEFINE_string(gpu, "any", "Graphics system. Use: [any, d3d12, vulkan, null]",
+DEFINE_string(gpu, "any",
+              "Graphics system. Use: [any, d3d12, vulkan, null]\nWARNING: "
+              "Vulkan support is obsolete known to be buggy.\nIt is highly "
```
Triang3l, line 52: "obsolete known", "recommen**e**ded" are not correct in English.

src/xenia/app/xenia_main.cc
```diff
@@ -47,7 +47,10 @@
 #include "third_party/xbyak/xbyak/xbyak_util.h"
 
 DEFINE_string(apu, "any", "Audio system. Use: [any, nop, sdl, xaudio2]", "APU");
-DEFINE_string(gpu, "any", "Graphics system. Use: [any, d3d12, vulkan, null]",
+DEFINE_string(gpu, "any",
```
Triang3l, line 50:
```
DEFINE_string(
  gpu, "any",
  "long multiline string"
```

Triang3l (changes requested):

src/xenia/app/xenia_main.cc
AnnoyingRain5: I have no idea how I missed that, I have fixed both of the typos in my latest commit.

AnnoyingRain5:
> I don't think having such a message is important, considering it's already written on our wiki.

[This issue](https://github.com/xenia-project/xenia/issues/1579) has been left open and according to Etokapa in discord, 

> There's some text in the titlebar, but many people don't read it and then wonder why Vulkan doesn't work.

So I still believe that this PR should be merged.

Triang3l:
Since we are moving quickly to a new Vulkan-based GPU emulation implementation (#2028), and it's developed on the main branch, this is not actual anymore.

PR #1836, opened 2021-06-16
[Kernel] Improved cross-title save support via XamContentAggregate*

emoose:
This improves support for cross-title saves, for titles that use XamContentAggregate* to access them at least.

AFAIK that seems to be the most common way games access saves cross-title (though I know Mass Effect 2 at least [uses a different way](https://github.com/xenia-project/game-compatibility/issues/1769), involving XContentType::Publisher & I guess the publisher portion of title-id, which this PR doesn't support atm)

Main changes:
- Merge `ContentData` into XCONTENT_DATA & `ContentAggregateData` into XCONTENT_AGGREGATE_DATA (we can handle the purpose of `ContentData` with getters/setters instead)

- Update ContentManager functions, so they use `XCONTENT_AGGREGATE_DATA` instead of `ContentData` (as aggregate contains extra info about title_id, and can be easily converted from/to XCONTENT_DATA), and update ResolvePackagePath to make use of XCONTENT_AGGREGATE_DATA::title_id.

(this also adds a `kCurrentlyRunningTitleId` constant of -1/FFFFFFFF to signal using kernel_state->title_id(), I figured 0 would probably be used by some homebrew, but hopefully nothing uses FFFFFFFF)

- Add support for loading alternate-title-IDs header into XexModule (reads unsafe header into a safer std::vector<uint32_t>), that header seems to tell what titles can be accessed for cross-title support

- Update XamContentAggregateCreateEnumerator to use the alternate-title-ids header when searching for saves

- Rename XamContentCreateEx -> xeXamContentCreate, so a `content_data_size` param can be added to tell what kind of content_data is used (needed for XamContentCreateInternal to pass along a XCONTENT_AGGREGATE_DATA & let it be handled properly)

- Simplify XamAppEnumerateContentAggregate code - previously it had a useless XSUCCEEDED check (XSUCCEEDED is for NTSTATUS but it was checking a win32 result, so it couldn't detect any errors), & would also manually copy in title_id with pointer offsets/reinterpret-cast, just moved the title_id field inside XCONTENT_AGGREGATE_DATA instead so it's at the offset the pointer stuff was writing to. (I'm curious if there's a reason AGGREGATE_DATA wasn't already changed like this though)

### Titles affected

Tested with the following:
- BlazBlue: Continuum Shift Extend allows loading saves from the previous BlazBlue: Continuum Shift game, when starting the game normally the gallery shows 0% unlocked, but using the load-previous-save option now works without errors, and with a completed (and extracted) save from previous game the gallery now shows 16% unlocked.

- Borderlands 2 displays a message and unlocks some in-game items if a BL1 save is detected
16_06_21_02_34_36_xenia

Other titles like Lost Planet 2 use XamContentAggregate too and seem to work cross-title on my stfs-writer branch, but haven't tested it with this branch yet.

Normal non-cross-title saves still appear to work fine, but I haven't really tested many games with it.


gibbed:
The reason I distinguished `ContentData` and `X_CONTENT_DATA` is `X_CONTENT_DATA` is the guest form of the structure, and at some point in the future we may want to track additional host-specific data in `ContentData`, and if we're using `X_CONTENT_DATA` host-side then we aren't able to add anything.

Thoughts?

emoose:
Ah right, well with XCONTENT_AGGREGATE_DATA containing the title_id I'm not sure how much else would need to be tracked really, at least on my stfs-writer branch I've been able to add in STFS & TU support without adding anything to those.

Of course I can't tell what might be needed in the future though, but if extra things do end up needing to be tracked host-side it wouldn't be too hard to bring those structs back.

PR #1848, opened 2021-06-18
[Base] Fix Clock platform functions on POSIX.

JoelLinn:
It's best to have this when working with #1431 because broken clock freaks ImGUI out and freezes hid demo
@bwrsandman put another version of this in #1433 originally but It's being held up by the other changes

### DIFF
--- a/src/xenia/base/clock_posix.cc
+++ b/src/xenia/base/clock_posix.cc
@@ -2,41 +2,51 @@
  ******************************************************************************
  * Xenia : Xbox 360 Emulator Research Project                                 *
  ******************************************************************************
- * Copyright 2019 Ben Vanik. All rights reserved.                             *
+ * Copyright 2021 Ben Vanik. All rights reserved.                             *
  * Released under the BSD license - see LICENSE in the root for more details. *
  ******************************************************************************
  */
 
-#include "xenia/base/clock.h"
-
 #include <sys/time.h>
-#include <time.h>
+
+#include "xenia/base/assert.h"
+#include "xenia/base/clock.h"
 
 namespace xe {
 
 uint64_t Clock::host_tick_frequency_platform() {
   timespec res;
-  clock_getres(CLOCK_MONOTONIC_RAW, &res);
+  int error = clock_getres(CLOCK_MONOTONIC_RAW, &res);
+  assert_zero(error);
+  assert_zero(res.tv_sec);  // Sub second resolution is required.
 
-  return uint64_t(res.tv_sec) + uint64_t(res.tv_nsec) * 1000000000ull;
+  // Convert nano seconds to hertz. Resolution is 1ns on most systems.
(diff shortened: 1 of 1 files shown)


src/xenia/base/clock_posix.cc
```diff
@@ -7,40 +7,48 @@
...
 uint64_t Clock::QueryHostSystemTime() {
-  struct timeval tv;
-  gettimeofday(&tv, NULL);
-
-  uint64_t ret = tv.tv_usec;
-  ret /= 1000;  // usec -> msec
-
-  ret += (tv.tv_sec * 1000);  // sec -> msec
-  return ret;
+  static_assert(sizeof(timeval::tv_sec) == 8);
```
Triang3l, line 36: An explicit cast to uint64_t is preferred.

Triang3l (changes requested):

src/xenia/base/clock_posix.cc
```diff
@@ -7,40 +7,48 @@
...
+  constexpr uint64_t seconds_per_day = 3600 * 24;
+  // Don't forget the 89 leap days.
+  constexpr uint64_t seconds_1601_to_1970 =
```
Triang3l, line 39: Preferably add a link to https://docs.microsoft.com/en-us/windows/win32/sysinfo/converting-a-time-t-value-to-a-file-time please as a reference.

Triang3l (changes requested):

src/xenia/base/clock_posix.cc
```diff
@@ -7,40 +7,51 @@
...
+
+  timeval now;
+  auto error = gettimeofday(&now, nullptr);
+  assert_true(!error);
+
+  // NT systems use 100ns intervals.
+  return static_cast<uint64_t>(
+             (static_cast<int64_t>(now.tv_sec) + seconds_1601_to_1970)) *
+             10000000ull +
+         now.tv_usec * 10;
```
Triang3l, line 50: tv_usec is also signed, the final result needs to be converted.

src/xenia/base/clock_posix.cc
```diff
@@ -7,40 +7,51 @@
...
+#include "xenia/base/assert.h"
+#include "xenia/base/clock.h"
 
 namespace xe {
 
 uint64_t Clock::host_tick_frequency_platform() {
   timespec res;
-  clock_getres(CLOCK_MONOTONIC_RAW, &res);
+  auto error = clock_getres(CLOCK_MONOTONIC_RAW, &res);
+  assert_true(!error);
```
Triang3l, line 20: There are assert_false and assert_zero, but no difference really.

Triang3l (changes requested):

src/xenia/base/clock_posix.cc
JoelLinn: wtf yes

PR #1860, opened 2021-06-29
[Base] Implement message boxes on Linux

JoelLinn:
Use SDL for message boxes.
Loads SDL dynamically so we don't depend on it in `xenia-base`

### DIFF
--- a/src/xenia/base/system_linux.cc
+++ b/src/xenia/base/system_linux.cc
@@ -7,14 +7,19 @@
  ******************************************************************************
  */
 
+#include <alloca.h>
+#include <dlfcn.h>
 #include <stdlib.h>
 
-#include <string>
+#include <cstring>
 
 #include "xenia/base/assert.h"
-#include "xenia/base/platform_linux.h"
+#include "xenia/base/logging.h"
 #include "xenia/base/string.h"
 #include "xenia/base/system.h"
+
+// Use headers in third party to not depend on system sdl headers for building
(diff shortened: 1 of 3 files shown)


src/xenia/base/system_linux.cc
```diff
@@ -27,7 +32,39 @@ void LaunchWebBrowser(const std::string_view url) {
 void LaunchFileExplorer(const std::filesystem::path& path) { assert_always(); }
 
 void ShowSimpleMessageBox(SimpleMessageBoxType type, std::string_view message) {
-  assert_always();
+  void* libsdl2 = dlopen("libSDL2.so", RTLD_LAZY | RTLD_LOCAL);
+  assert_not_null(libsdl2);
```
Triang3l, line 36: The situation in which the user doesn't have libSDL2 is normal, at least in the `base` part (as opposed to `apu` and `hid`), so this is not really a static-ish assertion. I think a simple console and/or log output fallback should be provided (in case of a crash, the log is likely the first place to look for a message in) when SDL is not available.

Triang3l (changes requested):

src/xenia/base/system_linux.cc
JoelLinn: My thinking was that assertions are only in debug code and developers  can handle it.
Also the constellation where this code is run but the user doesn't have SDL is pretty exotic.
Also log won't work in most cases this is used _currently_ - on startup when the options are checked the logger isn't up yet

PR #1861, opened 2021-06-30
[Kernel/XAM] More time/date/locale.

gibbed:
RFC.

There's still stuff to be implemented.

Need some sort of way to auto-detect user's locale, country, language, timezone, all that nonsense to reasonable defaults. Fall back to en_US + maybe EST? otherwise.


JoelLinn:
If we were using C++20 I'd say use things like `std::chrono::current_zone`....

I think a reasonable way could be to dynamically load the ICU C library, which is included as standard on Windows 10 and Desktop Linux distributions. We only need the most basic apis from it. Get locale, extract country and language; Get timezone and extract offset from it. Then try to find matching values for 360

If loading icuc fails or no matching locale/time zone is found, fall back to a user configurable value (default english, but not sure about time zone - UTC0 seems more reasonable)

gibbed:
There's a timezone header in the date library we're already using, so would probably leverage that:

https://howardhinnant.github.io/date/tz.html

The biggest issue is doing the mapping to the stuff the 360 has.

JoelLinn:
We have the ISO codes in the include files, do we? What about looping through them and finding a match?
Trouble may be that we might be setting combinations of locales languages that are unsupported on 360 and trigger weird bugs in guests. I for example use `en_DE` on my system and there is a number of apps which simply crash due to it.

gibbed:
The tables I have for locale shows exactly what combinations are possible. It's the timezone stuff I haven't done yet.

src/xenia/kernel/xboxkrnl/xboxkrnl_xconfig.cc
```diff
@@ -66,6 +70,68 @@ X_STATUS xeExGetXConfigSetting(uint16_t category, uint16_t setting,
     case 0x0003:
       // XCONFIG_USER_CATEGORY
       switch (setting) {
+#if XE_PLATFORM_WIN32
+          // TODO(gibbed): *REQUIRED BEFORE MERGING* all of the time zone stuff
+          // needs to get appropriate value from user settings / system.
```
halotroop2288, line 75: You could always just set the time zone to UTC by default. I've checked the other semi-modern emulators, and they barely even bother with time zone. RPCS3 defaults to UTC as well.

halotroop2288 (approved):

PR #1867, opened 2021-07-15
[VFS/Kernel] Improved cache partition support via XMountUtilityDrive workarounds

emoose:
(This is mostly the same as #1740, cleaned up a little & without the CriticalRegion stuff that was merged in with #1849, made this separate since that PR has some info I didn't want to delete)

With this PR the **XMountUtilityDrive** function statically-linked into games will hopefully always return success, which should help with games that use that function to test HDD availability (eg. all the Halo games)

This is mostly done with workarounds though, as the linked-in code seems to include nearly a whole filesystem implementation inside it AFAIK, using things like the Stfs* kernel functions that we don't currently handle.

Fortunately we can just return some status codes that the utility drive code expects instead, and let it think that the cache0:/cache1:/cache: symlinks were created by it (while those were actually already created by us earlier on, pointing to HostPathDevices), and with that hardly any of the cache-FS code inside the game actually needs to be ran.

(maybe in future it could be worth investigating how to make that FS stuff work properly, but with the end-result of that being a 500MB+ binary blob that no tools have any support for I didn't see much point in trying it myself... especially since all the cache partition accesses map over to Win32 folders fine, and there's not any STFS metadata or console-compatibility to worry about here)

---

A new **VFS::NullDevice** class is used to handle the `\Device\Harddisk0\Partition0`, `Cache0` & `Cache1` devices, as the XMountUtilityDrive code requires reads/writes to those paths to succeed. Right now a single NullDevice gets created for `\Device\Harddisk0` which then handles the Partition0/Cache0/Cache1 paths.

(maybe it'd be better if those paths were all separate NullDevice instances instead, but IIRC there's some limit with VFS that doesn't let device roots have handles opened for them, so this pretty much just treats Partition0/Cache0/Cache1 as files, which works fine for how the game accesses those devices)

**XamTaskSchedule** has been updated to spawn a new thread for each task - this is needed because XMountUtilityDrive seems to communicate some stuff with the task that it scheduled, updating some addrs & waiting for events to be signalled in response etc.

Haven't really looked into how Xam handles those tasks yet, mostly just got lucky that letting it spawn a new thread worked fine. (I'd guess it's probably not a single-thread task-queue at least since the code for these tasks can include busy/wait-loops, which would block any other tasks afaik)

There's also some functions for checking status of a task/forcing task to quit/etc, but haven't handled those yet, just stubbed ones that the XMountUtilityDrive/task code needed.

(If anyone knows what game the original XamTaskSchedule code was created for it'd be great if you can try it with this new threaded code!)

---

Some things I haven't handled here but might be nice to add in future:

- Cache currently persists between titles, as explained [in comments of #1664](https://github.com/xenia-project/xenia/pull/1664/files#r550715241) some games can have issues with cache from other titles though - X360 itself seems to clear the cache _partition_ between different titleids.

We could maybe just clear the cache folders whenever a game is launched, but this would destroy any data that games can use between runs (eg. Halo auto-recorded demos are stored in cache until user decides to save them properly IIRC)

(since `mount_cache` has to be opted-in atm this shouldn't really be much of a problem right now, if they're enabling mount_cache then hopefully they'd know to try clearing cache if they have issues)

- Could probably detect when game requests cache: access (inside NullDevice maybe, or when game tries running ObCreateSymbolicLink), and use that as the trigger for creating the cache folders & mounting cache HostPathDevices, instead of creating/mounting on Xenia launch & needing `mount_cache` to be set (could still include mount_cache as a way to disable cache stuff though)

---
### Tested titles

Tested with `mount_cache = true`
- Halo 3 MP Beta: previously would dirty-disc error if XMountUtilityDrive wasn't successful, now seems to boot up fine & writes some data into cache folder.

- Halo 3: now writes 300MB+ cacheXXX.map files into cache0/cache1, and allows access to Theater mode without any "No HDD connected" errors.
Theater demos from actual X360 can also be played here too (but quickly exits out with a desync error, could be worth looking at this some more...)

- Halo Reach: writes cacheXXX.map files into cache0/1 folders (game doesn't seem to render menus with latest xenia-master though so not sure about theater mode, maybe need to redump it...)

- Forza Motorsport 2: now writes some files into cache folder

- Sega Rally Revo: previous XMountUtilityDrive attempts (eg forcing TitleInsecureUtilityDrive privilege) would cause game to hang with a black-screen on boot, now continues onto loading screens

The XMountUtilityDrive code did change a lot over time (going from cache: to cache0: & cache1:, XEX privilege checks being added/removed...), so there's a chance some title might not work well with these workarounds, fortunately haven't ran into any yet though!

### DIFF
--- a/src/xenia/app/xenia_main.cc
+++ b/src/xenia/app/xenia_main.cc
@@ -329,6 +329,22 @@ int xenia_main(const std::vector<std::string>& args) {
         emulator->file_system()->RegisterSymbolicLink("cache1:", "\\CACHE1");
       }
     }
+
+    // Some (older?) games try accessing cache:\ too
+    // NOTE: this must be registered _after_ the cache0/cache1 devices, due to
+    // substring/start_with logic inside VirtualFileSystem::ResolvePath, else
+    // accesses to those devices will go here instead
+    auto cache_device =
+        std::make_unique<xe::vfs::HostPathDevice>("\\CACHE", "cache", false);
+    if (!cache_device->Initialize()) {
+      XELOGE("Unable to scan cache path");
+    } else {
+      if (!emulator->file_system()->RegisterDevice(std::move(cache_device))) {
+        XELOGE("Unable to register cache path");
+      } else {
+        emulator->file_system()->RegisterSymbolicLink("cache:", "\\CACHE");
+      }
+    }
   }
 
   // Set a debug handler.
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -42,6 +42,7 @@
 #include "xenia/ui/imgui_dialog.h"
 #include "xenia/vfs/devices/disc_image_device.h"
 #include "xenia/vfs/devices/host_path_device.h"
+#include "xenia/vfs/devices/null_device.h"
 #include "xenia/vfs/devices/stfs_container_device.h"
 #include "xenia/vfs/virtual_file_system.h"
 
@@ -279,7 +280,7 @@ X_STATUS Emulator::LaunchXexFile(const std::filesystem::path& path) {
   // and then get that symlinked to game:\, so
   // -> game:\foo.xex
 
-  auto mount_path = "\\Device\\Harddisk0\\Partition0";
+  auto mount_path = "\\Device\\Harddisk0\\Partition1";
 
   // Register the local directory in the virtual filesystem.
   auto parent_path = path.parent_path();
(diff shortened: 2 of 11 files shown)


Etokapa:
This PR allows for Madden NFL 08, 09, 10, 11, 12, & 13 to boot now.

0wenModding:
PR fixes Minecraft's graphical corruptions but introduces no audio, frequent crashing on boot / soft locks, and crashes upon trying to generate new worlds. This was tested on the latest update tu75

emoose:
@gibbed many thanks for the merge!
BTW do you recall what XamTaskSchedule was originally added for? I'm not sure if the threading changes here might affect that at all.

One thing I should have mentioned here, seems games can provide a custom partition size to XMountUtilityDrive, which I guess should probably affect [what NtDeviceIoControlFile returns](https://github.com/xenia-project/xenia/blob/master/src/xenia/kernel/xboxkrnl/xboxkrnl_io.cc#L685).

IIRC I tried finding some way to capture the param given to XMountUtilityDrive but didn't have any luck, I think it gets used in the code that handles formatting/file-system headers/etc, maybe can be recovered by checking headers written to Partition0/Cache0/Cache1 or something similar.

Games I tested with all seem to use 0xFF000 but there's most likely some using a different value, not sure if NtDeviceIoControlFile returning incorrect values will prevent XMountUtilityDrive from succeeding or not, if any games are still having cache issues that's the likely culprit though.

gibbed:
I don't recall offhand what I implemented `XamTaskSchedule` for, I'd say it's likely it was for dash maybe?

Gliniak:
``XamTaskSchedule`` was implemented for aggregated content function. IIRC every title that uses ``XamContentAggregateCreateEnumerator`` executes ``XamTaskSchedule`` right after it

PR #1877, opened 2021-08-10
[SDL2] Update to version 2.0.16

jgoyvaerts:
As requested by holiday boy @JoelLinn


third_party/SDL2-static.lua
```diff
@@ -240,6 +240,7 @@ project("SDL2")
     "SDL2/src/joystick/dummy/SDL_sysjoystick.c",
     "SDL2/src/joystick/hidapi/SDL_hidapijoystick.c",
     "SDL2/src/joystick/hidapi/SDL_hidapi_gamecube.c",
+    "SDL2/src/joystick/hidapi/SDL_hidapi_luna.c",
```
Triang3l, line 243: https://github.com/libsdl-org/SDL/commit/6c92bf540cce9dab25b63a76dc3dddabe0c817b5 seems to be missing from this update, can you run some find & replace to generate the new list from SDL.vcxproj?

Triang3l (changes requested):

third_party/SDL2-static.lua
jgoyvaerts: Done

PR #1881, opened 2021-08-22
[XAM] Implemented XamCreateEnumeratorHandle

Gliniak:
- Thanks Rick for providing names for parameters

After spending a some time I figured out that extra_size is just a size of single item, so I changed its name 🐑

This resolves guest crashes in few games that lately started to crash in main menu


src/xenia/kernel/xam/xam_enum.cc
```diff
@@ -82,13 +82,24 @@ dword_result_t XamEnumerate(dword_t handle, dword_t flags, lpvoid_t buffer,
...
 
-dword_result_t XamCreateEnumeratorHandle(unknown_t unk1, unknown_t unk2,
-                                         unknown_t unk3, unknown_t unk4,
-                                         unknown_t unk5, unknown_t unk6,
-                                         unknown_t unk7, unknown_t unk8) {
-  return X_ERROR_INVALID_PARAMETER;
+dword_result_t XamCreateEnumeratorHandle(dword_t user_index, dword_t app_id,
+                                         dword_t open_message,
+                                         dword_t close_message,
+                                         dword_t item_size, dword_t item_count,
```
gibbed, line 88: It's `extra_size` because there's a base size it's added to. Please don't rename the argument.

https://github.com/xenia-project/xenia/blob/f6f524b8148145c5b5b5ac0a79809df979069bc6/src/xenia/kernel/xenumerator.cc#L23-L27

gibbed (changes requested):

src/xenia/kernel/xam/xam_enum.cc
Gliniak: I renamed it after a few calculation based on PGR4 disassembly, but sure I can revert that change

gibbed:
The main problem of this implementation is that it will be empty regardless. We need to match this against a list of `app_id`, `open_message`, `close_message`, and `item_size` to create the appropriate enumerator. Defaulting to an empty static untyped enumerator is probably fine (with warnings).

src/xenia/kernel/xam/xam_enum.cc
gibbed: It's not the size of an item, but the extra size that gets added to the native enumerator object for that type enumerator specific data.

Gliniak:
I agree, however I'm not sure if item_size is important there 🤔 
In my opinion: ``app_id``, ``open_message`` and ``close_message`` should be enough to determine exactly what kind of enumerator we need.

gibbed:
> I agree, however I'm not sure if item_size is important there 🤔
> In my opinion: `app_id`, `open_message` and `close_message` should be enough to determine exactly what kind of enumerator we need.

`extra_size` is important if we're create our own enumerator implementations to pass back instead of the untyped enumerator that has nothing in it. If we get a different size, then that suggests it's not the same enumerator.

gibbed:
Just to clarify here, `extra_size` does not map to `item_size` argument of the constructor. This is a value passed to `Initialize(...)`.

`item_size` is implicit with the type of enumerator being created, and is not passed in as an argument. Which is why we need to match against those arguments I mentioned to figure out what enumerator is actually being created.

Gliniak:
Ok, good to know. I was worried that this value might not be always correct and can be manipulated freely. That is why I insisted to provide at least one correct value for each app_id, open & close message and check it with what we received.

halotroop2288 (approved):

halotroop2288 (approved):

PR #1885, opened 2021-08-30
Added OS details to the log #1884

neha200:
Added a function to retrieve OS in  XELOGI in emulator.cc using QString

### DIFF
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -10,6 +10,7 @@
 #include "xenia/emulator.h"
 
 #include <cinttypes>
+#include <QOperatingSystemVersion>
 
 #include "config.h"
 #include "third_party/fmt/include/fmt/format.h"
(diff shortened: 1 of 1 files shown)


src/xenia/emulator.cc
```diff
@@ -704,6 +705,7 @@ X_STATUS Emulator::CompleteLaunch(const std::filesystem::path& path,
   auto xam = kernel_state()->GetKernelModule<kernel::xam::XamModule>("xam.xex");
 
   XELOGI("Launching module {}", module_path);
+  XELOGI("[OS] : ", QOperatingSystemVersion::current().name());
```
Triang3l, line 708: We don't have Qt globally currently, and even if we did (though if added, its usage will still likely be limited to the UI), it would have to be registered in `links` of Premake projects of every module referencing the Emulator class; it also needs to be made sure that Qt is present in the `third_party` directory.

Also, we don't use excessive punctuation such as brackets, or spaces before colons — log messages are written in plain English where possible.

This (after launching a module) also seems to be a very random place for emitting this kind of info — it would be more appropriate to place it in `xe::InitializeLogging`, which is called from the platform-specific `main` function (not only in the emulator itself, but also in all the tools), where platform-dependent code for getting the version can be localized, without using Qt or exposing a function from `xenia-base` that will be used only in one place — in this case, it would be an argument of `xe::InitializeLogging`.

Triang3l (changes requested):

PR #1891, opened 2021-09-20
Rebindable keyboard controls

### LINKED ISSUE #1333: Custom Controls Support
I don't think anyone has suggested this yet (at least I couldn't find anything asking for it), but a custom controls dialog or .conf file would be very much appreciated. The keyboard controls are a little awkward as they stand now.

Also some way to use generic gamepads would be appreciated to. I tried x360ce but I couldn't get it to work at all.

epozzobon:
This pull request makes it possible to rebind the keyboard controls in the xenia config file. This fixes #1333.
It supports switching layers using caps lock (as the current default key bindings) and multiple keys for the same gamepad input.

I acknowledge that the syntax is a bit awkward, and I would be happy to get some feedback on how I can improve it while maintaining compatibility with xenia's current key bindings.

As it stands, the syntax works like this:
`keybind_a = "0xBA"`      # binds key code 0xBA to the A button
`keybind_b = "'X'" `     # binds the X key to the B button. Chars are indicated by `'` as per C convention.
`keybind_a = "0xBA"`      # binds key code 0xBA to the A button
`keybind_right_trigger = "'E' 'O'"`      # binds both keys E and O to the right trigger. Space is used as a separator.

Bind the S key to dpad down when caps lock or shift are enabled, and to left thumb down otherwise: (notice ^ and _ preceding the char/keycode)
```
keybind_dpad_down = "^'S'" 
keybind_left_thumb_down = "_'S'"
```

### DIFF
--- a/src/xenia/hid/winkey/winkey_input_driver.cc
+++ b/src/xenia/hid/winkey/winkey_input_driver.cc
@@ -2,22 +2,83 @@
  ******************************************************************************
  * Xenia : Xbox 360 Emulator Research Project                                 *
  ******************************************************************************
- * Copyright 2014 Ben Vanik. All rights reserved.                             *
+ * Copyright 2022 Ben Vanik. All rights reserved.                             *
  * Released under the BSD license - see LICENSE in the root for more details. *
  ******************************************************************************
  */
 
 #include "xenia/hid/winkey/winkey_input_driver.h"
 
+#include "xenia/base/logging.h"
 #include "xenia/base/platform_win.h"
 #include "xenia/hid/hid_flags.h"
 #include "xenia/hid/input_system.h"
 #include "xenia/ui/virtual_key.h"
 #include "xenia/ui/window.h"
 
+#define XE_HID_WINKEY_BINDING(button, description, cvar_name, \
+                              cvar_default_value)             \
+  DEFINE_string(cvar_name, cvar_default_value,                \
+                "List of keys to bind to " description        \
+                ", separated by spaces",                      \
+                "HID.WinKey")
+#include "winkey_binding_table.inc"
+#undef XE_HID_WINKEY_BINDING
+
 namespace xe {
 namespace hid {
 namespace winkey {
+
+bool __inline IsKeyToggled(uint8_t key) {
+  return (GetKeyState(key) & 0x1) == 0x1;
+}
+
+bool __inline IsKeyDown(uint8_t key) {
+  return (GetAsyncKeyState(key) & 0x8000) == 0x8000;
+}
+
+bool __inline IsKeyDown(ui::VirtualKey virtual_key) {
+  return IsKeyDown(static_cast<uint8_t>(virtual_key));
+}
+
+void WinKeyInputDriver::ParseKeyBinding(ui::VirtualKey output_key,
+                                        const std::string_view description,
+                                        const std::string_view source_tokens) {
+  for (const std::string_view source_token :
+       utf8::split(source_tokens, " ", true)) {
+    KeyBinding key_binding;
+    key_binding.output_key = output_key;
+
+    std::string_view token = source_token;
+
+    if (utf8::starts_with(token, "_")) {
+      key_binding.lowercase = true;
+      token = token.substr(1);
+    } else if (utf8::starts_with(token, "^")) {
+      key_binding.uppercase = true;
+      token = token.substr(1);
+    }
+
+    if (utf8::starts_with(token, "0x")) {
+      token = token.substr(2);
+      key_binding.input_key = static_cast<ui::VirtualKey>(
+          string_util::from_string<uint16_t>(token, true));
+    } else if (token.size() == 1 && (token[0] >= 'A' && token[0] <= 'Z') ||
+               (token[0] >= '0' && token[0] <= '9')) {
+      key_binding.input_key = static_cast<ui::VirtualKey>(token[0]);
+    }
+
+    if (key_binding.input_key == ui::VirtualKey::kNone) {
+      XELOGW("winkey: failed to parse binding \"{}\" for controller input {}.",
+             source_token, description);
+      continue;
+    }
+
+    key_bindings_.push_back(key_binding);
+    XELOGI("winkey: \"{}\" binds key 0x{:X} to controller input {}.",
+           source_token, key_binding.input_key, description);
+  }
+}
 
 WinKeyInputDriver::WinKeyInputDriver(xe::ui::Window* window)
     : InputDriver(window), packet_number_(1) {
@@ -50,6 +111,13 @@ WinKeyInputDriver::WinKeyInputDriver(xe::ui::Window* window)
     key.repeat_count = evt->repeat_count();
     key_events_.push(key);
   });
+
+#define XE_HID_WINKEY_BINDING(button, description, cvar_name,          \
+                              cvar_default_value)                      \
+  ParseKeyBinding(xe::ui::VirtualKey::kXInputPad##button, description, \
+                  cvars::cvar_name);
+#include "winkey_binding_table.inc"
+#undef XE_HID_WINKEY_BINDING
 }
 
 WinKeyInputDriver::~WinKeyInputDriver() = default;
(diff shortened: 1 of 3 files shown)


src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -10,17 +10,121 @@
...
+          bind.key = (ui::VirtualKey)(uint8_t) ul;
+        }
+      }
+
+      char buffer[101];
+      buffer[100] = 0;
+      if (bind.key != ui::VirtualKey::kNone) {
+        bind.button = controllerKey;
+        key_bindings_.push_back(bind);
+        snprintf(buffer, 100, "KeyBind: token \"%s\" binds key 0x%x to controller input 0x%x", token, bind.key, controllerKey);
```
gibbed, line 86: Don't use snprintf. `XELOGI` supports formatting. <https://fmt.dev/latest/index.html#format-api>

src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -10,17 +10,121 @@
...
+        XELOGI(buffer);
+      } else {
+        snprintf(buffer, 100, "KeyBind: failed to parse token \"%s\" for controller input 0x%x", token, controllerKey);
```
gibbed, line 89: Don't use snprintf. `XELOGW` supports formatting. <https://fmt.dev/latest/index.html#format-api>

src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -10,17 +10,121 @@
...
+
 namespace xe {
 namespace hid {
 namespace winkey {
 
+void WinKeyInputDriver::ParseKeyBindingCVar(ui::VirtualKey controllerKey, std::string keyBindCVar) {
+  const char *cstr = keyBindCVar.c_str();
+  char token[16];
+  std::size_t off = 0;
+  while (keyBindCVar[off] != 0) {
```
gibbed, line 52: `std::string` is not obligated to be null terminated, this should be a length check.

src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -10,17 +10,121 @@
...
+      KeyBinding bind;
+      if (*s == '_') {
+        bind.lowercase = true;
+        s++;
+      }
+      if (*s == '^') {
+        bind.uppercase = true;
+        s++;
+      }
+      if (s[0] == '\'' && s[1] != '\0') {
```
gibbed, line 71: No verification that there is enough data left.

src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -98,115 +199,86 @@ X_RESULT WinKeyInputDriver::GetState(uint32_t user_index,
...
-      }
-      if (IS_KEY_DOWN('S')) {
-        // S
-        thumb_ly += SHRT_MIN;
-      }
-      if (IS_KEY_DOWN('W')) {
-        // W
-        thumb_ly += SHRT_MAX;
+    bool capital = IS_KEY_TOGGLED(VK_CAPITAL) || IS_KEY_DOWN(VK_SHIFT);
+    for (KeyBinding &b : key_bindings_)
```
gibbed, line 203: `const Keybinding &b`

src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -10,17 +10,121 @@
...
+      }
+      if (*s == '^') {
+        bind.uppercase = true;
+        s++;
+      }
+      if (s[0] == '\'' && s[1] != '\0') {
+        bind.key = (ui::VirtualKey)(uint8_t) s[1];
+      } else {
+        char *endptr;
+        unsigned long ul = strtoul(s, &endptr, 0);
```
gibbed, line 75: Don't use `strtoul`, we have helper functions for this in `base/string_util.h` (see `from_string`).

src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -10,17 +10,121 @@
...
+  const char *cstr = keyBindCVar.c_str();
+  char token[16];
+  std::size_t off = 0;
+  while (keyBindCVar[off] != 0) {
+    std::size_t end = keyBindCVar.find(' ', off);
+    if (end == std::string::npos) {
+      end = keyBindCVar.length();
+    }
+    if (end - off < 15) {
+      memcpy(token, keyBindCVar.c_str() + off, end - off);
```
gibbed, line 58: This seems completely unnecessary to have a raw `char` buffer. A substring of `keyBindCVar` should end up producing a `std::string_view` anyway.

src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -10,17 +10,121 @@
...
+DEFINE_string(keybind_back,              "'Z'",     "XINPUT_GAMEPAD_BACK"             , "KeyBind")
+DEFINE_string(keybind_start,             "'X'",     "XINPUT_GAMEPAD_START"            , "KeyBind")
+DEFINE_string(keybind_left_shoulder,     "'1'",     "XINPUT_GAMEPAD_LEFT_SHOULDER"    , "KeyBind")
+DEFINE_string(keybind_right_shoulder,    "'3'",     "XINPUT_GAMEPAD_RIGHT_SHOULDER"   , "KeyBind")
+
 namespace xe {
 namespace hid {
 namespace winkey {
 
+void WinKeyInputDriver::ParseKeyBindingCVar(ui::VirtualKey controllerKey, std::string keyBindCVar) {
```
gibbed, line 48: `ParseKeyBindingCVar` seems a bit misrepresentative. You take in the value of the cvar here, so it's not really parsing a cvar, it's just parsing a key binding string which came from a cvar.

`ParseKeyBinding` would be a better name.

gibbed (changes requested):
In general, I like what you've done with the token stuff.

I don't really like the single quotes for character key definitions, but I haven't really thought of a better format yet. I'll think on it.

gibbed:
Additionally, I will look at implementing support for TOML arrays in our config reader/writer so we can have that without having a custom thing just for keybinds.

epozzobon:
All the issues you pointed out should now be fixed.
I also thought about a better syntax for the bindings: single quotations could be removed entirely, and instead all bindings are considered ascii by default except when they start with 0x.

I think it would make them look quite natural:
```
keybind_right_trigger = "E O"
keybind_a = "0xBA"
```

epozzobon:
(`xb format` makes a mess out of `DEFINE_string` statements so I'm not able to fix the continuous integration check.)

gibbed:
You can selectively toggle off and on clang-format.

Surround it with:

```// clang-format off```

```// clang-format on```

**Comment by epozzobon** [comment]
I must be doing something wrong, but even with those lines, `xb format` tries to tab the entire namespace

File: src/xenia/hid/winkey/winkey_input_driver.cc [hunk 9]
```diff
@@ -9,16 +9,96 @@
...
 namespace hid {
 namespace winkey {
 
+void WinKeyInputDriver::ParseKeyBinding(ui::VirtualKey controllerKey,
+                                        std::string bindVar) {
+  const char* cstr = bindVar.c_str();
+  std::size_t off = 0;
+  while (off < bindVar.length()) {
+    std::size_t end = bindVar.find(' ', off);
+    end = std::min(end, bindVar.length());
```
gibbed, line 65: I'd prefer end is properly checked if it's `std::string::npos`.

```cpp
if (end != std::string::npos) {
  end = std::min(end, bindVar.length());
} else {
  end = bindVar.length();
}
```

src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -9,16 +9,96 @@
...
+    std::string token = bindVar.substr(off, end - off);
```
gibbed, line 66: `token` can be a `std::string_view`, unless I'm mistaken.

src/xenia/hid/winkey/winkey_input_driver.cc
```diff
@@ -9,16 +9,96 @@
...
+// clang-format on
+
+#define IS_KEY_TOGGLED(key) ((GetKeyState(key) & 0x1) == 0x1)
+#define IS_KEY_DOWN(key) ((GetAsyncKeyState(key) & 0x8000) == 0x8000)
+
 namespace xe {
 namespace hid {
 namespace winkey {
 
+void WinKeyInputDriver::ParseKeyBinding(ui::VirtualKey controllerKey,
```
gibbed, line 59: Please clean up your variable names according to [the style guide](https://google.github.io/styleguide/cppguide.html#Variable_Names).

gibbed (changes requested):

gibbed (approved):

PR #1892, opened 2021-09-22
Fix gcc builds

JoelLinn:
- Replace the GCC aggressive loop optimization bug workaround by a more sane version
- Remove anonymous union and struct in RenderTargetKey (Conflicts with explicit constructor when using GCC). This also adds index sanitizing.

### DIFF
--- a/src/xenia/gpu/render_target_cache.h
+++ b/src/xenia/gpu/render_target_cache.h
@@ -538,13 +538,8 @@ class RenderTargetCache {
     // float32 value to that of an unorm24 with a totally wrong value). If the
     // range hasn't been used yet (render_target.IsEmpty() == true), these are
     // empty too.
-    union {
-      struct {
-        RenderTargetKey host_depth_render_target_unorm24;
-        RenderTargetKey host_depth_render_target_float24;
-      };
-      RenderTargetKey host_depth_render_targets[2];
-    };
+    RenderTargetKey host_depth_render_target_unorm24;
+    RenderTargetKey host_depth_render_target_float24;
     OwnershipRange(uint32_t end_tiles, RenderTargetKey render_target,
                    RenderTargetKey host_depth_render_target_unorm24,
                    RenderTargetKey host_depth_render_target_float24)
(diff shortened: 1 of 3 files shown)


src/xenia/gpu/render_target_cache.h
```diff
@@ -538,20 +538,34 @@ class RenderTargetCache {
...
+    RenderTargetKey host_depth_render_target_float24;
     OwnershipRange(uint32_t end_tiles, RenderTargetKey render_target,
                    RenderTargetKey host_depth_render_target_unorm24,
                    RenderTargetKey host_depth_render_target_float24)
         : end_tiles(end_tiles),
           render_target(render_target),
           host_depth_render_target_unorm24(host_depth_render_target_unorm24),
           host_depth_render_target_float24(host_depth_render_target_float24) {}
+    const RenderTargetKey& GetHostDepthRenderTarget(
+        uint32_t resource_format) const {
```
Triang3l, line 551: A simple `switch (format)` on `xenos::DepthRenderTargetFormat format`, or even better `return format != xenos::DepthRenderTargetFormat::kD24S8 ? float24 : unorm24;` so there is no `default` case, would be enough. Casts need to be done explicitly by the callers or via `GetDepthFormat` if coming from a `RenderTargetKey` in this case.

Triang3l (changes requested):

src/xenia/gpu/render_target_cache.h
JoelLinn: mh, I like switch more becuase it's actually an enum.
default would be a bug, I think I will use assert_always with it

Triang3l: With a : 1 bitfield source, assuming inlining, what is the compiler output for both the if and the switch? Ideally it should be the index in an effective address operand, not a jump table. We return a reference though — there's no third option in release builds also without the non-returning assertion, and changing the return type to a pointer allowing nullptr would overcomplicate all callers for no reason.

Triang3l: This is a widespread practice in the Xenia GPU code to use the depth format in an "is float?" way, specifically to avoid handling the unreachable third option completely.

JoelLinn: default as fallthrough will just compile to a cmov(ne) https://godbolt.org/z/bqW9adEe1

Could still change the default to fall through to float

JoelLinn: So you insist on a ternary operator (plus assert before)?

JoelLinn: Also I don't think that effortlessly understanding switch fall-through is too much to ask from a gpu emulation developer

Triang3l: > So you insist on a ternary operator (plus assert before)?

In the first place, I don't want simple things to be overcomplicated. In the original version, there was some mask constant, and now there's the unnatural non-top-down switch case ordering that I haven't encountered in the Xenia code anywhere else, and which has probably never been used by most C++ programmers, with a deliberate fallthrough (itself not being a very common practice) — only to choose between two options, which is already done in many places throughout the Xenia GPU code (unless there's an invalid value like DXGI_FORMAT_UNKNOWN which is mapped to wrong input, that is already handled gracefully for a different reason and doesn't need additional effort, in this case, a true three-way switch is used).

JoelLinn: Changed as requested

Triang3l (approved):

PR #1903, opened 2021-10-14
[Kernel] Added %L to formatter

Gliniak:
Based on [58410826](https://github.com/xenia-project/game-compatibility/issues/304):

It uses formatter for string: ``memory://%08LX,%LX``

Simple addition of 'L' to formatter resolves issues with this title.

Feel free to propose better PR/Commit name 🙇 Thanks

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_strings.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_strings.cc
@@ -290,6 +290,13 @@ int32_t format_core(PPCContext* ppc_context, FormatData& data, ArgList& args,
           }
           state = FS_Type;
           continue;
+        } else if (c == 'L') {
+          // 58410826 incorrectly uses 'L' instead of 'l'.
+          // TODO(gibbed): L appears to be treated as an invalid token by
+          // xboxkrnl, investigate how invalid tokens are processed in xboxkrnl
+          // formatting when state FF_Type is reached.
+          state = FS_Type;
+          continue;
         } else if (c == 'h') {
           flags |= FF_IsShort;
           state = FS_Type;


Gliniak:
> Given the usage here in the format string, it's probably meant to be `%lX`, ie `FF_IsLong` not `FF_IsLongLong`?
> 
> `FF_IsLongLong` is meant for 64-bit values, where that format string explicitly formats to `%08X`, so 32-bit.
> 
> IE, the change should _probably_ be:
> 
> ```c++
>         if (c == 'l' || c == 'L') {
>           if (data.peek(0) == 'l' || data.peek(0) == 'L') {
> ```

I was thinking about it, but I saw that L is reference to type long long unsigned that is why I marked it that way (I didn't though at all of how console handles it)

Sure I will change it, but maybe it will be better to use ``tolower(c)``?

Gliniak:
> > Ok! Understandable. Changing now
> 
> I've posted one comment about mixed-case L in one formatting token — does `%LX` work for you normally on Windows (for numbers above 2^32), so `%LlX` and `%lLX` could also be checked?

Both `%LlX` and `%lLX` gets warning and returns %LX or %lX

Triang3l:
> Both `%LlX` and `%lLX` gets warning and returns %LX or %lX
What do they print for a number that can't fit in 32 bits at runtime if you ignore the warning?

Gliniak:
@Triang3l  Based on online compiler
image

gibbed:
How are values getting formatted without any handling for `L` in Xenia that it breaks `58410826`?

Triang3l:
Oh, interesting, what about other incorrect character? But `%LLX` doesn't work on OnlineGDB, yes.

gibbed:
> Oh, interesting, what about other invalid characters? But `%LLX` doesn't work on OnlineGDB, yes.

I haven't gone into that depth yet, I just know it's ignoring that token. Either what's happening is `format("memory://%08LX,%LX", 123, 123)` results in `memory://00000123X,123X` or it results in `memory://0000007B,7B`, that much I've determined (the former is likely I think).

I've put it on my TODO list to investigate, the change I've suggested will work for now.

Actually @Gliniak can you add a `TODO(gibbed): investigate this in xboxkrnl` after the existing comment?

Gliniak:
Done

gibbed:
> I'd add some info about the character being ignored in xboxkrnl to the TODO comment, however (but @gibbed can word it better than me), so anyone who wants to investigate that will know roughly what to look for and what surprises to expect.

Sorry @Gliniak, but Triangle is right. Amend the comment to (and make sure it lints OK):

```cpp
// TODO(gibbed): L appears to be treated as an invalid token by xboxkrnl, investigate how invalid tokens are processed in xboxkrnl formatting when state FF_Type is reached.
```

PR #1904, opened 2021-10-18
[APU] Skip audio header when there is no valid input

Gliniak:
Thanks Cancerous1/Randprint for initial research in this topic!

### Point of research:
image

r24 - Amount of invalid input buffers
r26 - Input offset
Interesting is ``lbz r10, 0x40(r11)`` I might be wrong, but that would mean that it checks a byte value from 10th dword from ``XMA_CONTEXT_DATA`` structure? Which is reserved/unknown?

### Research & changes
Initially I though that in such case output buffer should be invalidated, but after longer analysis and breaking game on console I figured out that it doesn't work like that. Instead it sets input offset to 0x20 (aka skips header) and go on.
Additionally I changed size of header constant because it doesn't match with xma_helpers and that offset.

### Concerns
What about context with no packets? Should it also set it to 0x20 or be invalidated completely?

### What it resolved:
This fixes buzzing sound in certain titles.

### What it breaks:
- https://github.com/xenia-project/game-compatibility/issues/293 - Causes to stuck before going ingame, but goes wrong even earlier

### DIFF
--- a/src/xenia/apu/xma_context.h
+++ b/src/xenia/apu/xma_context.h
@@ -134,7 +134,7 @@ class XmaContext {
  public:
   static const uint32_t kBytesPerPacket = 2048;
   static const uint32_t kBitsPerPacket = kBytesPerPacket * 8;
-  static const uint32_t kBitsPerHeader = 33;
+  static const uint32_t kBitsPerHeader = 32;
 
   static const uint32_t kBytesPerSample = 2;
   static const uint32_t kSamplesPerFrame = 512;
--- a/src/xenia/apu/xma_context.cc
+++ b/src/xenia/apu/xma_context.cc
@@ -301,7 +301,9 @@ void XmaContext::Decode(XMA_CONTEXT_DATA* data) {
 
   // No available data.
   if (!data->input_buffer_0_valid && !data->input_buffer_1_valid) {
-    data->output_buffer_valid = 0;
+    // 4156081D checks specifically for offset 0x20 when both input buffers
+    // are invalid.
+    data->input_buffer_read_offset = kBitsPerHeader;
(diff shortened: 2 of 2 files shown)


halotroop2288 (approved):

src/xenia/apu/xma_context.h
```diff
@@ -134,7 +134,7 @@ class XmaContext {
  public:
   static const uint32_t kBytesPerPacket = 2048;
   static const uint32_t kBitsPerPacket = kBytesPerPacket * 8;
-  static const uint32_t kBitsPerHeader = 33;
+  static const uint32_t kBitsPerHeader = 32;
```
Triang3l, line 137: If we have information on the contents of the header, it would be useful to describe the meaning of the bits, or to refer to some place in the code where the header is already documented some way.

Triang3l (changes requested):

src/xenia/apu/xma_context.h
halotroop2288: 32. 0x20. A nice round power of 2. Makes sense. Is that all that's holding this PR back?

Gliniak: Meaning of these bits is described here: https://github.com/xenia-project/xenia/blob/master/src/xenia/apu/xma_helpers.h

However there is a small regression due to that PR in https://github.com/xenia-project/game-compatibility/issues/293 which causes it to stuck on loading screen and that is why it waits

Gliniak:
I'm closing this PR. This behaviour is not correct (current one isn't correct either), but part about size of header is still valid.

I will create new PR with this and more fixes as soon as I will finish research and testing

PR #1908, opened 2021-11-01
[Kernel/Notify] Resolved missing version and mask_index

Gliniak:
As we talked on Discord.

RIght now created ``XNotificationKey`` lacks ``version`` and ``mask_index`` due to fields order in union (Thanks @Triang3l for sharing that wisdom).

That change resolves this issue and allows notifications from different groups to be corectly interpreted.

### DIFF
--- a/src/xenia/kernel/xnotifylistener.h
+++ b/src/xenia/kernel/xnotifylistener.h
@@ -30,13 +30,8 @@ union XNotificationKey {
   };
   XNotificationID id;
 
-  static constexpr XNotificationID get_id(uint8_t mask_index,
-                                          uint16_t local_id) {
-    XNotificationKey key = {};
-    key.mask_index = mask_index;
-    key.local_id = local_id;
-    return key.id;
-  }
+  constexpr XNotificationKey(XNotificationID notification_id)
+      : id(notification_id){};
 };
 
 class XNotifyListener : public XObject {
--- a/src/xenia/kernel/xnotifylistener.cc
+++ b/src/xenia/kernel/xnotifylistener.cc
@@ -32,7 +32,7 @@ void XNotifyListener::Initialize(uint64_t mask, uint32_t max_version) {
 }
 
 void XNotifyListener::EnqueueNotification(XNotificationID id, uint32_t data) {
-  auto key = XNotificationKey{id};
+  auto key = XNotificationKey(id);
   // Ignore if the notification doesn't match our mask.
   if ((mask_ & uint64_t(1ULL << key.mask_index)) == 0) {
     return;


src/xenia/kernel/xnotifylistener.h
```diff
@@ -22,13 +22,13 @@ namespace xe {
 namespace kernel {
 
 union XNotificationKey {
+  XNotificationID id;
```
Triang3l, line 25: This is not something obvious to rely on (and also you may want to initialize bit fields specifically instead), it's better to specify the intention to set the `id` clearly in the problematic place.

Triang3l (changes requested):

src/xenia/kernel/xnotifylistener.h
Gliniak: Indeed. I also don't like the fact that it is opaque and required knowledge of this fact, but in terms of overall code structure it seems to be a better way to handle this.

Surely if this is such a problematic case then I can initialize it normally or even better create a constructor for it?

Triang3l: Yes, a constructor (maybe not even `explicit` so XNotificationID can be passed to functions expecting XNotificationKey directly, but not sure) for creation from a complete notification ID would be nice, as this seems to be an important special case.

src/xenia/kernel/xnotifylistener.h
```diff
@@ -30,13 +30,8 @@ union XNotificationKey {
...
 
-  static constexpr XNotificationID get_id(uint8_t mask_index,
-                                          uint16_t local_id) {
-    XNotificationKey key = {};
-    key.mask_index = mask_index;
-    key.local_id = local_id;
-    return key.id;
-  }
+  constexpr XNotificationKey(XNotificationID notification_id)
+      : id(notification_id){};
```
Triang3l, line 34: For consistency with other parts of Xenia (except for `reenter_exception`…), please add a space between the `)` and the `{` and remove the `;`.

Triang3l (changes requested):

src/xenia/kernel/xnotifylistener.h
Triang3l: Also, this structure requires a proper `XNotificationID` for creation now. Please add:
```cpp
constexpr XNotificationKey() : id(XNotificationID(0)) {}
```
so it can still be build manually as a `struct` by assigning values to the bit fields (but zeroing by default is still needed so all unset bits are zeroed).

Alternatively, `id` can have a default value of `= XNotificationID(0)` specified (and the constructor without arguments should be `= default`), but in this case, as far I know, `id` must be the first field, and that's pretty easy to forget (see my message with the text "Zero-initialized to clear irrelevant bits" in #dev on the Discord server).

On the other hand though, it's better to have `id` as the first field that is `= 0` by default so we can have other constructors which would also have all unused bits zeroed (because fields are initialized in the order they're declared in the class — so if you have both `id(0)` and `local_id(something)` in the initializer list, with the current code, `local_id` will be initialized first, and will be erased by `id` zeroing).

Triang3l (changes requested):

src/xenia/kernel/xnotifylistener.h
```diff
@@ -30,13 +30,8 @@ union XNotificationKey {
...
   XNotificationID id;
 
-  static constexpr XNotificationID get_id(uint8_t mask_index,
-                                          uint16_t local_id) {
-    XNotificationKey key = {};
-    key.mask_index = mask_index;
-    key.local_id = local_id;
-    return key.id;
-  }
+  constexpr XNotificationKey(XNotificationID notification_id)
```
Triang3l, line 33: Additionally, `constexpr operator XNotificationID() const { return id; }` may be nice to have, so the two types can be used interchangeably, for example, while passing them to functions.

Or maybe we should just drop `XNotificationID` at all, and have `constexpr` IDs build via the `uint32_t` constructors? Since the two types mean pretty much the same thing. But this needs bigger changes, such as adding `operator==`, `operator!=`.

Triang3l (changes requested):

Gliniak:
Resolved in https://github.com/xenia-project/xenia/pull/1909

PR #1909, opened 2021-11-01
[Code] Make union usage more consistent

Triang3l:
Unions of bit fields and underlying storage (which may need to have unused bits set to a deterministic value, such as 0, for use as a lookup key or to avoid leaking data) now always have the storage type specified as the first element. Previously this was inconsistent (the underlying type was the last mostly just in my code). Though most of my assumptions about the order turned out not to match the reality, this makes it harder to accidentally initialize a bit field union as `{field1, field2, …}`, which would leave the unused bits undefined.

`= 0` defaults for the underlying types are not used anymore, as they're misleading. There have been cases when I assumed that this would result in zeroing of unused bits when you have a constructor that sets up individual fields using an initializer list. However, the initializer list causes the `= 0` default to be completely ignored. Also, on GCC, you can't have multiple initializers for aliasing union members in a single initializer list. This pull request also provides an example of a field-initializing constructor in `xe::gpu::PrimitiveConverter::CacheKey`.

Most unions of this kind now have a size assertion without duplication of the name of the underlying type, in the second compilation pass (during function compiling) so they can access both the size of the union and a named field within that union.

This pull request also supersedes #1908, also fixing the issue covered by it.

### DIFF
--- a/src/xenia/cpu/backend/x64/x64_op.h
+++ b/src/xenia/cpu/backend/x64/x64_op.h
@@ -46,6 +46,7 @@ enum KeyType {
 
 #pragma pack(push, 1)
 union InstrKey {
+  uint32_t value;
   struct {
     uint32_t opcode : 8;
     uint32_t dest : 5;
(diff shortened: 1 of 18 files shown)


gibbed (approved):
No complaints here, looks good!

PR #1911, opened 2021-11-12
Rename control_flow_analysis_pass.cpp to control_flow_analysis_pass.cc

rerpha:
All of the (non third-party) cpp impl files use the .cc extension, this one doesn't. I was digging through the code and found this one so thought I might as well rename it whilst I'm here!


Triang3l:
That's code doing nothing though and probably not useful that will be gone in a pretty close future. I'd expect OpSelectionMerge and OpPhi to be treated not very differently overall by the compiler in the driver (at least by the AMD one), with the main factor being whether the condition is uniform (likely a jump if yes, write masking and possibly a uniformity test and a jump if not) in the wavefront.

PR #1913, opened 2021-11-17
[XAM/Net] Implementation of getsockname & unification of error handling

Gliniak:
I'm waiting for rough review 💃 

If you have any better place where to put:  ``GetLastWSAError`` I would be glad to change its location.

### Benefits
- This fixes Evolution engine games


src/xenia/kernel/xsocket.cc
```diff
@@ -257,5 +257,25 @@ bool XSocket::QueuePacket(uint32_t src_ip, uint16_t src_port,
...
+
+  int ret = getsockname(native_handle_, &sa, (socklen_t*)buf_len);
+
+  if (!ret) {
+    std::memcpy(buf, &sa, *buf_len);
+  }
+  return ret;
+}
+
+const uint32_t XSocket::GetLastWSAError() const {
```
Triang3l, line 271: I'm not sure if this platform implementation detail — the fact that the host sockets expose a thread-local variable storing the last error — should be revealed to the platform-independent `XSocket` interface. Though we don't have something like `xsocket_posix.cc`/`xsocket_win.cc`, and this file is heavily specific to BSD sockets in its current form which do rely on `errno`/`WSAGetLastError`, but still, it's not ruled out that some future implementation may have error codes reported immediately from the functions as return values or output arguments rather than via a TLS variable, for instance. We already have functions returning `X_STATUS` here — they can be changed to returning 0 on success or the extended error code on error (assuming, however, that the host doesn't for some reason set the last error to 0 on failure in certain cases — if that's the case though, we need to check what the guest expects, and either split the return value into two variables or replace such 0 with a true error code) so there's better isolation of methods in this interface without global state where it's not clearly needed.

Note that we should never ever go in the opposite direction, however (proxying the host `WSAGetLastError` directly to the guest instead of maintaining our own TLS variable on the guest side), as we may in theory use host sockets in implementations of other kernel functions (such as HTTP, device connection).

src/xenia/kernel/xam/xam_net.cc
```diff
@@ -626,8 +626,7 @@ dword_result_t NetDll_socket(dword_t caller, dword_t af, dword_t type,
   if (XFAILED(result)) {
     socket->Release();
 
-    uint32_t error = xboxkrnl::xeRtlNtStatusToDosError(result);
-    XThread::SetLastError(error);
+    XThread::SetLastError(socket->GetLastWSAError());
```
Triang3l, line 629: Not specific to the changes, but does the Xbox 360 have a common storage for both `GetLastError` and `NetDll_WSAGetLastError`? On Windows they're separate if I understand correctly — if that's the case on the console, we need to separate them on the guest as well (though I'm not sure if it has a predefined location in the thread context or just goes to `TlsAlloc` when `WSAStartup` is called — needs investigation).

Triang3l (changes requested):

src/xenia/kernel/xsocket.h
```diff
@@ -116,6 +116,9 @@ class XSocket : public XObject {
                N_XSOCKADDR_IN* from, uint32_t* from_len);
   int SendTo(uint8_t* buf, uint32_t buf_len, uint32_t flags, N_XSOCKADDR_IN* to,
              uint32_t to_len);
+  int GetSockName(uint8_t* buf, int* buf_len);
+
+  const uint32_t GetLastWSAError() const;
```
Triang3l, line 121: Unnecessary `const` on the return type (the left one) that has no effect.

Triang3l (changes requested):

src/xenia/kernel/xam/xam_net.cc
```diff
@@ -1016,6 +1004,30 @@ void NetDll_WSASetLastError(dword_t error_code) {
...
+  auto socket =
+      kernel_state()->object_table()->LookupObject<XSocket>(socket_handle);
+  if (!socket) {
+    // WSAENOTSOCK
+    XThread::SetLastError(0x2736);
+    return -1;
+  }
+
+  int buffer_len = sizeof(sockaddr);
+  if (len_ptr) {
```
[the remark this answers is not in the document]
Gliniak: Good point! I saw somewhere that it was optional, but I cannot find it now :O 
If it is mandatory then it will be easier!

src/xenia/kernel/xam/xam_net.cc
Triang3l: According to @gibbed, no, but I think we need a comment in the documentation of NetDll_WSAGetLastError that it's actually shared on the console.

src/xenia/kernel/xsocket.cc
```diff
@@ -257,5 +257,26 @@ bool XSocket::QueuePacket(uint32_t src_ip, uint16_t src_port,
...
+  return X_STATUS_SUCCESS;
+}
+
+uint32_t XSocket::GetLastWSAError() const {
+  // Todo(Gliniak): Provide error mapping table
+  // Xbox error codes might not match with what we receive from OS
+#ifdef XE_PLATFORM_WIN32
+  return WSAGetLastError();
+#endif
+  return errno;
```
Triang3l, line 278: It's preferable to place this `return` in `#else` so there's no possibility of compilers warning about dead code, though I'll fix this.

src/xenia/kernel/xsocket.cc
```diff
@@ -257,5 +257,26 @@ bool XSocket::QueuePacket(uint32_t src_ip, uint16_t src_port,
...
+  }
+
+  std::memcpy(buf, &sa, *buf_len);
+  return X_STATUS_SUCCESS;
+}
+
+uint32_t XSocket::GetLastWSAError() const {
+  // Todo(Gliniak): Provide error mapping table
+  // Xbox error codes might not match with what we receive from OS
+#ifdef XE_PLATFORM_WIN32
```
Triang3l, line 275: We use `#if` generally for `XE_PLATFORM`, though the original code used `#ifdef` in a different place so that explains it.

src/xenia/kernel/xam/xam_net.cc
```diff
@@ -1031,6 +1008,28 @@ void NetDll_WSASetLastError_entry(dword_t error_code) {
...
+dword_result_t NetDll_getsockname_entry(dword_t caller, dword_t socket_handle,
+                                        lpvoid_t buf_ptr, lpdword_t len_ptr) {
+  auto socket =
+      kernel_state()->object_table()->LookupObject<XSocket>(socket_handle);
+  if (!socket) {
+    XThread::SetLastError(uint32_t(X_WSAError::X_WSAENOTSOCK));
+    return -1;
+  }
+
+  int buffer_len = *len_ptr;
```
Triang3l, line 1020: According to the [MSDN documentation for `getsockname`](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-getsockname):

> WSAEFAULT
> The name or the namelen parameter is not a valid part of the user address space, or the namelen parameter is too small.

I'm not sure if that's the case on the Xbox 360 exactly, but I think `WSAEFAULT` should be returned at least for nullptr `buf_ptr` or `len_ptr`, though generally since it's a kernel function it can have more thorough validation of pointers against the memory page tables.

However, a crash may be fine too since games aren't supposed to be doing that anyway. I was just looking if nullptr inputs needed to be handled specially, but looks like they don't need to.

Triang3l: Some validation, however, may be useful for consistency of error codes. On Windows, errors related to `namelen` are reported as `WSAEFAULT`, but on POSIX, they produce the `EINVAL` errno.

Also, there may be differences between the behavior in case the length is too small, but those need to be verified. On POSIX, the result is implicitly truncated if the length is too small, and the function probably succeeds (it only returns `EINVAL` if the length is invalid, like negative). On Windows, I'm not sure, but I can't find any info on MSDN about truncation, and the conditions of `WSAEFAULT` include "the `namelen` parameter is too small". However, I think if the function fails, it shouldn't write anything to any outputs (and `WSAEFAULT` doesn't distinguish between too small `namelen` and invalid `namelen`, so there's no fully reliable way to check if the function has actually written the full `namelen` needed — if the buffer is just too small, the function should probably succeed anyway) — but need to verify.

Though all of this probably should be handled on the `XSocket` side so it provides a more or less unified API between platforms, not in the kernel guest entry points.

src/xenia/kernel/xam/xam_net.cc
```diff
@@ -288,6 +288,9 @@ dword_result_t NetDll_WSACleanup_entry(dword_t caller) {
 }
 DECLARE_XAM_EXPORT1(NetDll_WSACleanup, kNetworking, kImplemented);
 
+// Instead of using dedicated storage for WSA error like on OS.
```
Triang3l, line 291: "like on Windows", as POSIX also seems to share error numbers (also this line and the next can be in one sentence).

Triang3l (changes requested):

PR #1914, opened 2021-11-23
Add Runtime Permissions

sahil-bhandari:
With the introduction of Android 6.0 (SDK 23), users are prompted for some specific permissions at runtime when they become necessary to use.

### DIFF
--- a/android/android_studio_project/app/src/main/AndroidManifest.xml
+++ b/android/android_studio_project/app/src/main/AndroidManifest.xml
@@ -26,6 +26,8 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
+        android:usesCleartextTraffic="true"
+        android:requestLegacyExternalStorage="true"
         android:supportsRtl="true"
         android:theme="@android:style/Theme.Material.Light">
 
--- a/android/android_studio_project/app/src/main/java/jp/xenia/emulator/WindowedAppActivity.java
+++ b/android/android_studio_project/app/src/main/java/jp/xenia/emulator/WindowedAppActivity.java
@@ -1,12 +1,17 @@
 package jp.xenia.emulator;
 
+import android.Manifest;
 import android.app.Activity;
+import android.app.AlertDialog;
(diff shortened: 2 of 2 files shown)


Triang3l:
Thanks for the contribution, but `WindowedAppActivity` is not the right place for the storage permission request — this is the base activity for windows displaying video output from the native emulation logic (or demos, most of which don't require the storage permission — possibly only for logging, but logging to a file likely won't be enabled by default on Android due to storage amount limitations, relying on Logcat instead — for the demos, logging is also mostly unnecessary), and the native emulation code assumes, in many places, that the permission to access the storage has already been obtained.

We may, however, at some point add some permission request logic for the main emulation activity to handle the case when the process was killed, the user has then denied the permission in the system settings, but later comes back to the task and relaunches the game via the last used game path/arguments stored in `savedInstanceState` — but the permission request will need to delay launching of the game in this case (this needs to be carefully integrated into the native game launching logic — possibly will even need to defer all native emulator loading, as even logging requires the permission). Or, as a much simpler alternative, we may even throw an error if the storage permission is unavailable in this case and return the user to the main menu instead of reloading the game, not much will be lost anyway as we won't be saving the actual game state in `onDestroy` as it's hundreds of megabytes large and we don't have state saving mechanism overall (it may also be complicated to implement due to usage of native host objects such as pthread ones for emulation of guest OS objects in Xenia), just the closest approximation — the launch path; the user can relaunch the game manually.

The primary path for requesting the storage permission (as well as optionally microphone/camera, though those will need to be requestable mid-game too) will be not in the emulation activity, but rather, in the launcher/settings activities, which are completely non-existent at this point in development (well, even the basic non-emulation demos requiring `xe::ui::WindowedAppContext`/`xe::ui::Window` functionality haven't been launched yet at the current development stage).

A toast, especially with mixed text case, also won't be the way we'll be providing feedback from the permission request flow — rather, it will be some fullscreen activity or a part of one, with user-friendly instructions (such as, if the user has chosen "don't ask again", what should be done in the system settings).

As a lot of prerequisites for this change haven't yet been met at the current development stage, I don't think this pull request may be going forward, since the logic proposed in it will need a completely different implementation and architecture.

PR #1919, opened 2021-12-15
[Kernel/Memory] Check for required protect_bits

Gliniak:
There are situations where titles can execute ``MmSetAddressProtect`` without one of basic protection flags. (Their state would be undefined!)

I was thinking about creating constexpr in header file somewhere instead of function body. Additionally feel free to provide better name for it. Thanks 😸 

# Theory:
These protection changes should be omitted as there is no way to define new state.
I was thinking if it could be correct to just append these new flags to actual state of page, however we would need to check if every page is defined, as there might be situation where we try to change it over unallocated space and cause 🔥 
(Update: However note in ``BaseHeap::Protect`` suggest that indeed these cases should be omitted)

# Affected Titles:
https://github.com/xenia-project/game-compatibility/issues/1781
https://github.com/xenia-project/game-compatibility/issues/146


src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -424,7 +424,11 @@ DECLARE_XBOXKRNL_EXPORT2(MmQueryAddressProtect, kMemory, kImplemented,
 
 void MmSetAddressProtect(lpvoid_t base_address, dword_t region_size,
                          dword_t protect_bits) {
-  if (!protect_bits) {
+  constexpr uint32_t required_protect_bits =
+      X_PAGE_NOACCESS | X_PAGE_READONLY | X_PAGE_READWRITE |
```
Triang3l, line 428: These sound mutually exclusive — are there real cases when games provide multiple of these at once? If there are none, the function should fail if `xe::bit_count(protect_bits & required_protect_bits) != 1`.

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -424,7 +424,11 @@ DECLARE_XBOXKRNL_EXPORT2(MmQueryAddressProtect, kMemory, kImplemented,
...
+      X_PAGE_EXECUTE_READ | X_PAGE_EXECUTE_READWRITE;
+
+  if (!(protect_bits & required_protect_bits)) {
     XELOGE("MmSetAddressProtect: Failed due to incorrect protect_bits");
     return;
```
Triang3l, line 433: An assertion failure is also recommended to be able to catch games with invalid behavior more quickly in debug builds.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
Gliniak: Ohh great idea! 
I'm not aware of any game that was defining multiple bits at once.

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -425,8 +425,12 @@ DECLARE_XBOXKRNL_EXPORT2(MmQueryAddressProtect, kMemory, kImplemented,
...
 void MmSetAddressProtect_entry(lpvoid_t base_address, dword_t region_size,
                                dword_t protect_bits) {
-  if (!protect_bits) {
-    XELOGE("MmSetAddressProtect: Failed due to incorrect protect_bits");
+  constexpr uint32_t required_protect_bits =
+      X_PAGE_NOACCESS | X_PAGE_READONLY | X_PAGE_READWRITE |
+      X_PAGE_EXECUTE_READ | X_PAGE_EXECUTE_READWRITE;
+
+  if (xe::bit_count(protect_bits & required_protect_bits) != 1) {
+    assert_false(xe::bit_count(protect_bits & required_protect_bits) > 1);
```
Triang3l, line 433: The bit count may also be 0 (smaller than 1) — is that a special case that is correct? I think `assert_always` (which also accepts a description string) should be fine since the condition is already checked outside.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
Gliniak: That was conscious decision, such case (0 bits) is quite common and I don't want to spam assertions in that particular case.

Triang3l: A comment about this case being common would be good for avoiding the same question in the future.

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -425,8 +425,14 @@ DECLARE_XBOXKRNL_EXPORT2(MmQueryAddressProtect, kMemory, kImplemented,
 
 void MmSetAddressProtect_entry(lpvoid_t base_address, dword_t region_size,
                                dword_t protect_bits) {
-  if (!protect_bits) {
-    XELOGE("MmSetAddressProtect: Failed due to incorrect protect_bits");
+  constexpr uint32_t required_protect_bits =
+      X_PAGE_NOACCESS | X_PAGE_READONLY | X_PAGE_READWRITE |
```
Triang3l, line 429: What are the exact bits set by those titles that aren't just zero, are those merely something like `X_PAGE_NOCACHE`/`X_PAGE_WRITECOMBINE` purely, or values like `X_PAGE_EXECUTE` also?

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -425,8 +425,14 @@ DECLARE_XBOXKRNL_EXPORT2(MmQueryAddressProtect, kMemory, kImplemented,
...
+      X_PAGE_EXECUTE_READ | X_PAGE_EXECUTE_READWRITE;
```
Triang3l, line 430: I think `X_PAGE_EXECUTE` should be allowed to as it's somewhat like `X_PAGE_NOACCESS` that you can execute, I think, by the way, as long as it's supported in other parts of the emulator's memory management.

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
```diff
@@ -425,8 +425,14 @@ DECLARE_XBOXKRNL_EXPORT2(MmQueryAddressProtect, kMemory, kImplemented,
...
+
+  if (xe::bit_count(protect_bits & required_protect_bits) != 1) {
+    // Many titles use invalid combination with zero valid bits set.
```
Triang3l, line 433: Can you please provide the title IDs and the bit values passed by those titles (and, if needed, reproduction steps) in the code comments so it's easier to look up the reason and regression test cases in the future (without having to search for the pull request, and without depending on a specific location of the repository on the web)?

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
Gliniak: Titles that I'm aware of (these two in description) use 0 or X_PAGE_NOCACHE, so it's quite difficult to precisely say

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
Gliniak: Added

src/xenia/kernel/xboxkrnl/xboxkrnl_memory.cc
Gliniak: Added

PR #1924, opened 2021-12-17
Add vsync_interval option

Margen67:
vsync_duration is no longer tied to vsync;
Old behavior was to set it to 16 if vsync is enabled, and 1 if disabled.
Default value is 16. (60 frames per second)
The value represents frametime in milliseconds. For example, for 120FPS, 8 would be specified because 1000/120 = ~8.

Thanks to JoelLinn for the help.

### DIFF
--- a/src/xenia/gpu/gpu_flags.cc
+++ b/src/xenia/gpu/gpu_flags.cc
@@ -20,6 +20,9 @@ DEFINE_path(
 
 DEFINE_bool(vsync, true, "Enable VSYNC.", "GPU");
 
+DEFINE_uint64(vsync_interval, 16,
+              "VSYNC interval. Value is frametime in milliseconds.", "GPU");
+
 DEFINE_bool(
     gpu_allow_invalid_fetch_constants, false,
     "Allow texture and vertex fetch constants with invalid type - generally "
--- a/src/xenia/gpu/graphics_system.cc
+++ b/src/xenia/gpu/graphics_system.cc
@@ -7,6 +7,8 @@
  ******************************************************************************
  */
 
+#include <algorithm>
+
 #include "xenia/gpu/graphics_system.h"
 
 #include "xenia/base/byte_stream.h"
@@ -122,7 +124,8 @@ X_STATUS GraphicsSystem::Setup(cpu::Processor* processor,
   vsync_worker_running_ = true;
   vsync_worker_thread_ = kernel::object_ref<kernel::XHostThread>(
       new kernel::XHostThread(kernel_state_, 128 * 1024, 0, [this]() {
-        uint64_t vsync_duration = cvars::vsync ? 16 : 1;
+        uint64_t vsync_duration =
+            cvars::vsync ? std::max<uint64_t>(5, cvars::vsync_interval) : 1;
         uint64_t last_frame_time = Clock::QueryGuestTickCount();
         while (vsync_worker_running_) {
           uint64_t current_time = Clock::QueryGuestTickCount();
--- a/src/xenia/gpu/gpu_flags.h
+++ b/src/xenia/gpu/gpu_flags.h
@@ -18,6 +18,8 @@ DECLARE_path(dump_shaders);
 
 DECLARE_bool(vsync);
 
+DECLARE_uint64(vsync_interval);
+
 DECLARE_bool(gpu_allow_invalid_fetch_constants);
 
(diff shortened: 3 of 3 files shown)


illusion0001:
Shouldn't uint64 be double instead? 1000/16 = 62.5 FPS. 60.0FPS is 16.66666666666667.

halotroop2288 (approved):

halotroop2288 (approved):
I get the feeling there's a reason this hasn't been merged yet, but it would help if someone else reviewed this to make it known.

If there's not, then maybe just merge it?

src/xenia/gpu/graphics_system.cc
```diff
@@ -122,7 +124,8 @@ X_STATUS GraphicsSystem::Setup(cpu::Processor* processor,
   vsync_worker_running_ = true;
   vsync_worker_thread_ = kernel::object_ref<kernel::XHostThread>(
       new kernel::XHostThread(kernel_state_, 128 * 1024, 0, [this]() {
-        uint64_t vsync_duration = cvars::vsync ? 16 : 1;
+        uint64_t vsync_duration =
+            cvars::vsync ? std::max<uint64_t>(5, cvars::vsync_interval) : 1;
```
Triang3l, line 128: Why is an arbitrary limit needed (that isn't just prevention of division by zero)? 1 is a valid value, 2…4 should likely be as well. Possibly 0 even considering there's an explicit 1ms sleep.

Triang3l (changes requested):

src/xenia/gpu/graphics_system.cc
Margen67: @JoelLinn

JoelLinn: I remember we discussed this but have to look it up again 🙃

src/xenia/gpu/gpu_flags.cc
```diff
@@ -20,6 +20,9 @@ DEFINE_path(
 
 DEFINE_bool(vsync, true, "Enable VSYNC.", "GPU");
 
+DEFINE_uint64(vsync_interval, 16,
```
Triang3l, line 23: I fully agree with @illusion0001, I think we shouldn't be exposing an explicitly incorrect value, it should be either a floating-point 1000/60 (or 1/60), or, similar to how DXGI exposes the monitor's refresh rate, a numerator and a denominator (either for the refresh rate, in which case they would be 60 and 1, or for the interval, 1 and 60 — though this is merely a naming question).

Our vsync logic is mathematically broken anyway. We count the time _since the last wakeup_ that has resulted in a vsync signal. So, if sleep resolution is, for instance, 3ms, we may be triggering vsyncs at 18ms, 36ms, and so on. If the sleep timing resolution is 15ms, vsyncs will be at 30ms, 60ms. With 16ms resolution — roughly the resolution of the sleep timer on Windows without `timeBeginPeriod` — the actual interval will be pretty much random, sometimes 16ms and 32ms.

Instead, at every wakeup, we need to trigger `(current time - last vsync time) / vsync interval` signals, and add `signal count * vsync interval` to the last vsync time (but maybe limit the number of actual signals sent to the game to 1 in case the user specifies some extremely low value, or if it's significantly lowered mid-game, by loading the game config — though Windows's ~16ms timer resolution without `timeBeginPeriod` may result in 2 signals for a wakeup, for instance, if wakeups happen at 15ms and at 15ms+18ms, the limit should be 2 in this case probably). This will make sure that if a wakeup happened, for instance, 18ms after the last vsync, the next will happen after at least 14ms, not 16ms. Note that this must be done with fixed-point values (integers — microseconds or nanoseconds), not with floating-point values, first, because absolute time is involved (so precision doesn't drop over time), and also so after we add `signal count * vsync interval` to the last vsync time, dividing the result by the vsync interval always gives 0, not sometimes 1 due to rounding.

Since time overall needs to be calculated as fixed-point, that's why it would be highly preferable to expose the interval as a rational number, not as a number (arbitrarily) rounded to a 23e8/53e11 float.

Also, in scenarios where the system's sleep resolution is too low (UWP on Windows, for instance, without `timeBeginPeriod`), we may need to fall back to some other kind of waiting (a yielding busy loop) — @JoelLinn should know a lot about this, I think.

Triang3l (changes requested):

Issue #665, opened 2017-01-31
Save states freeze Xenia


Margen67:
Happens with GL4 and Vulkan renderer, and the same game.
--protect_zero=false doesn't do anything.
There's no crash. Xenia just hangs and doesn't respond if I try to save or load a savestate. (F7/F8)

I haven't tested any older builds so I'm not sure if it's a regression.

Sazpaimon:
I went back and checked older builds to see when this stopped working, and it appears that 6777ce6668fe9010b78b7bcf4e7365bfbc844ef0 was what broke saving savestates.

Parovozik:
This is a specific error when the game does not see the connected hard drive. There should be a "dialogue" for selecting a storage device, but probably it is not completed in the emulator.

Parovozik:
There some system messages and i think this not works on Xenia on current state
1 - Screenshot from Warrior Orochi 3 on Xenia (after any choose infinity loading)
other from x360 concole
xenia message:
![bandicam 2018-06-21 16-15-09-524](https://user-images.githubusercontent.com/19775644/41728564-c4636afc-757f-11e8-9cc9-c550834204f3.jpg)
x360 1st message
![20180621181538 mp4_snapshot_00 09_ 2018 06 21_18 16 46](https://user-images.githubusercontent.com/19775644/41728581-d170084a-757f-11e8-8bb7-0bf511062a11.jpg)
x360 2nd message:
![20180621181538 mp4_snapshot_00 12_ 2018 06 21_18 16 57](https://user-images.githubusercontent.com/19775644/41728586-d3b414d4-757f-11e8-8118-c08602b62dd1.jpg)
x360 3d message:
![20180621181538 mp4_snapshot_00 17_ 2018 06 21_18 17 08](https://user-images.githubusercontent.com/19775644/41728600-d787ce2a-757f-11e8-8f6c-16b5cdc5bb7e.jpg)
x360 4th OK:
![20180621181538 mp4_snapshot_00 21_ 2018 06 21_18 17 18](https://user-images.githubusercontent.com/19775644/41728604-da0e3832-757f-11e8-8f90-554289c7f8a2.jpg)

Margen67:
@Parovozik This issue is about save *state*s, **not** saves.
Unless you mean that save states don't work because xenia doesn't reconnect the hard drive once a save state is loaded.

Parovozik:
[...]

PR #1925, opened 2021-12-17
[App] Disable stuff that crashes the emulator

Margen67:
Save states are broken (#665) and swapping discs isn't supported so all these do is cause crashes.

### DIFF
--- a/src/xenia/app/emulator_window.cc
+++ b/src/xenia/app/emulator_window.cc
@@ -118,6 +118,7 @@ bool EmulatorWindow::Initialize() {
       case ui::VirtualKey::kF5: {
         GpuClearCaches();
       } break;
+#ifdef DEBUG
       case ui::VirtualKey::kF7: {
         // Save to file
         // TODO: Choose path based on user input, or from options
@@ -130,6 +131,7 @@ bool EmulatorWindow::Initialize() {
         // TODO: Spawn a new thread to do this.
         emulator()->RestoreFromFile("test.sav");
       } break;
+#endif  // #ifdef DEBUG
       case ui::VirtualKey::kF11: {
         ToggleFullscreen();
       } break;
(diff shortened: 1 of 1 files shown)


src/xenia/app/emulator_window.cc
```diff
@@ -186,9 +186,9 @@ bool EmulatorWindow::Initialize() {
     file_menu->AddChild(
         MenuItem::Create(MenuItem::Type::kString, "&Open...", "Ctrl+O",
                          std::bind(&EmulatorWindow::FileOpen, this)));
-    file_menu->AddChild(
-        MenuItem::Create(MenuItem::Type::kString, "Close",
-                         std::bind(&EmulatorWindow::FileClose, this)));
+//    file_menu->AddChild(
```
Triang3l, line 189: I'd recommend `#if 0` to avoid breaking formatting. I'm not sure if this will even pass clang-format.

Triang3l (changes requested):

src/xenia/app/emulator_window.cc
Margen67: Like this?
```
      #if 0
      case ui::VirtualKey::kF7: {
        // Save to file
        // TODO: Choose path based on user input, or from options
        // TODO: Spawn a new thread to do this.
        emulator()->SaveToFile("test.sav");
      } break;
      case ui::VirtualKey::kF8: {
        // Restore from file
        // TODO: Choose path from user
        // TODO: Spawn a new thread to do this.
        emulator()->RestoreFromFile("test.sav");
      } break;
      #endif
```

Triang3l: Yes, though possibly without indentation for `#if 0` and `#endif`, but clang-format will make it correct.

Margen67: Done.

src/xenia/app/emulator_window.cc
```diff
@@ -118,6 +118,7 @@ bool EmulatorWindow::Initialize() {
       case ui::VirtualKey::kF5: {
         GpuClearCaches();
       } break;
+#if 0
```
gibbed, line 121: I'd say make these `#ifdef DEBUG` instead, since end-users are using the release build.

gibbed (changes requested):

gibbed (approved):

Triang3l (approved):

PR #1928, opened 2021-12-20
[Kernel] Added missing month to RtlTimeFieldsToTime

Gliniak:
Additionally added check for highest possible month day

Source: https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtltimefieldstotime (like from above implementation in Xenia)

``Month
Specifies a value from 1 to 12.`` but we had it up to 11 only.

# Affected Titles
https://github.com/xenia-project/game-compatibility/issues/29 - Resolves: Xbox live connection lost error

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
@@ -529,9 +529,10 @@ DECLARE_XBOXKRNL_EXPORT1(RtlTimeToTimeFields, kNone, kImplemented);
 dword_result_t RtlTimeFieldsToTime(pointer_t<X_TIME_FIELDS> time_fields_ptr,
                                    lpqword_t time_ptr) {
   if (time_fields_ptr->year < 1601 || time_fields_ptr->month < 1 ||
-      time_fields_ptr->month > 11 || time_fields_ptr->day < 1 ||
-      time_fields_ptr->hour > 23 || time_fields_ptr->minute > 59 ||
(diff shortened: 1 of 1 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_rtl.cc
```diff
@@ -529,9 +529,10 @@ DECLARE_XBOXKRNL_EXPORT1(RtlTimeToTimeFields, kNone, kImplemented);
 dword_result_t RtlTimeFieldsToTime(pointer_t<X_TIME_FIELDS> time_fields_ptr,
                                    lpqword_t time_ptr) {
   if (time_fields_ptr->year < 1601 || time_fields_ptr->month < 1 ||
-      time_fields_ptr->month > 11 || time_fields_ptr->day < 1 ||
-      time_fields_ptr->hour > 23 || time_fields_ptr->minute > 59 ||
-      time_fields_ptr->second > 59 || time_fields_ptr->milliseconds > 999) {
+      time_fields_ptr->month > 12 || time_fields_ptr->day < 1 ||
+      time_fields_ptr->day > 31 || time_fields_ptr->hour > 23 ||
```
Triang3l, line 533: Should we verify the number of days in the month according to the month and the year?

Gliniak: No idea how it works under the hood, but we can clearly define that day of month never should be higher than 31. 
I guess we could run few tests on windows to see what will happen, but it still would be a blackbox test

Triang3l:
Still, tests for 28-, 29- and 30-day long months would be useful.

PR #1930, opened 2021-12-27
[XModule] Remove module and its functions while unloading

Gliniak:
### Summary
RIght now if title have dlls that are being loaded and unloaded within a same addressing space there is a chance that two functions share same "starting point", which now causes issue that Xenia do not unload initially loaded function and executes first one instead of actually allocated one.

If I forget about anything at least it is super easy to add inside ``Processor::RemoveModule``

### Concerns
 - As Rick told on Discord: "What about code cache?" - No idea seems like this part is not affected by it?
 - I tried to put ``RemoveModule`` execution within: ``UnregisterModule``, but it is unloaded already and doesn't have name at that point. Any idea how to resolve it better?
 - Naming for ``Module::GetAddressedFunctions``. I don't like it and would change it, but no idea how to call it. Any ideas?
 - Returning status of removal? But how even we can handle it?

### Affected Titles
- https://github.com/xenia-project/game-compatibility/issues/563
- https://github.com/xenia-project/game-compatibility/issues/575
- https://github.com/xenia-project/game-compatibility/issues/577 (Doesn't guest crash anymore)
- Probably more


THEKILERFRIEND:
did you find a fix to this?

PR #1931, opened 2021-12-28
[GPU] Implemented D1MODE_V_COUNTER

Gliniak:
### Subject (probably) to change

I have lack of knowledge what exactly D1MODE and D2MODE is, but what I can see that value should be changed constantly. No idea if there are any conditions to stop/start these counters.

### Affected Titles:
https://github.com/xenia-project/game-compatibility/issues/710


src/xenia/gpu/graphics_system.cc
```diff
@@ -118,6 +118,20 @@ X_STATUS GraphicsSystem::Setup(cpu::Processor* processor,
       reinterpret_cast<cpu::MMIOReadCallback>(ReadRegisterThunk),
       reinterpret_cast<cpu::MMIOWriteCallback>(WriteRegisterThunk));
 
+  // GPU counters update thread
+  auto counters_worker_thread = kernel::object_ref<kernel::XHostThread>(
+      new kernel::XHostThread(kernel_state_, 128 * 1024, 0, [this]() {
+        while (1) {
+          WriteRegister(
```
Triang3l, line 125: This looks like a race condition between the new thread and the command processor, though not sure how it's handled with MMIO writes currently.

src/xenia/gpu/graphics_system.cc
```diff
@@ -118,6 +118,20 @@ X_STATUS GraphicsSystem::Setup(cpu::Processor* processor,
       reinterpret_cast<cpu::MMIOReadCallback>(ReadRegisterThunk),
       reinterpret_cast<cpu::MMIOWriteCallback>(WriteRegisterThunk));
 
+  // GPU counters update thread
+  auto counters_worker_thread = kernel::object_ref<kernel::XHostThread>(
+      new kernel::XHostThread(kernel_state_, 128 * 1024, 0, [this]() {
+        while (1) {
```
Triang3l, line 124: This is going to be the most CPU-intensive thread in the entire emulator likely, plus the units of the counter are completely random now, just tied to the host CPU/memory/OS performance.

Triang3l (changes requested):

src/xenia/gpu/graphics_system.cc
```diff
@@ -187,8 +201,6 @@ uint32_t GraphicsSystem::ReadRegister(uint32_t addr) {
       return 0x08100748;
     case 0x0F01:  // RB_BC_CONTROL
       return 0x0000200E;
-    case 0x194C:  // R500_D1MODE_V_COUNTER
-      return 0x000002D0;
```
Triang3l, line 191: If we need this to provide a value that changes at some rate, we need to read the system clock at register reads and to remap it to the expected frequency, not to take 100% of one host CPU core.

Triang3l (changes requested):

src/xenia/gpu/graphics_system.cc
Gliniak: I agree with it. That is why I marked it as "SUBJECT TO CHANGE" and wrote that probably there should be some conditions when/how often it is iterated 💃

Gliniak:
I'm closing this PR until better solution is found

PR #1934, opened 2021-12-29
[x64] Add `GFNI`-based optimization for `VECTOR_SHA_V128(Int8)`

Wunkolo:
In the `Int8` case of `VECTOR_SHA_V128`, when all the values are the same, then a single-instruction `gf2p8affineqb` can be emitted that does an int8-based arithmetic-shift, utilizing GF(8) arithmetic.

More info here:
https://wunkolo.github.io/post/2020/11/gf2p8affineqb-int8-shifting/

As of now(Dec 2021): Tremont(Lakefield), Jasper Lake, Ice lake, Tigerlake, and Rocket Lake support GNFI.


gibbed:
Minor nitpicking: `ull` (or `ULL`) suffix on constants?

Wunkolo:
Remarking as a draft since we need a new pattern around conditionally emitting these instructions, as this is breaking the `use_haswell_instructions`-conditional that is guarding this `GFNI`-feature.

Wunkolo:
This is ready to go now that #1952 is merged. A new pattern for testing all the `x64_extension_mask` bits can come in a later PR so it's not so noisy there.

PR #1943, opened 2022-01-07
[xboxkrnl] Fix up XexGetProcedureAddress logging.

gibbed:
[xboxkrnl] Fix up `XexGetProcedureAddress` failure logging.

### DIFF
--- a/src/xenia/kernel/xboxkrnl/xboxkrnl_modules.cc
+++ b/src/xenia/kernel/xboxkrnl/xboxkrnl_modules.cc
@@ -166,7 +166,15 @@ dword_result_t XexGetProcedureAddress(lpvoid_t hmodule, dword_t ordinal,
       *out_function_ptr = ptr;
       result = X_STATUS_SUCCESS;
     } else {
-      XELOGW("ERROR: XexGetProcedureAddress ordinal not found!");
+      if (is_string_name) {
+        XELOGW("ERROR: XexGetProcedureAddress export '{}' in '{}' not found!",
+               string_name, module->name());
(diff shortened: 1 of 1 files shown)


src/xenia/kernel/xboxkrnl/xboxkrnl_modules.cc
```diff
@@ -166,7 +166,13 @@ dword_result_t XexGetProcedureAddress(lpvoid_t hmodule, dword_t ordinal,
       *out_function_ptr = ptr;
       result = X_STATUS_SUCCESS;
     } else {
-      XELOGW("ERROR: XexGetProcedureAddress ordinal not found!");
+      if (is_string_name) {
+        XELOGW("ERROR: XexGetProcedureAddress export '{}' in '{}' not found!",
+               string_name, module->name());
+      } else {
+        XELOGW("ERROR: XexGetProcedureAddress ordinal {:X} in '{}' not found!",
```
Triang3l, line 173: This isn't clear if the ordinal doesn't have any A-F characters in it — if we need it as hexadecimal, possibly it should be prefixed with 0x.

Triang3l (changes requested):

src/xenia/kernel/xboxkrnl/xboxkrnl_modules.cc
gibbed: Fair enough, I'll just log in both dec and hex since it's more or less Xenia convention to refer to ordinals in hex (ie export tables, etc).

PR #1952, opened 2022-01-11
[CPU] Remove `use_haswell_instructions` in favor of `x64_extension_mask`

Wunkolo:
Rather than having a single bool to conditionally detect haswell-level
instruction features. The granularity is increased with a new
`x64_extension_mask` where individual features within the x64 backend
can be turned on or off in a bit-mask manner. Since we have an ARM
backend on the horizon, I've added this to the new `x64`
configuration-group rather than `CPU`. This new pattern will hopefully
allow for testing to be more targetted to certain processor features and
allows the user to determine if they want certain features to be enabled
or disabled(such as avoiding BMI2 on certain AMD processors due to
pdep/pext being incredibly slow). The default configuration is to allow
for haswell-level features, to recreate the fact that the default
configuration for `use_haswell_instructions` is `true`.

Handling a bitmask in the configuration toml is a little awkward. I'm open to breaking each of these features up into their own bool as well. Whatever allows the granularity to be increased.

### DIFF
--- a/src/xenia/cpu/backend/x64/x64_backend.cc
+++ b/src/xenia/cpu/backend/x64/x64_backend.cc
@@ -26,10 +26,23 @@
 #include "xenia/cpu/processor.h"
 #include "xenia/cpu/stack_walker.h"
 
-DEFINE_bool(
-    use_haswell_instructions, true,
-    "Uses the AVX2/FMA/etc instructions on Haswell processors when available.",
-    "CPU");
+DEFINE_int32(x64_extension_mask, -1,
+             "Allow the detection and utilization of specific instruction set "
+             "features.\n"
+             "    0 = x86_64 + AVX1\n"
+             "    1 = AVX2\n"
+             "    2 = FMA\n"
+             "    4 = LZCNT\n"
+             "    8 = BMI1\n"
+             "   16 = BMI2\n"
+             "   32 = F16C\n"
+             "   64 = Movbe\n"
+             "  128 = AVX512F\n"
+             "  256 = AVX512VL\n"
+             "  512 = AVX512BW\n"
+             " 1024 = AVX512DQ\n"
+             "   -1 = Detect and utilize all possible processor features\n",
+             "x64");
 
 namespace xe {
 namespace cpu {
@@ -84,7 +97,7 @@ bool X64Backend::Initialize(Processor* processor) {
   }
 
   // Need movbe to do advanced LOAD/STORE tricks.
-  if (cvars::use_haswell_instructions) {
+  if (cvars::x64_extension_mask & kX64EmitMovbe) {
     machine_info_.supports_extended_load_store =
         cpu.has(Xbyak::util::Cpu::tMOVBE);
   } else {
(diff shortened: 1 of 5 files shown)


gibbed:
This looks okay but I'm not sure I'm a fan of directly mapping the enum here like that. That being said, when I eventually add array support to cvars, we can migrate to that then and it won't be an issue.

I also think we don't really need to default here to Haswell specifically, rather instead default to all available. The cvar being Haswell-specific before was purely because it was the only thing we currently took advantage of.

Wunkolo:
> This looks okay but I'm not sure I'm a fan of directly mapping the enum here like that
Me neither. I'd rather allow for some kind of human intuition rather than having to calculate magic numbers just because you want to disable BMI2 or whatever. If something later on can make this much easier to handle then this field can be the first client of that. I'd much rather use names like `x86-64-v2` and `x86-64-v4` and `native` and such rather than awkwardly expose an enum to our config files.

> I also think we don't really need to default here to Haswell specifically, rather instead default to all available. The cvar being Haswell-specific before was purely because it was the only thing we currently took advantage of.

For sure. I'll default to `-1` instead.

gibbed:
> If something later on can make this much easier to handle then this field can be the first client of that

It's on my list to implement array support for cvars.

https://toml.io/en/v1.0.0#array

Something like:
```toml
x64_extensions = [ "AVX2", "BMI1" ]
```

src/xenia/cpu/backend/x64/x64_emitter.cc
```diff
@@ -74,23 +74,32 @@ X64Emitter::X64Emitter(X64Backend* backend, XbyakAllocator* allocator)
       backend_(backend),
       code_cache_(backend->code_cache()),
       allocator_(allocator) {
-  if (cvars::use_haswell_instructions) {
+  if (cvars::x64_extension_mask & kX64EmitAVX2)
```
Triang3l, line 77: A bit too late, but we use brackets for all conditionals/loops.

src/xenia/cpu/backend/x64/x64_backend.cc
```diff
@@ -26,10 +26,23 @@
...
 #include "xenia/cpu/stack_walker.h"
 
-DEFINE_bool(
-    use_haswell_instructions, true,
-    "Uses the AVX2/FMA/etc instructions on Haswell processors when available.",
-    "CPU");
+DEFINE_int32(x64_extension_mask, -1,
+             "Allow the detection and utilization of specific instruction set "
+             "features.\n"
+             "    0 = x86_64 + AVX1\n"
```
Triang3l, line 32: Has `xb format` been done between `git add -A` and `git commit`? Arguments containing multi-line string literals are usually moved to a separate, indented line rather than aligned by it.

src/xenia/cpu/backend/x64/x64_emitter.cc
gibbed: Ah shit, I didn't even notice that. Probably should refactor it anyway and use function to do the bit setting to make the code less noisy.

Issue #1936, opened 2022-01-02
Xenia crashes on start-up when Wayland is used as the display server


danemadsen:
![Screenshot from 2022-01-02 20-43-02](https://user-images.githubusercontent.com/11537699/147873297-79536bae-0ebf-44e3-9d49-736acd7e213b.png)
[xenia.log](https://github.com/xenia-project/xenia/files/7798935/xenia.log)
Seems to be an issue with GTK when Wayland is in use. When Xenia is started on a Linux Desktop with Wayland running as the display server it immediately crashes with no display. This issue isn't present when X11 is in use. 

There doesn't seem to be anything interesting in the log but ill attach it anyway.

PR #1955, opened 2022-01-12
[UI] Force use of Xwayland when running on Wayland

thrimbor:
This adds a workaround to make Xenia use Xwayland when creating a GTK window using Wayland, instead of creating a Wayland window and then triggering an `assert` in the code.

I contributed pretty much the same change to ares a while back: https://github.com/ares-emulator/ares/pull/9

Please note that I only made sure that Xenia starts properly with this change. I can't test any further due to this system not having a fully Vulkan-compatible driver.

Closes https://github.com/xenia-project/xenia/issues/1936

### DIFF
--- a/src/xenia/ui/windowed_app_main_posix.cc
+++ b/src/xenia/ui/windowed_app_main_posix.cc
@@ -18,6 +18,10 @@
 #include "xenia/ui/windowed_app_context_gtk.h"
 
 extern "C" int main(int argc_pre_gtk, char** argv_pre_gtk) {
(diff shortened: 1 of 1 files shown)


Triang3l:
I think we'll be able to add true Wayland support more or less easily later, by the way (once I switch from native window handles to the Surface class).

Triang3l:
Thanks for the fix! I didn't expect that to suddenly break in the first place considering that there's probably a lot of software that assumes that GTK runs on X11.

PR #1959, opened 2022-01-21
[Base, aarch64] `copy_and_swap` implementations

JoelLinn:
probably one of the last things on the priority list for an aarch64 port

### DIFF
--- a/src/xenia/base/memory.cc
+++ b/src/xenia/base/memory.cc
@@ -2,7 +2,7 @@
  ******************************************************************************
  * Xenia : Xbox 360 Emulator Research Project                                 *
  ******************************************************************************
- * Copyright 2014 Ben Vanik. All rights reserved.                             *
+ * Copyright 2022 Ben Vanik. All rights reserved.                             *
  * Released under the BSD license - see LICENSE in the root for more details. *
  ******************************************************************************
  */
@@ -10,6 +10,10 @@
 #include "xenia/base/memory.h"
 #include "xenia/base/cvar.h"
 #include "xenia/base/platform.h"
+
+#if XE_ARCH_ARM64
+#include <arm_neon.h>
+#endif
 
 #include <algorithm>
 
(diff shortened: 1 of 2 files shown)


src/xenia/base/memory.cc
```diff
@@ -215,7 +219,102 @@ void copy_and_swap_16_in_32_unaligned(void* dest_ptr, const void* src_ptr,
...
+                              size_t count) {
+  copy_and_swap_16_unaligned(dst_ptr, src_ptr, count);
+}
+
+void copy_and_swap_16_unaligned(void* dst_ptr, const void* src_ptr,
+                                size_t count) {
+  auto dst = reinterpret_cast<uint16_t*>(dst_ptr);
+  auto src = reinterpret_cast<const uint16_t*>(src_ptr);
+
+  const uint8x16_t tbl_idx = vcombine_u8(vcreate_u8(0x0607040502030001),
```
Triang3l, line 240: For safety across compilers — though maybe literals of this width are fine everywhere, but to explicitly disambiguate anyway — wrapping these literals in `UINT64_C` (from `<cstdint>`) may be desirable.

src/xenia/base/memory.cc
```diff
@@ -215,7 +219,102 @@ void copy_and_swap_16_in_32_unaligned(void* dest_ptr, const void* src_ptr,
...
+                                         vcreate_u8(0x0E0F0C0D0A0B0809));
+
+  int64_t i;
```
Triang3l, line 243: Signedness is not needed here — `size_t` is the preferred type for indices in C++, and the comparison can be changed to `i + 7 < count`.

src/xenia/base/memory.cc
```diff
@@ -215,7 +219,102 @@ void copy_and_swap_16_in_32_unaligned(void* dest_ptr, const void* src_ptr,
...
+void copy_and_swap_16_in_32_aligned(void* dst, const void* src, size_t count) {
+  return copy_and_swap_16_in_32_unaligned(dst, src, count);
+}
+
+void copy_and_swap_16_in_32_unaligned(void* dst_ptr, const void* src_ptr,
+                                      size_t count) {
+  auto dst = reinterpret_cast<uint64_t*>(dst_ptr);
+  auto src = reinterpret_cast<const uint64_t*>(src_ptr);
+  for (size_t i = 0; i < count; ++i) {
+    dst[i] = (src[i] >> 16) | (src[i] << 16);
```
Triang3l, line 312: Is vtbl more or less like pshufb, or is it not enough for 128-bit 16-in-32 swapping?

src/xenia/base/memory.cc
```diff
@@ -215,7 +219,102 @@ void copy_and_swap_16_in_32_unaligned(void* dest_ptr, const void* src_ptr,
...
+// processing.
+
+void copy_and_swap_16_aligned(void* dst_ptr, const void* src_ptr,
+                              size_t count) {
+  copy_and_swap_16_unaligned(dst_ptr, src_ptr, count);
+}
+
+void copy_and_swap_16_unaligned(void* dst_ptr, const void* src_ptr,
+                                size_t count) {
+  auto dst = reinterpret_cast<uint16_t*>(dst_ptr);
```
Triang3l, line 237: `static_cast` is enough for `void*` <> pointer.

Triang3l (changes requested):

src/xenia/base/memory.cc
JoelLinn: I did that on purpose to simplify the loop condition check. (count + x) can be precomputed and I think the compiler can't make that optimization (from (i + 8 <= count) to (i <= count + 7)) itself because it requires typecasting size_t -> int64_t.

src/xenia/base/memory.cc
JoelLinn: vtbl is like shuffle, yes. I didn't bother to implement it since it's only used in vulkan code in one place afaik.
Also since it's not used in D3D12 right now, I figured it might not be needed anymore eventually.

Also if its 16 in 32 then why are the `dst` and `src` pointers of type `uint64_t*`??? I guess @bwrsandman stumbled across this as well and hence didn't add tests for it.

src/xenia/base/memory.cc
JoelLinn: `static_cast` can adjust the address when casting from `void*` to a pointer of any type, for example to satisfy alignment requirements. [see 10)](https://en.cppreference.com/w/cpp/language/static_cast)

`reinterpret_cast` is guaranteed (weird architectures aside) to not interfere with the pointer (not generating any cpu instructions)

even if it might be ok on aarch64 I vote to use `reinterpret_cast`. Because it's _"more correct"_ and code like this get's copied around and might end up in our planned Alpha AXP support and not be fine anymore :/

src/xenia/base/memory.cc
Triang3l: Ah, yes, if I recall correctly, the gpu::PrimitiveProcessor has its own byte swapping logic combined with index buffer modifications. The vtbl code needs to be written there eventually as well, currently it uses vrev (if there's swapping code in it already at all). But in the main use case — GPUs without 32-bit vertex index support, in cases when manual index fetch cannot be done (such as with primitive restart) — raw copy and swap is not very useful anyway as the primitive restart index needs to be replaced with 0xFFFFFFFF as well while copying.

Regarding uint64_t — all I can say here probably is 🤔🤔🤔🙃🤷‍♂️

src/xenia/base/memory.cc
Triang3l: Is the generated code good if the loop counter is changed from the element index to the vector index?

src/xenia/base/memory.cc
JoelLinn: It should be seriously broken. garbling half of the values (in 32bit units) and writing over the buffer end

src/xenia/base/memory.cc
JoelLinn: didn't check that explicitly. we need element index for residual handling later anyways. 
I might rework it  so that it downcounts `count` like:

```c
while (count >= 1234) {
  // vector stuff , dst, src
  count -= 1234;
  dst += 1234 * sizeof();
  src += 1234 * sizeof();
}
while (count > 0) {
 // single bswap
 count--;
 dst += sizeof();
 src += sizeof();
}
```

That might be easier for a risc arch which doesn't have the crazy memory stride calculations (`LEA`) that x86 does.

JoelLinn: Sooo... armv8-a clang 11.0.1  -O2

inst_count
65              using size_t and swapping the condition like you suggested
60              current int64_t solution
53              solution from last comment

JoelLinn:
copy_and_swap_16_in_32 should be fixed now as well

Triang3l (approved):

PR #1961, opened 2022-01-23
[Kernel] Add PVR opcode (includes cvars support)

halotroop2288:
Adds the PVR opcode to the list of recognized opcodes. Includes useful comments.

***Revival of #1464 (deleted fork)***

### DIFF
--- a/src/xenia/cpu/cpu_flags.cc
+++ b/src/xenia/cpu/cpu_flags.cc
@@ -38,6 +38,13 @@ DEFINE_bool(
 DEFINE_bool(validate_hir, false,
             "Perform validation checks on the HIR during compilation.", "CPU");
 
+DEFINE_uint64(
+    pvr, 0x710700,
+    "Processor version and revision number.\nBits 0 to 15 are the version "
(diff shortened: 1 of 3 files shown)


Triang3l:
Please squash two commits into a single one, and run `xb format` between `git add` and `git commit` to ensure style consistency in the code, including correct line widths. Also, the meanings of the specific values are still missing from the cvar description.

halotroop2288:
image

Triang3l:
> image

It needs to be executed after adding the modified files to Git staging — it only formats the files that are displayed as green in `git status`.

halotroop2288:
> It needs to be executed after adding the modified files to Git staging — it only formats the files that are displayed as green in `git status`.

I know. There were no changes needed when I was done.

gibbed:
You can do `xb format --all` I think (at least `xb lint` has that option). Though it may result in unrelated files to this commit being formatted.

JoelLinn:
We use `clang-format` version 13. There will be subtle differences with older versions.

halotroop2288:
@Triang3l Have I satisfied your request?

halotroop2288:
*ahem* This is ready for review...

PR #1963, opened 2022-01-23
[x64] Add `GFNI`-based optimization for `VECTOR_SH{R,L}_V128(Int8)` and tests

Wunkolo:
In the `Int8` case of `VECTOR_SH{R,L}_V128`, when all the values are the
same, then a single-instruction `gf2p8affineqb` can be emitted that does
an int8-based arithmetic-shift, utilizing GF(8) arithmetic.

More info here:
https://wunkolo.github.io/post/2020/11/gf2p8affineqb-int8-shifting/

Also fixes the iteration-type for when detecting if all of the simd
lanes are the same value(was iterating `u16` and not `u8`).

Also adds some tests for these cases

### DIFF
--- a/src/xenia/cpu/backend/x64/x64_seq_vector.cc
+++ b/src/xenia/cpu/backend/x64/x64_seq_vector.cc
@@ -742,9 +742,9 @@ struct VECTOR_SHL_V128
         }
         if (all_same) {
           // Every count is the same, so we can use gf2p8affineqb.
-          const uint8_t shift_amount = shamt.u8[0];
+          const uint8_t shift_amount = shamt.u8[0] & 0b111;
           const uint64_t shift_matrix =
-              0x0102040810204080 >> (shift_amount * 8);
+              UINT64_C(0x0102040810204080) >> (shift_amount * 8);
           e.vgf2p8affineqb(i.dest, i.src1,
                            e.StashConstantXmm(0, vec128q(shift_matrix)), 0);
           return;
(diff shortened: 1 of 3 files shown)


src/xenia/cpu/backend/x64/x64_seq_vector.cc
```diff
@@ -731,6 +731,25 @@ struct VECTOR_SHL_V128
...
+          if (shamt.u8[n] != shamt.u8[n + 1]) {
+            all_same = false;
+            break;
+          }
+        }
+        if (all_same) {
+          // Every count is the same, so we can use gf2p8affineqb.
+          const uint8_t shift_amount = shamt.u8[0];
+          const uint64_t shift_matrix =
+              0x0102040810204080 >> (shift_amount * 8);
```
gibbed, line 747: Missing `ull` (or `ULL`) on constants again. 🙂

gibbed (changes requested):

src/xenia/cpu/backend/x64/x64_seq_vector.cc
Triang3l: Or better `UINT64_C(0x0102040810204080)` since `ULL` is for `unsigned long long`, not specifically `uint64_t`.

Triang3l: By the way, can the shift amount overflow, and is that handled according to the PowerPC specification?

gibbed: That's fair, I wasn't aware of `UINT64_C`.

Wunkolo: > By the way, can the shift amount overflow, and is that handled according to the PowerPC specification?

chrome_Zubs3WCDOJ

[AltiVec](https://www.nxp.com/docs/en/reference-manual/ALTIVECPEM.pdf) doc seems to say that it just masks the lower `log2(n)` bits. I'll do the masking and const-literal stuff in a new commit.

Issue #1526, opened 2020-01-04
Log files are huge


wheelmac92:
While playing some Demos of games, 1 megabyte of log is written every 2 seconds. After playing for a while, the log is 700 megabytes large. When the game crashes, it feels inconvenient to upload it, because it is so large. Is there any way to narrow down the log to only the most relevant information? For example, if a game crashes after playing for an hour, is there any way to determine the last few minutes of gameplay? Does this happen in all games?

It happened for me while using the Dec 4 release of Xenia (latest) while playing Diablo 3, Pixar Up, and Sonic Unleashed. I can test more games if you'd like me to

The rapid speed of writing to log was happening during 3D scenes.
During 2D intro cutscenes, the logs were increasing at a slower rate

Razzile:
setting the command line argument `log_level` can help with this. using `--log_level=0` will only log errors. Also it's possible to zip (or otherwise compress) the log file which can shrink its size by sometimes as much as 10x.

Margen67:
Xenia should automatically compress logs like rpcs3 does to avoid this.

jpb18:
> Xenia should automatically compress logs like rpcs3 does to avoid this.

Think using gzip to compress the logs would be enough?

jpb18:
Also, another sugestion I've been thinking: instead of having a single monolithic file, why don't we split the log into several files as it goes along? Probably, if there's an issue, we don't need the entire 700 megabytes (example given before) log to track a problem. Maybe the last 100000 lines could be enough...

halotroop2288:
Maybe think about tracking the number of times a line or chunk gets repeated instead of printing them to the log repeatedly, as well?

A very simple implementation could look like this...
Here's some pseudocode:
```csharp
string log_buffer;
uint repetition_count = 0;

Main() {
	// An example
	Log("HELLO WORLD!");
	Log("HELLO WORLD!");
	Log("HELLO WORLD!");
	Log("HI UNIVERSE!");
	Log("GREETINGS GITHUB");
	Log("GREETINGS GITHUB");
	Log("HI UNIVERSE!");
}

Log(string message) {
	if (message == log_buffer.last(message.length)) { // Check if the message has already been logged once
		repetition_count++; // Increase the counter
		// Don't log yet
	} else {
		// If the message was repeated, add the counter to the line
		if (repetition_count > 0) {
			  // Add the message count to the line
			  log_buffer +=  " [x" + repetition_count + "]"; // Append counter string to the message line
		} else {
[...]

PR #1964, opened 2022-01-24
Update README to Reflect Merged Changes

halotroop2288:
#1333 has been closed, so the README should reflect that, quite like we did in #1743.
This line can be replaced with another suggestion from the `good-first-issue` label, such as #1526.


Triang3l:
Not exactly a simple topic as a lot of questions arise from the potential solutions, but at least it doesn't require Xbox 360 specifics, so why not.

PR #1966, opened 2022-01-24
Pages: update download URL to GitHub Releases

illusion0001:


Triang3l:
I think it's better (or, at least, fine for making the least changes for now) that it points to the archive itself rather than the release info page, since we have other files like xenia-vfs-dump there, and that may be confusing.

Issue #1089, opened 2018-02-25
UI - Frame Rate


zemzema:
It would be great if there is an option to show frame rate for current game.
 (ui)

Parovozik:
Just use Bandicam software. It correctly works on Vulkan

DiamondMaster07:
As well does Shadowplay(if you have an Nvidia card)

shadowbrony33:
AMD ReLive works great with Vulkan (Fuck bandicam because it's freenium ware)

DrChat:
This isn't too pressing of an issue, and microprofile already provides mspf. Hit F3 to see microprofile's overlay.

Margen67:
Fixed by #1362

illusion98:
it hasn't been merged yet.

ReaperOfSouls1909:
MERGE IT
![509](https://user-images.githubusercontent.com/18043258/61544237-5e090880-aa1b-11e9-9a96-e3b6cceb6b38.png)

TGLTClutch:
Why has it not been merged yet 😠 .

ReaperOfSouls1909:
![cortana_png_331929](https://user-images.githubusercontent.com/18043258/61544847-9eb55180-aa1c-11e9-9a89-b0eef307dc3f.png)
merge it or she gets it!! please thanks

ReaperOfSouls1909:
![cortana_png_331929](https://user-images.githubusercontent.com/18043258/61545263-6febab00-aa1d-11e9-814a-c0ac2a853612.png)
you made me do it i hope ya know that and banjo is next!


(haha im done hope it gets merged soon)

PR #1968, opened 2022-01-24
Add FPS counter to microprofiler

illusion0001:
Perhaps it's a better to use `fAverageMs`?

### DIFF
--- a/third_party/microprofile/microprofileui.h
+++ b/third_party/microprofile/microprofileui.h
@@ -2948,7 +2948,8 @@ void MicroProfileDrawMenu(uint32_t nWidth, uint32_t nHeight)
 		float fMs = fToMs * (S.nFlipTicks);
 		float fAverageMs = fToMs * (S.nFlipAggregateDisplay / nAggregateFrames);
 		float fMaxMs = fToMs * S.nFlipMaxDisplay;
-		int nLen = snprintf(FrameTimeMessage, sizeof(FrameTimeMessage)-1, "Time[%6.2f] Avg[%6.2f] Max[%6.2f]", fMs, fAverageMs, fMaxMs);
+		float fps = 1000 / fMs; // maybe fAverageMs is better cuz it's averages?
+		int nLen = snprintf(FrameTimeMessage, sizeof(FrameTimeMessage)-1, "FPS [%6.2f] Time [%6.2f] Avg [%6.2f] Max [%6.2f]", fps, fMs, fAverageMs, fMaxMs);
 		pMenuText[nNumMenuItems++] = &FrameTimeMessage[0];
 		MicroProfileDrawText(nWidth - nLen * (MICROPROFILE_TEXT_WIDTH+1), 0, -1, FrameTimeMessage, nLen);
 	}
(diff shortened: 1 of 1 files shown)


Margen67:
Fixes #1089

halotroop2288 (approved):

halotroop2288:
LGTM!

third_party/microprofile/microprofileui.h
```diff
@@ -2948,7 +2948,8 @@ void MicroProfileDrawMenu(uint32_t nWidth, uint32_t nHeight)
 		float fMs = fToMs * (S.nFlipTicks);
 		float fAverageMs = fToMs * (S.nFlipAggregateDisplay / nAggregateFrames);
 		float fMaxMs = fToMs * S.nFlipMaxDisplay;
-		int nLen = snprintf(FrameTimeMessage, sizeof(FrameTimeMessage)-1, "Time[%6.2f] Avg[%6.2f] Max[%6.2f]", fMs, fAverageMs, fMaxMs);
+		float fps = 1000 / fMs; // maybe fAverageMs is better cuz it's averages?
```
Triang3l, line 2951: - If floating-point numbers are used, it's better to clearly display that in the literals used in the code (`1000.0f` instead of the implicitly converted integer `1000`).
- "maybe" should be a "// TODO(illusion0001): ", but if `fAverageMs` just works now, there's no need to leave technical debt.
- Comments in Xenia code are generally placed on a separate line (but if on the same line, separated with 2 spaces).
- "maybe" → "Maybe", "cuz" → "because", "averages" → "averaged".

Overall though this feature should preferably be implemented in @zeux's upstream for easier updating in the future, possibly with a nicer division by zero handling (though displaying something like 1.#INF is possibly acceptable too).

Triang3l (changes requested):

third_party/microprofile/microprofileui.h
halotroop2288: > "cuz" → "because"

Code consistency is one thing, but why are you policing their spelling?

Triang3l: English language consistency 🙃

illusion0001: I did some testing, it's better to use `fAverageMs` here because converting from `fMs` every frame can produce incorrect results. Because this behavior does not average out the time it takes to draw a frame, one frame can be draw for 8.3ms, the other can spike up to 33, thus a sudden jump in "averaged" frame counting.

illusion0001: > If floating-point numbers are used, it's better to clearly display that in the literals used in the code (1000.0f instead of the implicitly converted integer 1000).

Done.

illusion0001: Though I should possibly account for if frame time is < 0 because during xenia startup, it does not count frames and leads to NaN.

halotroop2288: If it could show up during startup, it would probably just show a garbage number for NaN. But it can't because the screen it is used on has to be invoked after startup.

PR #1969, opened 2022-01-27
[App] rename flag to use config on debug flag

illusion0001:


src/xenia/app/emulator_window.cc
```diff
@@ -391,7 +391,8 @@ void EmulatorWindow::CpuBreakIntoDebugger() {
   if (!cvars::debug) {
     xe::ui::ImGuiDialog::ShowMessageBox(window_.get(), "Xenia Debugger",
                                         "Xenia must be launched with the "
```
Margen67, line 393:
```suggestion
                                        "Xenia must be launched with "
```

Razzile:
I don't think either of these are necessarily correct as this change makes it seem like you can't use the flag anymore. "Xenia must be launched with the debug config flag set" is better imo.

halotroop2288:
The most clear and concise way I can think of phrasing this is:
"Xenia must be launched with the debug flag enabled to enable debugging."

Or maybe just "Please restart Xenia with the debug flag enabled to use debugging."

But that's super redundant and really begs the question: If you have to show this redundant message box in the first place, perhaps you're just doing something wrong?

PR #1970, opened 2022-01-29
Help improvements

gibbed:
[App] Rename `ShowCommitID` to `ShowBuildCommit`.
[App] Replace Website help menu item with FAQ.
[App] Add Compatibility help menu item.


Margen67:
What do you think of adding an option to copy the title id+gamename?

gibbed:
@Margen67 If you mean like, to the clipboard, sure. That's beyond the scope of this PR. I don't recall if we have any clipboard related code in Xenia currently (Windows clipboard stuff is actually really annoying).

PR #1977, opened 2022-02-01
Replace system() with posix_spawn in LaunchFileExplorer.

apemax:
This commit replaces `system()` with `posix_spawn`.

### DIFF
--- a/src/xenia/base/system_gnulinux.cc
+++ b/src/xenia/base/system_gnulinux.cc
@@ -9,7 +9,9 @@
 
 #include <alloca.h>
 #include <dlfcn.h>
+#include <spawn.h>
 #include <stdlib.h>
+#include <unistd.h>
 
 #include <cstring>
 
@@ -21,6 +23,8 @@
 // Use headers in third party to not depend on system sdl headers for building
 #include "third_party/SDL2/include/SDL.h"
 
+extern char** environ;
+
 namespace xe {
 
 void LaunchWebBrowser(const std::string_view url) {
@@ -29,7 +33,19 @@ void LaunchWebBrowser(const std::string_view url) {
   system(cmd.c_str());
 }
 
-void LaunchFileExplorer(const std::filesystem::path& path) { assert_always(); }
+void LaunchFileExplorer(const std::filesystem::path& path) {
+  pid_t pid;
+  int n = path.string().length();
+  char cmdxdg[] = "xdg-open";
+  char cmdpath[n + 1];
+  char* argv[] = {cmdxdg, cmdpath, NULL};
+  int status = -1;
+  strcpy(cmdpath, path.c_str());
+  status = posix_spawn(&pid, "/usr/bin/xdg-open", NULL, NULL, argv, environ);
+  if (status != 0) {
+    perror("posix_spawn");
+  }
+}
 
 void ShowSimpleMessageBox(SimpleMessageBoxType type, std::string_view message) {
   void* libsdl2 = dlopen("libSDL2.so", RTLD_LAZY | RTLD_LOCAL);
--- a/src/xenia/base/system_gnulinux.cc
+++ b/src/xenia/base/system_gnulinux.cc
@@ -23,7 +23,9 @@
 // Use headers in third party to not depend on system sdl headers for building
 #include "third_party/SDL2/include/SDL.h"
 
+extern "C" {
 extern char** environ;
+}
 
 namespace xe {
 
@@ -35,15 +37,13 @@ void LaunchWebBrowser(const std::string_view url) {
 
 void LaunchFileExplorer(const std::filesystem::path& path) {
   pid_t pid;
-  int n = path.string().length();
-  char cmdxdg[] = "xdg-open";
-  char cmdpath[n + 1];
-  char* argv[] = {cmdxdg, cmdpath, NULL};
-  int status = -1;
-  strcpy(cmdpath, path.c_str());
-  status = posix_spawn(&pid, "/usr/bin/xdg-open", NULL, NULL, argv, environ);
+  std::string content_path = path.string();
+  char executable_name[] = "xdg-open";
+  char* const argv[] = {executable_name, content_path.data(), NULL};
+  int status =
+      posix_spawn(&pid, "/usr/bin/xdg-open", NULL, NULL, argv, environ);
   if (status != 0) {
-    perror("posix_spawn");
+    XELOGE("Failed to launch File Browser.");
   }
 }
 


src/xenia/base/system_gnulinux.cc
```diff
@@ -29,7 +29,11 @@ void LaunchWebBrowser(const std::string_view url) {
   system(cmd.c_str());
 }
 
-void LaunchFileExplorer(const std::filesystem::path& path) { assert_always(); }
+void LaunchFileExplorer(const std::filesystem::path& path) {
+  auto cmd = std::string("xdg-open ");
+  cmd.append(path);
```
Triang3l, line 34: All shell-interpreted characters, as well as spaces, possibly need to be escaped if they are actually interpreted if `system` is used — this applies to launching the web browser too.

Triang3l (changes requested):

src/xenia/base/system_gnulinux.cc
Triang3l: I think ideally this shouldn't go through the shell at all — is `posix_spawn` the right way of doing the same, but with explicit `argv`?

apemax: Because the input to `system`, `xdg-open`, the content directory path and specific urls is hard-coded essentially, and all of those work correctly there isn't really a need to escape anything, Unless I'm missing something?

Regarding using `system` or `posix_spawn`, `posix_spawn` would probably be the "better" way of doing it in that sense from what I have seen, I haven't used `posix_spawn` before though and am currently unable to get it to compile without complaining about converting the `path` and `xdg-open` strings to `char *` when it comes to `argv`. I will continue looking into this as it's probably just something I'm not quite getting right now.

Triang3l: The content directory path isn't hardcoded, it's by default relative to the user's home directory, but can be changed arbitrarily in the config.

apemax: Right, Makes sense. I have put together a commit to replace `system()` with `posix_spawn`. What do you think? I'll do the same with `LaunchWebBrowser` if your ok with this one.

Triang3l: If it works fine, and replicates the normal behavior of launching from `system` (such as, all environment variables are properly passed), that would be awesome.

Triang3l:
Can you please squash the two commits into one as the new one largely supersedes the first commit?

src/xenia/base/system_gnulinux.cc
apemax: Yep, The users environment variables are pulled in using the `environ` variable so xdg-open works correctly. Works the same as `system()` from the users perspective.

Razzile (changes requested):

src/xenia/base/system_gnulinux.cc
```diff
@@ -29,7 +33,19 @@ void LaunchWebBrowser(const std::string_view url) {
...
 
-void LaunchFileExplorer(const std::filesystem::path& path) { assert_always(); }
+void LaunchFileExplorer(const std::filesystem::path& path) {
+  pid_t pid;
+  int n = path.string().length();
+  char cmdxdg[] = "xdg-open";
+  char cmdpath[n + 1];
+  char *argv[] = {cmdxdg, cmdpath, NULL};
+  int status = -1;
+  strcpy(cmdpath, path.c_str());
```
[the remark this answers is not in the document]
apemax: Right, So for this specific application using strncpy would be fine? So like:

`strncpy(cmdpath, path.c_str(), path.string().length());`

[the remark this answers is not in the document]
Triang3l: This is equally as insecure, as you're disregarding the size of the _destination_ buffer as much as `strcpy` does. Also, we don't use the non-standard and also dangerous variable-length local arrays. Just create a new `std::string` and gather all the needed parts there, then use `c_str()`.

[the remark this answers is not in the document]
Triang3l: `strncpy` is overall broken by design, by the way, since, if I recall correctly, it doesn't insert the null terminator if overflow has happened. `strlcpy` is the safely truncating function, but it's non-standard. But overall, in Xenia, we don't hesitate to use dynamically allocated temporaries on non-hot paths as long that's more flexible, simple and safe.

src/xenia/base/system_gnulinux.cc
```diff
@@ -29,7 +33,19 @@ void LaunchWebBrowser(const std::string_view url) {
   system(cmd.c_str());
 }
 
-void LaunchFileExplorer(const std::filesystem::path& path) { assert_always(); }
+void LaunchFileExplorer(const std::filesystem::path& path) {
+  pid_t pid;
+  int n = path.string().length();
```
Triang3l, line 38: Implicit truncation from `size_t` to `int`.

src/xenia/base/system_gnulinux.cc
```diff
@@ -29,7 +33,19 @@ void LaunchWebBrowser(const std::string_view url) {
...
+  char cmdxdg[] = "xdg-open";
+  char cmdpath[n + 1];
+  char* argv[] = {cmdxdg, cmdpath, NULL};
+  int status = -1;
+  strcpy(cmdpath, path.c_str());
+  status = posix_spawn(&pid, "/usr/bin/xdg-open", NULL, NULL, argv, environ);
+  if (status != 0) {
+    perror("posix_spawn");
```
Triang3l, line 46: Except for the part of the platform-specific initialization where logging still hasn't yet been initialized, we use the Xenia logging (`XELOGE` from `"xenia/base/logging.h"`) for error messages, not `perror`.

src/xenia/base/system_gnulinux.cc
```diff
@@ -21,6 +23,8 @@
 // Use headers in third party to not depend on system sdl headers for building
 #include "third_party/SDL2/include/SDL.h"
 
+extern char** environ;
```
Triang3l, line 26: Is `extern "C"` not required for it? By the way, since it's needed only once, I think it's better to declare it at the local scope where it's used, so there won't be a conflict if some other header also declares `environ`.

src/xenia/base/system_gnulinux.cc
Triang3l: `n` is not a good variable name choice also, as well as `cmdxdg` and `cmdpath` — shortenings are generally not recommended unless the full word is very long, and words should be separated with an underscore.

src/xenia/base/system_gnulinux.cc
```diff
@@ -29,7 +33,19 @@ void LaunchWebBrowser(const std::string_view url) {
...
 }
 
-void LaunchFileExplorer(const std::filesystem::path& path) { assert_always(); }
+void LaunchFileExplorer(const std::filesystem::path& path) {
+  pid_t pid;
+  int n = path.string().length();
+  char cmdxdg[] = "xdg-open";
+  char cmdpath[n + 1];
+  char* argv[] = {cmdxdg, cmdpath, NULL};
+  int status = -1;
```
Triang3l, line 42: Declaring this variable upfront is not required as it's first assigned to in the same scope.

src/xenia/base/system_gnulinux.cc
apemax: Right, I was using `perror` as that outputs a description of the error, Not just the error number that is assigned to `status`. Will switch to XELOGE and have a look about getting the error description through that as well.

src/xenia/base/system_gnulinux.cc
apemax: Right, Realised `n` wouldn't even be necessary anyway, Will remove that and also change the name of the other two variables.

src/xenia/base/system_gnulinux.cc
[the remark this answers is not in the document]
Triang3l: `std::string.data()` is guaranteed to return a pointer to null-terminated memory since C++11, and since C++17 it returns a mutable string. C++17 is our target, so it should be fine as one of the `argv`.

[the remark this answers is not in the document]
gibbed: Additionally, the `argv` should be made `const` for clarity (`posix_spawn` takes a `const` for its `argv` argument).

apemax:
Implemented all corrections apart from the `extern char** environ;` being global, However it seems like the only way for it to work is by being global. `getenv` can be used but it would require a lot more code to get all the required environment variables. Any thoughts on this or anything else outstanding?

JohnHenry150 (commented):
There are a LOT of issues with this code.

First off, including <alloca.h> is a big no-no. That shit is unsafe and can lead to all sorts of buffer overflows and memory leaks. Same goes for "<cstring>" , which is basically the devil's playground for string manipulation. 

Secondly, why the fuck are you using third-party headers for SDL2? Don't you know that's like playing Russian roulette with your codebase? Just use the damn system SDL headers like a responsible programmer, would ya?

And finally, what's up with that "extern char** environ;" at the end? Are you trying to make me cry? That's some old-school C shit right there, and it's not even necessary in modern code. 

Fix your fucking code.

Razzile:
@JohnHenry150 This tone is not appreciated on this project, nor are any of the issues you've mentioned relevant to the PR in any way

PR #1979, opened 2022-02-03
[x64] Add AVX512 optimization for `VECTOR_ROTATE_LEFT(Int32)`

Wunkolo:
`vprolvd` is an almost 1:1 analog with this opcode and can be
conditionally emitted when the host supports `AVX512{F,VL}`.

Altivec docs say that `vrl{bhw}` masks the lower log2(n) bits of the
element-size.
image

[vprold](https://www.felixcloutier.com/x86/vprold:vprolvd:vprolq:vprolvq) modulos the shift-value by the element size in bits, which is the same
as masking the lower log2(n) bits. So `vrlw` maps exactly to `vprold`.

Also edited the unit tests to test the overflow-case when the shift amount is larger than the element size.


JoelLinn (changes requested):
Please change the copyright date of the three files you touched to 2022.

Apart from that this is an amazing improvement of the vector rotate!
Unfortunately this instruction is probably much less called than other vector fp math.

Normally I would suggest to check real xbox360 for overflow case but I am too lazy to boot it up...

JoelLinn (approved):
looks good

gibbed (approved):

PR #1985, opened 2022-02-12
Reducing Size of Photos in Repo

adgsenpai:


Triang3l:
I don't think this is desirable. Git is decentralized — when you clone a repository, you download _its entire commit history_, so any changes will only _increase_ the size of the repository, especially if the content is changed significantly (such as, the actual image data, not just the headers — but I don't even know how Git even handles situations with small changes to binary files, Git is designed for text in the first place), you may end up with more or less twice the original image size.

halotroop2288:
> but I don't even know how Git even handles situations with small changes to binary files

Git stores a zipped copy *every single change that's ever been commited*. That's probably why Xenia's git repo is so large when it's cloned, and if you told me that's why you're so anal about squashing history, I'd have to believe you!

Here's a video if you want to learn more about how it works:

[YouTube Link](https://youtu.be/bSA91XTzeuA)

As for this change, while it won't make the repo smaller, it will make downloads of snapshots (the "download as zip" option) *ever so slightly* smaller. Not worth it!

Triang3l:
Decreasing the size of the Android icons (as long as they still can be displayed on a device afterwards — I've had issues with opening images from some web platforms in Photoshop, I think, but can't remember the details) may be nice, however, but currently they're temporary, and if possible, they'll be changed to vector ones.

PR #1988, opened 2022-02-16
Improve vrefp & vrsqrtefp accuracy

JoelLinn:
Improving vrefp accuracy is especially important for 5454082B as it fixes:

- Grass flickering
- Cactus collision
- Random events (bandits, "pls help me", ...)
- Camp spawning
- Horse spawning

That particular title was affected at load time. The objects where probably loaded incorrectly into memory, vrefp was not involved in those issues once in-game.


Linear524:
Finally those accuracy bugs and logic bugs was fixed in RDR !!! (^_^)
Thank you million times, JoelLinn ! :)

P.S. - so I suppose RDR now fully playable ))
P.P.S. - maybe this will also fix Gears of War script bugs after boss fights as well, need some time to test ))

PR #1989, opened 2022-02-17
[Wildcard] Match exact string if it doesn't contain any pattern

Gliniak:
TLDR:
This resolves such situation:

Provided pattern: ``abc.file``
Files in directory: ``aaabc.file, abc.file``

Actual behavior:
Returned file is: ``aabc.file``

Expected Behavior: 
Returned file is: ``abc.file``

This resolves:
https://github.com/xenia-project/game-compatibility/issues/480 - Requirement for protect_zero (was loading incorrect shader files)
https://github.com/xenia-project/game-compatibility/issues/1650 - Wasn't booting at all

### DIFF
--- a/src/xenia/base/filesystem_wildcard.cc
+++ b/src/xenia/base/filesystem_wildcard.cc
@@ -19,6 +19,7 @@ namespace xe::filesystem {
 WildcardFlags WildcardFlags::FIRST(true, false, false);
 WildcardFlags WildcardFlags::LAST(false, true, false);
 WildcardFlags WildcardFlags::ANY(false, false, true);
+WildcardFlags WildcardFlags::FIRST_AND_LAST(true, true, false);
 
 WildcardFlags::WildcardFlags()
     : FromStart(false), ToEnd(false), ExactLength(false) {}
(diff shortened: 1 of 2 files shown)


halotroop2288 (approved):

src/xenia/base/filesystem_wildcard.cc
```diff
@@ -89,7 +89,9 @@ void WildcardEngine::PreparePattern(const std::string_view pattern) {
   }
   if (last != pattern.size()) {
     std::string str_str(pattern.substr(last));
-    rules_.push_back(WildcardRule(str_str, WildcardFlags::LAST));
+    rules_.push_back(
+        WildcardRule(str_str, last ? WildcardFlags::LAST
+                                   : WildcardFlags(true, true, false)));
```
gibbed, line 94: For clarity, maybe define a `FIRST_OR_LAST` up with the others and use that here.

gibbed (changes requested):

src/xenia/base/filesystem_wildcard.cc
Gliniak: For now I will only add  ``FIRST_AND_LAST`` (or ``MATCH_STRING``/``MATCH_PATTERN``) as it is used. 
If there will be a need for different combinations then we can add them.

PR #1991, opened 2022-02-21
[x64] Add GFNI optimization for SPLAT(int8)

Wunkolo:
`pxor` is a zero-uop register-rename and `gf2p8affineqb dest, zero, int8`
is a very quick single-instruction way to use affine galois
transformations to fill a register with an immediate byte without
touching memory.


Wunkolo:
Just pinging this PR. Tested on my Icelake laptop and Rocketlake desktop.

PR #1992, opened 2022-02-21
Fix posix threading stuff 😿

### LINKED ISSUE #1678: [Linux, threading] Events may deadlock whith auto_reset
From PR #1317 :

https://github.com/xenia-project/xenia/blob/d1f7ee35933b1eb0ba8a87c406538a8e81a30662/src/xenia/base/threading_posix.cc#L282-L293

@bwrsandman :
> It would seem that Events have a bug when they're created as auto reset. More tests are required to have feature parity with Win32.
> 
> For example, launching xenia and quitting will produce a deadlock.
> 
> The following fixes this deadlock but renders the behaviour the same as as manual reset event and this goes against MS documentation.
> 
> ```diff
> diff --git a/src/xenia/base/threading_posix.cc b/src/xenia/base/threading_posix.cc
> index 216d6714..9f1334e5 100644
> --- a/src/xenia/base/threading_posix.cc
> +++ b/src/xenia/base/threading_posix.cc
> @@ -284,7 +284,7 @@ class PosixCondition<Event> : public PosixConditionBase {
>      if (manual_reset_) {
>        cond_.notify_all();
>      } else {
> -      cond_.notify_one();
> +      cond_.notify_all();
>      }
>      return true;
>    }
> ```

JoelLinn:
~~Fixes #1678~~ (there are other pending issues with the posix threading code) 

In the general scope of #1430 

This is an attempt to fix the spurious Linux CI fails we are seeing.
I mainly attribute them to the misuse of `cond_.notify_one()` (see #1678)

Additionally I replaced a lot of `Sleep` calls used for synchronisation with spin waits because timing is very unreliable on CI due to heavy over-provisioning.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -277,7 +277,7 @@ def pipeline_linux_desktop(name, image, arch, cc, build_release_all):
                 'image': image,
                 'volumes': [volume_build('premake')],
                 'commands': [
-                    'valgrind --error-exitcode=99 ./build/bin/Linux/Debug/xenia-base-tests',
+                    'valgrind --error-exitcode=99 ./build/bin/Linux/Debug/xenia-base-tests --durations yes',
                 ],
                 'depends_on': ['build-premake-debug-tests'],
             },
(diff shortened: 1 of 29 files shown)


JoelLinn:
appveyor 🥲

JoelLinn:
This has become a bit bigger than I anticipated. I want to fix other issues in the threading code but that would definitely blow this up.

Drone has spurious hangs, since they are in the test as well as during compilation I attribute this to the drone infrastructure. But the tests should be much more robust against slow runners now.

gibbed (approved):

PR #1994, opened 2022-02-23
Add descriptive error message when pkg-config fails

amessier:
Small enhancement when calling pkg-config on package that is not installed. This makes it easier to see what is missing and what to install!

Inspired by the [SDL2 premake helper](https://github.com/xenia-project/xenia/blob/6b45cf84472c26436d4a5db61a7b50dab301e398/third_party/SDL2.lua#L20).

### DIFF
--- a/tools/build/scripts/pkg_config.lua
+++ b/tools/build/scripts/pkg_config.lua
@@ -2,12 +2,21 @@
 
 pkg_config = {}
 
+local function pkg_config_call(lib, what)
+  local result, code = os.outputof("pkg-config --"..what.." "..lib)
+  if result then
+    return result
+  else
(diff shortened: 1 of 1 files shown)


JoelLinn (approved):
@gibbed We should merge this simple fix.

Right now if you miss a lib that is imported using pkg-config you get:
`Error: .../premake-core/src/base/string.lua:36: attempt to index a nil value (upvalue 's')`
which is not very helpful and hard to debug. Also not very welcoming for new devs

pjsoberoi:
Can confirm this bit me just now. Please fix. I somehow copy-pasta-ed the package list incorrectly.

pjsoberoi:
Thanks!

PR #1998, opened 2022-03-07
[CPU] Added constant propagation pass for: OPCODE_AND_NOT

Gliniak:


src/xenia/cpu/hir/value.cc
```diff
@@ -750,6 +750,12 @@ void Value::Not() {
   }
 }
 
+void Value::AndNot(Value* other) {
+  assert_true(type == other->type);
+  other->Not();
```
Triang3l, line 755: This will modify the `other` passed to the function with an effect on the caller — is this what is expected from this function? I think we should create a local copy of `other` for this purpose if that's possible.

Triang3l (changes requested):

src/xenia/cpu/hir/value.cc
Triang3l: I also think that for this purpose, `other` should be `const` in every function where `other` is not expected to be modified.

Gliniak: I agree! To be perfectly clean we should use local copy then on it execute Not()

```
Value* second = new Value(*other);
second->Not();
And(second);
```

About constness. Sure but in different PR maybe?

Triang3l: In the example you've provided, `second` will never be destroyed.

src/xenia/cpu/compiler/passes/constant_propagation_pass.cc
```diff
@@ -648,6 +648,14 @@ bool ConstantPropagationPass::Run(HIRBuilder* builder, bool& result) {
             result = true;
           }
           break;
+        case OPCODE_AND_NOT:
+          if (i->src1.value->IsConstant() && i->src2.value->IsConstant()) {
+            v->set_from(i->src1.value);
+            v->AndNot(i->src2.value);
```
Triang3l, line 654: To keep the `Value` class simpler and to avoid introducing a somewhat possibly unsupported (although I'm not yet sure, but `Value` contains more fields than what `set_from` copies, and I'm not sure if you can just duplicate them) concept of constructing a `Value` from another `Value`, what do you think about implementing the "and not" logic here instead?
```cpp
v->set_from(i->src2.value);
v->Not();
v->And(i->src1.value);
```

Triang3l (changes requested):

src/xenia/cpu/compiler/passes/constant_propagation_pass.cc
Gliniak: Thanks! Yeah. It seems reasonable for it to be there. It's only used here and it is short and easy to understand anyway. Changed & Removed dedicated AndNot

PR #2000, opened 2022-03-09
Python/xb/xenia-build fixes

Margen67:
See commits.

### DIFF
--- a/xb.bat
+++ b/xb.bat
@@ -1,5 +1,5 @@
 @ECHO OFF
-REM Copyright 2015 Ben Vanik. All Rights Reserved.
+REM Copyright 2022 Ben Vanik. All Rights Reserved.
 
 SET "DIR=%~dp0"
 
@@ -7,10 +7,12 @@ REM ============================================================================
 REM Environment Validation
 REM ============================================================================
 
+SET "PYTHON_MINIMUM_VERSION[0]=3"
+SET "PYTHON_MINIMUM_VERSION[1]=6"
 CALL :check_python
 IF %_RESULT% NEQ 0 (
   ECHO.
-  ECHO Python 3.6+ must be installed and on PATH:
+  ECHO Python %PYTHON_MINIMUM_VERSION[0]%.%PYTHON_MINIMUM_VERSION[1]%+ must be installed and on PATH:
   ECHO https://www.python.org/
   GOTO :eof
 )
@@ -33,9 +35,12 @@ SETLOCAL ENABLEDELAYEDEXPANSION
 
 SET FOUND_PATH=""
 
-SET "CANDIDATE_PATHS[0]=C:\python37\python.exe"
-SET "CANDIDATE_PATHS[1]=C:\python36\python.exe"
-SET OUTPUT_INDEX=2
+SET "CANDIDATE_PATHS[0]=C:\python310\python.exe"
+SET "CANDIDATE_PATHS[1]=C:\python39\python.exe"
+SET "CANDIDATE_PATHS[2]=C:\python38\python.exe"
+SET "CANDIDATE_PATHS[3]=C:\python37\python.exe"
+SET "CANDIDATE_PATHS[4]=C:\python%PYTHON_MINIMUM_VERSION[0]%%PYTHON_MINIMUM_VERSION[1]%\python.exe"
+SET OUTPUT_INDEX=5
 
 FOR /F "usebackq delims=" %%L IN (`2^>NUL where python3`) DO (
   IF %%~zL NEQ 0 (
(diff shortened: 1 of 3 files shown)


Triang3l:
Which 3.7+ APIs does xenia-build use?

Margen67:
> Which 3.7+ APIs does xenia-build require?

None, [3.6 is just EOL](https://www.python.org/downloads/release/python-3615/). If we want to keep 3.6 support I can revert it.

Triang3l:
I think unless we already specifically require something added in 3.7 or in a newer version, we shouldn't be making commits only because the maintainers of a dependency have stated that some version is out of date while still technically sufficient for us — however, Python 3.6 being deprecated by its developers means that we can more safely ignore it if it becomes maintenance burden in the future, and just drop support if we need something not available in it, or, for instance, if our Python executable search becomes overcomplicated by it.

brichan:
Don't you think that EOL increases security risks?

Triang3l:
> Don't you think that EOL increases security risks?
We don't execute any scripts from untrusted/dynamic sources, nor do we work with things involving input from the internet aside from Git submodules, there's no way our build scripts can be exploited.

Miyabha:
Dear sir plz develop the Emulator for Android xbox classic

xb.bat
```diff
@@ -33,9 +35,12 @@ SETLOCAL ENABLEDELAYEDEXPANSION
 
 SET FOUND_PATH=""
 
-SET "CANDIDATE_PATHS[0]=C:\python37\python.exe"
-SET "CANDIDATE_PATHS[1]=C:\python36\python.exe"
-SET OUTPUT_INDEX=2
+SET "CANDIDATE_PATHS[0]=C:\python310\python.exe"
+SET "CANDIDATE_PATHS[1]=C:\python39\python.exe"
+SET "CANDIDATE_PATHS[2]=C:\python38\python.exe"
+SET "CANDIDATE_PATHS[2]=C:\python37\python.exe"
```
gibbed, line 41: There's two `[2]`.

gibbed (changes requested):

xb.bat
illusion0001:
```suggestion
SET "CANDIDATE_PATHS[2]=C:\python38\python.exe"
SET "CANDIDATE_PATHS[3]=C:\python37\python.exe"
SET "CANDIDATE_PATHS[4]=C:\python%PYTHON_MINIMUM_VERSION[0]%%PYTHON_MINIMUM_VERSION[1]%\python.exe"
```

brichan:
I am honestly disappointed that they are still using EOL.
The theory that if there is a security risk in EOL, it is not supported, but that is ok because we don't use it, is now a fatal security risk.

I don't understand why they can't consider the possibility that some intrusive code could search for and parasitize Python that has gone EOL.

And when that happens, the blame falls on the user, but in reality, it is the one who is condoning it.

How can developers who use Github and other internet repositories not think of that?


Translated with www.DeepL.com/Translator (free version)

Triang3l:
.bat files are even older and have way weirder and easier to break argument parsing, so should we consider them completely unsuitable for their primary purpose — local automation tasks — now? How is it suddenly a "fatal" risk in this context? While people still keep trying to use Xenia on Windows 7.

JoelLinn:
> I am honestly disappointed that they are still using EOL. 

Were exactly are _we_ using EOL Python?
 
 > but that is ok because we don't use it, is now a fatal security risk.
 
Should we blacklist msvcrt versions with known security issues?

> I don't understand why they can't consider the possibility that some intrusive code could search for and parasitize Python that has gone EOL.

We don’t consider because there is no Python _in_ this project to be found.

> And when that happens, the blame falls on the user, but in reality, it is the one who is condoning it.

I condone that someone is firing a rifle at a propane bottle that I sold them. I agree its a bad idea but I cant control how they use my propane.

> How can developers who use Github and other internet repositories not think of that?

I hope we do. But to be honest I cant see any real attack vector that we should fix. Maybe we should delete xenia because even loading a downloaded savegame is enough to get code execution I guess.

brichan:
Just to be clear, I am not talking about the Xenia execution environment.
We are talking about the Xenia build environment.
I think you know that by the purpose of this thread.

You are calling Python in xb.bat and xb.ps1.
(This is a side note, but even xb setup doesn't work with errors in the current mainstream Python 3.9.)
That's what I'm talking about there.

Your responsibility is based on Python 3.6, which is EOL and even the security fixes end at the end of 2021.
It is based on this thread's response to 3.7, which states that if you don't need the new features, you don't need to update.

It is mandatory to have Python 3.6 in the environment of Contributors and those who want to contribute to Xenia improvements, thereby posing a security risk.

Margen67:
> 

@brichan If `xb`/`xenia-build` is broken in newer versions of Python you should create an issue about that.

Triang3l:
It is not mandatory to have Python 3.6, any version starting with 3.6 can be used, just the scripts happen to work with 3.6 too.

gibbed (approved):

PR #2006, opened 2022-04-04
[GPU] Fixed size 0 point sprites enlarged to default

deaklajos:
Fable 2 writes 0.0f to the point size in some cases.
The code before turned it into a default sized sprite which resulted in huge snowflakes.


src/xenia/gpu/shaders/primitive_point_list.gs.hlsl
```diff
@@ -19,7 +19,7 @@ void main(point XeVertexPreGS xe_in[1],
   // Shader header writes -1.0f to point_size by default, so any positive value
   // means that it was overwritten by the translated vertex shader.
   float2 point_size =
-      xe_in[0].post_gs.pre_ps.point_params.z > 0.0f
+      xe_in[0].post_gs.pre_ps.point_params.z >= 0.0f
```
deaklajos, line 22: Not sure if point size 0.0f should be dropped or drawn with min size.
Makes no noticeable difference in Fable 2.

Triang3l:
Nice, thank you for the fix, and welcome to the project!!! I wonder if we might want to clamp the value to [0 on an actual write in vertex shaders — but first we need to investigate how special point size cases (negative, infinite, NaN), as well as conditional writing to the point size register, are handled on a real console (or an Adreno 200 phone via gl_PointSize). As well as whether the point size output in the microcode is the half-size or the whole size along one axis. To drop or not to drop — I'm not sure, it's an ALU (one more comparison in the general case) vs. export (for discarded snowflakes in Fable 2) balance question.

PR #2007, opened 2022-04-10
[GPU] Allowing patch draws with less vertices then required on the host

deaklajos:
4D5307F1 draws gound patches with one vertex and it seems like that is enough on the Xenos.

### DIFF
--- a/src/xenia/gpu/primitive_processor.cc
+++ b/src/xenia/gpu/primitive_processor.cc
@@ -823,6 +823,26 @@ bool PrimitiveProcessor::Process(ProcessingResult& result_out) {
     }
   }
 
+  // - 4D5307F1 - Continuous AutoIndexed QuadPatch draws initiated with only
+  // one vertex. host_draw_vertex_count needs to be rounded up to meet the
+  // requirements of the host hull shader.
+  if (tessellation_enabled &&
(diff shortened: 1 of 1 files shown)


src/xenia/gpu/primitive_processor.cc
```diff
@@ -823,6 +823,26 @@ bool PrimitiveProcessor::Process(ProcessingResult& result_out) {
     }
   }
 
+  // - 4D5307F1 - Continuous AutoIndexed QuadPatch draws initiated with only
+  // one vertex. host_draw_vertex_count needs to be rounded up to meet the
+  // requirements of the host hull shader.
+  if (tessellation_enabled &&
```
Triang3l, line 829: For indexed draws (the R400 can tessellate regular primitives in discrete and continuous modes, not only the specialized patch primitive types), this is dangerous as indices beyond `guest_index_buffer_needed_bytes` not covered by `shared_memory_.RequestRange` will be fetched by the draw. Though I'm not sure how it should work on the real hardware in this case.

Triang3l (changes requested):

Triang3l:
Implemented in a safer location in the function (together with moving the clamping that was already there), thank you!

deaklajos:
Thanks!

PR #2008, opened 2022-04-10
[GPU] Offset added to primitive index in tesselation domain shaders

deaklajos:
4D5307F1 ground shader expects patch indices to be offsetted.

### DIFF
--- a/src/xenia/gpu/dxbc_shader_translator.cc
+++ b/src/xenia/gpu/dxbc_shader_translator.cc
@@ -435,12 +435,19 @@ void DxbcShaderTranslator::StartVertexOrDomainShader() {
                                                   : dxbc::Dest::R(0, 0b0111),
                  dxbc::Src::VDomain(0b000110));
         if (register_count() >= 2) {
-          // Copy the primitive index to r1.x as a float.
+          // Copy the primitive index + offset to r1.x as a float.
           uint32_t primitive_id_temp =
               uses_register_dynamic_addressing ? PushSystemTemp() : 1;
           in_primitive_id_used_ = true;
+          a_.OpMov(dxbc::Dest::R(primitive_id_temp, 0b0001),
+                   dxbc::Src::VPrim());
+          a_.OpIAdd(dxbc::Dest::R(primitive_id_temp, 0b0001),
+                    dxbc::Src::R(primitive_id_temp, dxbc::Src::kXXXX),
+                    dxbc::Src::CB(cbuffer_index_system_constants_,
+                                  uint32_t(CbufferRegister::kSystemConstants),
+                                  1, dxbc::Src::kYYYY));
           a_.OpUToF(dxbc::Dest::R(primitive_id_temp, 0b0001),
-                    dxbc::Src::VPrim());
+                    dxbc::Src::R(primitive_id_temp, dxbc::Src::kXXXX));
           if (uses_register_dynamic_addressing) {
             a_.OpMov(dxbc::Dest::X(0, 1, 0b0001),
                      dxbc::Src::R(primitive_id_temp, dxbc::Src::kXXXX));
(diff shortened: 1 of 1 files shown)


src/xenia/gpu/dxbc_shader_translator.cc
```diff
@@ -518,11 +525,18 @@ void DxbcShaderTranslator::StartVertexOrDomainShader() {
...
-        // Copy the primitive index to r0.x as a float.
+        // Copy the primitive index + offset to r0.x as a float.
         uint32_t primitive_id_temp =
             uses_register_dynamic_addressing ? PushSystemTemp() : 0;
         in_primitive_id_used_ = true;
-        a_.OpUToF(dxbc::Dest::R(primitive_id_temp, 0b0001), dxbc::Src::VPrim());
+        a_.OpMov(dxbc::Dest::R(primitive_id_temp, 0b0001), dxbc::Src::VPrim());
+        a_.OpIAdd(dxbc::Dest::R(primitive_id_temp, 0b0001),
+                  dxbc::Src::R(primitive_id_temp, dxbc::Src::kXXXX),
+                  dxbc::Src::CB(cbuffer_index_system_constants_,
```
Triang3l, line 535: System constants should be loaded via `LoadSystemConstant` (if I recall correctly), so it is ensured that the system constant buffer binding exists, and that the needed constant is marked as used in the resource definitions.

src/xenia/gpu/dxbc_shader_translator.cc
```diff
@@ -435,12 +435,19 @@ void DxbcShaderTranslator::StartVertexOrDomainShader() {
                                                   : dxbc::Dest::R(0, 0b0111),
                  dxbc::Src::VDomain(0b000110));
         if (register_count() >= 2) {
-          // Copy the primitive index to r1.x as a float.
+          // Copy the primitive index + offset to r1.x as a float.
           uint32_t primitive_id_temp =
               uses_register_dynamic_addressing ? PushSystemTemp() : 1;
           in_primitive_id_used_ = true;
+          a_.OpMov(dxbc::Dest::R(primitive_id_temp, 0b0001),
```
Triang3l, line 442: Shader inputs can be used directly as source operands, there's no need to `mov` it.

src/xenia/gpu/dxbc_shader_translator.cc
```diff
@@ -435,12 +435,19 @@ void DxbcShaderTranslator::StartVertexOrDomainShader() {
                                                   : dxbc::Dest::R(0, 0b0111),
                  dxbc::Src::VDomain(0b000110));
         if (register_count() >= 2) {
-          // Copy the primitive index to r1.x as a float.
+          // Copy the primitive index + offset to r1.x as a float.
```
Triang3l, line 438: I can't remember if we already have a comment in the API-agnostic part of `xe::gpu`, but if yes, the need for the offset (referencing this title ID specifically and that it's used in ground draws there) should be documented there.

src/xenia/gpu/dxbc_shader_translator.cc
```diff
@@ -435,12 +435,19 @@ void DxbcShaderTranslator::StartVertexOrDomainShader() {
...
           uint32_t primitive_id_temp =
               uses_register_dynamic_addressing ? PushSystemTemp() : 1;
           in_primitive_id_used_ = true;
+          a_.OpMov(dxbc::Dest::R(primitive_id_temp, 0b0001),
+                   dxbc::Src::VPrim());
+          a_.OpIAdd(dxbc::Dest::R(primitive_id_temp, 0b0001),
```
Triang3l, line 444: Since this index appears to be treated like the vertex index, I think vPrim + offset should also be clamped to the min and the max indices from the registers, just like in vertex index loading. The clamping also restricts the index to the range with integers exactly representable as float32 (24 bits).

src/xenia/gpu/dxbc_shader_translator.cc
```diff
@@ -435,12 +435,19 @@ void DxbcShaderTranslator::StartVertexOrDomainShader() {
...
+                    dxbc::Src::R(primitive_id_temp, dxbc::Src::kXXXX),
+                    dxbc::Src::CB(cbuffer_index_system_constants_,
+                                  uint32_t(CbufferRegister::kSystemConstants),
+                                  1, dxbc::Src::kYYYY));
           a_.OpUToF(dxbc::Dest::R(primitive_id_temp, 0b0001),
```
Triang3l, line 449: As there are multiple instances of this code, and with clamping it will be even longer, it can be moved to `void LoadTessellationPatchIndex(uint32_t dest_temp, uint32_t dest_temp_component)` (see other functions accepting temporary register index/component in `DxbcShaderTranslator` for reference, such as 20e4/7e3 float conversion).

Or better, even to a function that would also be used for the vertex index after byte swapping (accepting the source as the third `const dxbc::Src& src` argument, which can safely be the same as the destination as it will be referenced only once, in the first operation), like `ConvertVertexIndex`, doing offsetting, clamping, and float32 conversion.

Triang3l (changes requested):

Triang3l:
I've applied the requested changes, if you have time, please verify that after the clamping the indices are still the ones that are needed (though I've never seen the min/max being different than 0 and 0xFFFFFF). Thanks for the finding and the initial implementation!

deaklajos:
Thanks!

PR #2016, opened 2022-05-08
Draft: some fixes

tobiasjakobi:
Hello,

I have tried to build xenia on Linux (Gentoo x86-64) for the first time, and noticed some build/linker/runtime errors.

I've flagged the assertion commit as WIP, as I'm unsure if removing the assert is the best solution here, or if the condition can be somehow relaxed.

Please let me know what you think.

With best wishes,
Tobias

### DIFF
--- a/src/xenia/emulator.cc
+++ b/src/xenia/emulator.cc
@@ -721,7 +721,7 @@ X_STATUS Emulator::CompleteLaunch(const std::filesystem::path& path,
                                   const std::string_view module_path) {
   // Making changes to the UI (setting the icon) and executing game config load
   // callbacks which expect to be called from the UI thread.
-  assert_true(display_window_->app_context().IsInUIThread());
+  //assert_true(display_window_->app_context().IsInUIThread());
 
   // Setup NullDevices for raw HDD partition accesses
   // Cache/STFC code baked into games tries reading/writing to these
--- a/src/xenia/gpu/trace_viewer.cc
+++ b/src/xenia/gpu/trace_viewer.cc
@@ -408,6 +408,12 @@ void TraceViewer::DrawPacketDisassemblerUI() {
         }
         break;
       }
+      case TraceCommandType::kRegisters: {
+        break;
+      }
+      case TraceCommandType::kGammaRamp: {
+        break;
+      }
     }
   }
   ImGui::EndChild();
--- a/src/xenia/gpu/register_file.h
+++ b/src/xenia/gpu/register_file.h
@@ -33,7 +33,7 @@ class RegisterFile {
 
   static const RegisterInfo* GetRegisterInfo(uint32_t index);
 
-  static const size_t kRegisterCount = 0x5003;
+  static constexpr size_t kRegisterCount = 0x5003;
   union RegisterValue {
     uint32_t u32;
     float f32;


src/xenia/emulator.cc
```diff
@@ -721,7 +721,7 @@ X_STATUS Emulator::CompleteLaunch(const std::filesystem::path& path,
                                   const std::string_view module_path) {
   // Making changes to the UI (setting the icon) and executing game config load
   // callbacks which expect to be called from the UI thread.
-  assert_true(display_window_->app_context().IsInUIThread());
```
Triang3l, line 724: This is important as CompleteLaunch does _a lot_ of UI interaction — setting the title (directly, as well as via the shader storage initialization feedback), the icon, and also loads the game config (which is already a mess from the threading point of view, but the config is assumed to be loaded and stored only in the UI thread by the recently added post-processing settings at least).

Triang3l (changes requested):

src/xenia/gpu/trace_viewer.cc
```diff
@@ -408,6 +408,12 @@ void TraceViewer::DrawPacketDisassemblerUI() {
         }
         break;
       }
+      case TraceCommandType::kRegisters: {
```
Triang3l, line 411: Oops, didn't notice them, will fix by myself possibly, thank you! These are used for internal purposes (not a part of guest packets — so no need to display them visually), and need to be handled like kMemoryWrite or kEdramSnapshot.

Triang3l (changes requested):

src/xenia/emulator.cc
tobiasjakobi: So any idea why this triggers in my case? Would a backtrace help here?

Triang3l: Yes — how specifically are you launching the game?

tobiasjakobi: Nothing special, just launching the binary from the build dir:
```
liquid@leena ~/sourcecode/xenia/build/bin/Linux/Debug $ SDL_VIDEODRIVER=x11 ./xenia /tmp/test.iso
```

I don't think setting the SDL envvar is really necessary here (as Xenia already does some setenv magic), but well, it certainly doesn't hurt.

Triang3l: If we allow running this in non-UI threads, UI interaction must be done via `CallInUIThreadSynchronous` of the `app_context()`.

tobiasjakobi: Here's the backtrace when hitting the assertion.
[xenia_gdb_2.txt](https://github.com/xenia-project/xenia/files/8647441/xenia_gdb_2.txt)

Let me know if I can provide more information.

gibbed: This also happens on Windows if you launch with a target on the commandline (or indirectly via a custom config with `General.target` cvar set).

tobiasjakobi: Makes sense, as the callstack indicates that the call originates from `EmulatorApp::EmulatorThread()`. I guess there's a missing `app_context().CallInUIThread()` here?

Triang3l: `CallInUIThread` is fire and forget, by the way — need to be careful either not to have references in the called function to anything that may be destroyed before/during the invocation, or to force flush the pending function queue (with an empty `CallInUIThreadSynchronous` if the initial request was made via `CallInUIThread` — or `CallInUIThreadSynchronous` doing `ExecutePendingFunctionsFromUIThread` if the request was done using `CallInUIThreadDeferred`) before destroying the dependencies of the called function.

tobiasjakobi: I tried using `CallInUIThreadSynchronous` but I run into a deadlock in the error path:
```
#0  0x00007ffff68f1b65 in pthread_cond_destroy () at /lib64/libc.so.6
#1  0x00007ffff68aa1d8 in  () at /lib64/libc.so.6
#2  0x00007ffff68aa2ea in  () at /lib64/libc.so.6
#3  0x0000000000590572 in xe::FatalError(std::basic_string_view<char, std::char_traits<char> >) (str="Unable to mount disc image; file not found or corrupt.")
```

I'm giving up here and closing this PR. Fixing this properly seems to require detailed knowledge about how threading works here. Something which I currently lack.

PR #2018, opened 2022-05-13
Support for SysV ABI in x64 emitter (and other fixes)

amessier:
This PR adds support for SysV ABI (which is used by Linux) to the x64 backend emitter. See #1339 and #803 for previous PRs related to this PR. While I reused some commits from those PRs, it was mainly for some of the trivial fixes. For the SysV ABI support itself, since there was issues in Release mode (as seen in #1339), I started from a clean page.

I decided to use a function that returns the register for a given index. That function is implemented differently per ABI (Windows or SysV) at compile time using defines.

Commit 2b9764dca90b561e3a7b1f6f31aca9b37fddc5e8 is similar to a change made in #1339 (bb9820add08caac145b9710ab153739d8f5b1248). Basically, for m128 values, clang would pass them by value in xmm registers, but the thunk expects arguments passed in general purpose registers (GPRs) only. By changing the functions to pass by reference, it forces clang to pass them in the GPRs.

As for the fixes in Release mode, there are two commits.
-  First, fe2b7197d9914269f306d4f9cee95a2f6f36316b fixes stack corruption. I'm pretty sure the fix is valid, but would certainly require review from others.
- As for a14ed35a6b1edd505e89a1bf17421689402fc2bf this was required with GCC to ensure proper initialization of sequence_table before being used by the other files. I went with the method of including the other files in the main module, but I'm definitively open to other suggestions.

Finally the last commit in this PR (7a7d17e540aa6d62b376d7d6b2336aea95fe6964) is a fix for an issue in the filesystem code, seen when trying to start a .xex (now that we can get to that point).

With these changes, xenia-cpu-ppc-tests passes on Linux. Xenia also starts and you can try to start a .xex.

I tested those changes on Windows too by running xenia-cpu-ppc-tests only (can't start a game on that machine).


gibbed (changes requested):
Please bump the copyright year in all of the files you've touched.

amessier:
> Please bump the copyright year in all of the files you've touched.

I've edited the commits to update the copyright year where applicable.

I also forgot to mention in the original PR message that to support the different ABI, I chose to reduce the number of GPR available to the emitter, since in SysV there are fewer preserved registers available. I assume the emitter can handle the reduced number of registers. There is probably a better way to do this (one that has the same number of registers as on Windows) but for now this seems to work.

PR #2024, opened 2022-05-28
Add cvars for video modes

halotroop2288:
This adds config variables for Interlaced, Hi-Def, and Widescreen video modes.

This removes the need for custom builds to test games in standard definition, Fullscreen video modes.


Triang3l:
Are all combinations of the values of these cvars possible on a real Xbox 360 combined with all the other values we currently write into that structure?

halotroop2288:
> Are all combinations of the values of these cvars possible on a real Xbox 360...?

In theory this depends on the display attached, so yes. You could hack a display to tell the 360 literally any combination of specs.
They might not be supported, but they may have interesting results. I just left the defaults as they were originally so it shouldn't cause any issues for a novice user.

PR #2029, opened 2022-06-06
[Docs] Improvements

Margen67:
See commits.


docs/building.md
```diff
@@ -8,68 +8,69 @@ drivers.
...
 * [Python 3.6+](https://www.python.org/downloads/)
   * Ensure Python is in PATH.
-* Windows 11 SDK version 10.0.22000.0 (for Visual Studio 2019, this or any newer version)
-
+  * Also see https://learn.microsoft.com/en-us/windows/python/faqs#why-does-running-python-exe-open-the-microsoft-store-
 ```
 git clone https://github.com/xenia-project/xenia.git
 cd xenia
-xb setup
+.\xb setup
```
Triang3l, line 21: Backslashes are Windows-specific, I'd recommend forward slashes, which are accepted by both Linux and Windows (also considering that they're even more important on Linux than on Windows, possibly only for PowerShell on the latter, as on Linux, not providing an explicit executable path results in the executable being searched for only in the system directories or something like that).

Triang3l (changes requested):

docs/building.md
Margen67: Forward slashes break cmd/Command Prompt. Should I just leave them out instead?

Triang3l: They work in the Command Prompt as normal, possibly with the exception of UNC paths, although I'm not sure if this case applies to relative paths at all.

illusion0001: That's not how you use it

PR #2046, opened 2022-08-11
[Emulator] Added option for content installation

Gliniak:
image

This allows less experienced users to have less issues with getting savefiles from console to Xenia and significantly simplifies process of installing DLCs.

Process right now:
1. Go to content directory and find/create title id of game.
2. Open/Create content type directory.
3. Create directory for savefile/DLCs. (Which for savefile usually must be very specific to even load)
4. Extract STFS package.
5. Put extracted files into content.

Process with install content:
1. Open Xenia.
2. Select "Install Content" and select content package that you want to install.
3. Profit.


Razzile:
This is a really good idea!

Blackbird88:
I don't know if this is possible to do, but it'd be nice if the title bar showed (Installing...) like compiling shaders does
image
Because if you install large DLC, Xenia just "freezes" while it's extracting it and it looks like crash in disguise.

src/xenia/vfs/virtual_file_system.cc
```diff
@@ -307,5 +310,74 @@ X_STATUS VirtualFileSystem::OpenFile(Entry* root_entry,
...
+  std::queue<vfs::Entry*> queue;
+  auto root = device->ResolvePath("/");
+  queue.push(root);
+
+  // Allocate a buffer when needed.
+  size_t buffer_size = 0;
+  uint8_t* buffer = nullptr;
+
+  while (!queue.empty()) {
+    auto entry = queue.front();
```
halotroop2288, line 325: I can't find the progress bar code anywhere (wierd), but you could probably just change the title bar saying the percentage of the queue that's completed at the beginning of this loop.

halotroop2288 (changes requested):

Gliniak:
I have two ideas. First is simple and easy, second is a bit more tricky.

1. Just create box that informs that content was installed correctly (or not)
image

2. Create progressBar, but I'm struggling with implementation and how it should be updated. (Adding dummy progress bar is easy, but functional one is a bit tricky)
image

Darkhost1999:
Progress bars are cool and all yet I'm fine with just a window saying "Processing your request..." "Loading..." or "Installing..." (these all will have cancel buttons) when finished will say "Installation successful" or "Installation failed" and cancel changes to ok. Or we can leave cancel there and say "Are you sure you'd like to remove your installed content? Yes or no" when pressed.

Issue #1852, opened 2021-06-26
[Code style] Inconsistent kernel notification ID usage


Triang3l:
In Xenia, there are two kinds of declarations of kernel notification IDs:

- System notifications IDs are used simply as hard-written values, without any enum/consts/#defines, with merely comments indicating their original XN_SYS name.
- XAM apps (such as XmpApp) have notification IDs declared in the classes, as `private static const uint32_t kMsg…`.

For future convenience especially when we start running the original app executables, we need to clean this up, probably as some kind of an `enum` with nice `k`-prefixed names.

We can use two approaches:

- An `enum class XNotificationID` containing both system and app notification IDs.
- A `typedef uint32_t XNotificationID`, and separate enums for system notifications (declared globally) and for each app (in their headers).

In my opinion, the former is preferable, as if we start running the original XAM apps, it's possible that we won't have our own classes for a subset of them. As each apparently has its own index in the upper bits, there should likely be no collisions, and there's natural namespacing. Also, if I understand correctly, notifications are used for inter-app communication, so conceptually they shouldn't be `private`. Plus, type safety is somewhat better with an `enum class`.

kitsiosvas:
Hi, I'm interested on working on this issue. 
If that's possible, I would greatly appreciate any tips on where to start looking for implementing such things. 
If I understand it correctly, the goal is to use a `enum` class that holds possible IDs and each notification's ID will be a object from this class?

PR #2047, opened 2022-08-17
[Code style] Changed hard-written system notification ID's to enum cl…

zaid-ismail031:
See issue #1852. Cleaned up the code by changing hard-written system notification ID's (and their corresponding definition `typedef uint32_t XNotificationID`). Replaced these with `enum class XNotificationID` (in xbox.h) and replaced all hard-written values with `XNotifcationID::*Notification ID Name*`.


src/xenia/xbox.h
```diff
@@ -382,6 +381,13 @@ enum class XContentType : uint32_t {
   kCommunityGame = 0x02000000,
 };
 
+enum class XNotificationID : uint32_t {
+  kXn_sys_ui = 0x00000009,
```
Triang3l, line 385: `Xn_` is not needed since it's a `enum class`, but if it was just an `enum` the prefix would be `kXNotificationID`. Also, `_` is rarely used in Xenia in `kCamelCase` cases (mostly when there are numbers to separate), and also the case is random in this commit. Also, `System` doesn't need to be shortened.

src/xenia/xbox.h
```diff
@@ -382,6 +381,13 @@ enum class XContentType : uint32_t {
   kCommunityGame = 0x02000000,
 };
 
+enum class XNotificationID : uint32_t {
```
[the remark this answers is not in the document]
zaid-ismail031: I've also realized that I forgot about the XAM Apps. 

From Issue #1852:
"XAM apps (such as XmpApp) have notification IDs declared in their classes, as `private static const uint32_t kMsg….`"

and

"Also, if I understand correctly, notifications are used for inter-app communication, so conceptually they shouldn't be `private`"

Should I change the XAM Apps' notification ID's to enums too?

src/xenia/xbox.h
zaid-ismail031: I've renamed the enum variables as requested. For example, changed `kXn_sys_ui` to `kSystemUI`.

src/xenia/xbox.h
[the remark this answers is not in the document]
zaid-ismail031: I implemented the changes from what I understood. I added back the `typedef uint32_t XNotificationID` to xbox.h. I declared the system notifications as `enum : XNotificationID {...` in kernel_state.h. I changed the app notifications in xmpapp.h from `private static const uint32_t` to an `enum : XNotificationID {...`.

illusion0001:
Have you tested these changes? https://ci.appveyor.com/project/benvanik/xenia/builds/44501665/job/doq1w8c0dx4sibad#L996

zaid-ismail031:
>Have you tested these changes? https://ci.appveyor.com/project/benvanik/xenia/builds/44501665/job/doq1w8c0dx4sibad#L996

I missed the issue with `xam_ui.cc`. I have fixed it now, by changing all instances of `XNotificationID::kXn_sys_ui` to `kSystemUI`. Tests seem to be fine now.

Razzile:
I'm always here for magic numbers being replaced by enums! Can you squash this PR into one commit though if you're done?

src/xenia/kernel/kernel_state.h
```diff
@@ -236,6 +236,13 @@ class KernelState {
   friend class XObject;
 };
 
+enum : XNotificationID {
+  kSystemUI = 0x00000009,
```
Triang3l, line 240: For `enum` rather than `enum class` (that have their values leaked into the scope of where they're defined), the values themselves need to be prefixed like `kXNotificationIDSystemUI`, to state clearly what they are in their usage locations, and to avoid potential collisions with unrelated values.

src/xenia/kernel/xam/apps/xmp_app.h
```diff
@@ -113,6 +109,12 @@ class XmpApp : public App {
   uint32_t next_song_handle_;
 };
 
+enum : XNotificationID {
+  kMsgStateChanged = 0x0A000001,
```
Triang3l, line 113: This should be kept in `XmpApp` (as these are messages specific to `XmpApp`, not global messages for all apps — while in the commit it's just `apps::kMsgStateChanged` without clear association to a specific app), though maybe in the `public` part even since they may be used to send messages to `XmpApp` from other places, not just for receiving them.

Triang3l (changes requested):

src/xenia/kernel/xam/apps/xmp_app.h
zaid-ismail031: Okay. I've moved the enum to the `public` part of the `XmpApp` class,

src/xenia/kernel/kernel_state.h
zaid-ismail031: Okay. I have changed the `enum` values accordingly.

Triang3l (changes requested):

zaid-ismail031:
Well, I screwed something up with this PR. Been working with Plastic SCM for the past year and forgot how GitHub works.

zaid-ismail031:
I kinda screwed up the PR a little with a bad force push

zaid-ismail031:
Okay, I have fixed it.

PR #2058, opened 2022-09-09
[x64] Add AVX512 optimization for `OPCODE_VECTOR_CONVERT_F2I`(unsigned)

Wunkolo:
[vcvttps2udq](https://www.laruence.com/x86/VCVTTPS2UDQ.html) already saturates overflowing and unordered values to `0xFFFFFFFF`. Using mask registers, zeroes are written to negative values within the same instruction.

The entire operation is just two instructions!

### DIFF
--- a/src/xenia/cpu/backend/x64/x64_seq_vector.cc
+++ b/src/xenia/cpu/backend/x64/x64_seq_vector.cc
@@ -83,6 +83,19 @@ struct VECTOR_CONVERT_F2I
                I<OPCODE_VECTOR_CONVERT_F2I, V128Op, V128Op>> {
   static void Emit(X64Emitter& e, const EmitArgType& i) {
     if (i.instr->flags & ARITHMETIC_UNSIGNED) {
+      if (e.IsFeatureEnabled(kX64EmitAVX512Ortho)) {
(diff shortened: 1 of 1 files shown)


src/xenia/cpu/backend/x64/x64_seq_vector.cc
```diff
@@ -83,6 +83,19 @@ struct VECTOR_CONVERT_F2I
...
     if (i.instr->flags & ARITHMETIC_UNSIGNED) {
+      if (e.IsFeatureEnabled(kX64EmitAVX512Ortho)) {
+        Opmask mask = e.k1;
+        // Mask positive values and unordered values
+        // _CMP_NLT_UQ
+        e.vcmpps(mask, i.src1, e.GetXmmConstPtr(XMMZero), 0x15);
+
+        // vcvttps2udq will saturate overflowing positive values and unordered
+        // values to UINT_MAX. Mask registers will write zero everywhere
+        // else(negative values)
```
gibbed, line 94: Nitpicking: space before the opening parenthesis.

gibbed (changes requested):

gibbed (approved):

PR #2059, opened 2022-09-09
[x64] Add AVX512 optimization for `OPCODE_VECTOR_ADD`(saturated)

Wunkolo:
Uses a single `vpternlogd` to test for signed/unsigned overflow/underflow. Then utilizes AVX512 mask operations to create either `0x7FFFFFFF` or `0x80000000` arithmetically.

### DIFF
--- a/src/xenia/cpu/backend/x64/x64_seq_vector.cc
+++ b/src/xenia/cpu/backend/x64/x64_seq_vector.cc
@@ -547,6 +547,15 @@ struct VECTOR_ADD
             case INT32_TYPE:
               if (saturate) {
                 if (is_unsigned) {
+                  if (e.IsFeatureEnabled(kX64EmitAVX512Ortho)) {
+                    e.vpaddd(dest, src1, src2);
+                    Opmask saturate = e.k1;
+                    // _mm_cmplt_epu32_mask
+                    e.vpcmpud(saturate, dest, src1, 0x1);
+                    e.vpternlogd(dest | saturate, dest, dest, 0xFF);
+                    return;
+                  }
+
                   // xmm0 is the only temp register that can be used by
                   // src1/src2.
                   e.vpaddd(e.xmm1, src1, src2);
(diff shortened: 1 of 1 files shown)


src/xenia/cpu/backend/x64/x64_seq_vector.cc
```diff
@@ -562,6 +571,20 @@ struct VECTOR_ADD
...
+                                         kX64EmitAVX512DQ)) {
+                    e.vmovdqa32(e.xmm3, src1);
+                    e.vpternlogd(e.xmm3, e.xmm1, src2, 0b00100100);
+                    Opmask saturate = e.k1;
+                    e.vpmovd2m(saturate, e.xmm3);
+
+                    e.vpsrad(e.xmm1 | saturate, e.xmm1, 31);
+                    e.vpxord(e.xmm1 | saturate, e.xmm1,
+                             e.GetXmmConstPtr(XMMSignMaskI32));
+                    e.vmovdqa32(dest, e.xmm1);
```
Wunkolo, line 584: This works fine. But it's not ideal. For some reason I have to operate upon `e.xmm1` and then move it into `dest` at the end with `vmovdqa32` when I could have just operated upon `dest` directly. But it seems to fail unit tests if I do that. This is still faster than the non-AVX512 route but needs some deeper investigation since it could save an additional instruction.

gibbed: This would probably do better as a TODO comment next to this code instead of as a comment in the PR?

Wunkolo: Currently going to look into it a bit longer before undrafting this PR.

Wunkolo: Good to go now. Turns out, in some cases, `src1` or `src2` can be the same as `dest`. Causing some overwriting to occur if writing to `dest` too early and then reading from `src1` or `src2` again.

gibbed (approved):

PR #2060, opened 2022-09-11
[CPU] Don't apply invalid patches

Margen67:


illusion0001:
Try defining your cvar here https://github.com/xenia-project/xenia/blob/90fffe1de7ca2f6ee404d9fe3181b02a6e316ac8/src/xenia/cpu/xex_module.cc#L58-L60

PR #2064, opened 2022-09-20
[App/UI/Kernel] Add console region setting and locale menu

beeanyew:
This adds a console region setting to the XConfig section of the Xenia config file and changes the behavior of `xeXGetGameRegion`, allowing for things like region-specific startup screens and language options to be available in games, such as the RIOT ACT title screen in Crackdown (`4D5307DC`) and the Piracy Warning and Japanese-specific title screen in Sonic Generations (`53450848`). It also enables use of the Japanese dub in Ace Combat 6 (not confirmed by me) and while Xenia master cannot currently load them, it also enables correct function of the language packs for Resident Evil 6.

Also adds a menu to the Xenia window for configuring the user language, user country and console region settings without having to manually edit the config file and typing in values. The menu checkmark code is a bit... strange, but the comment in `window_win.cc` explains why it currently looks the way it does. (This menu is disabled while a title is running.)

While Crackdown currently crashes on Xenia master due to an incorrect return code when loading a saved game, here are some example screenshots:
image

### DIFF
--- a/src/xenia/app/emulator_window.cc
+++ b/src/xenia/app/emulator_window.cc
@@ -31,6 +31,7 @@
 #include "xenia/emulator.h"
 #include "xenia/gpu/command_processor.h"
 #include "xenia/gpu/graphics_system.h"
+#include "xenia/kernel/xboxkrnl/xboxkrnl_xconfig.h"
 #include "xenia/ui/file_picker.h"
 #include "xenia/ui/graphics_provider.h"
 #include "xenia/ui/imgui_dialog.h"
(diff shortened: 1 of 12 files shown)


gibbed:
The menu selection for those settings is not a good idea unless you can confirm the Xbox 360 can change those settings while the game is running.

beeanyew:
Ah no, it technically can not change those settings while a game is running, since that requires launching the settings application part of the dashboard. But you're saying that the menu can't just be disabled while a title is running?

gibbed:
> Ah no, it technically can not change those settings while a game is running, since that requires launching the settings application part of the dashboard. But you're saying that the menu can't just be disabled while a title is running?
> 
> Edit: Sorry, I realize now that I forgot to copy the latest `emulator_window.cc` to the PR directory before committing, there's also a typo left in `xam_info.cc` that I fixed but also forgot to copy. I'll get those in there right away.

No worries, if you disable changing the settings via the menu when a game is running that should be fine.

beeanyew:
I realized that there was another `LaunchPath` call hiding out in `xenia_main.cc`, so I felt the need to rework disabling the console locale menu a little bit.

src/xenia/app/emulator_window.h
```diff
@@ -25,6 +25,15 @@
 #include "xenia/ui/windowed_app_context.h"
 #include "xenia/xbox.h"
 
+enum menuIndex {
```
Razzile, line 28: this should be made an `enum class` instead to match with our code style.

something like
```cpp
enum class MenuIndex {
  kFile,
  kCpu
  ...
```

src/xenia/kernel/xboxkrnl/xboxkrnl_xconfig.h
```diff
@@ -0,0 +1,40 @@
...
+ ******************************************************************************
+ */
+
+#ifndef XENIA_KERNEL_XBOXKRNL_XBOXKRNL_XCONFIG_H_
+#define XENIA_KERNEL_XBOXKRNL_XBOXKRNL_XCONFIG_H_
+
+#include "xenia/kernel/util/shim_utils.h"
+#include "xenia/xbox.h"
+
+enum xcConsoleRegion {
```
Razzile, line 16: Same as above, enum class is preferred (and PascalCase beginning with k for the enumerations "kJapan")

src/xenia/kernel/xboxkrnl/xboxkrnl_xconfig.h
```diff
@@ -0,0 +1,40 @@
...
+  xcConsoleRegion_US,
+  xcConsoleRegion_Japan,
+  xcConsoleRegion_Asia,
+  xcConsoleRegion_Europe,
+  xcConsoleRegion_RegionFree,
+};
+
+enum xcRegionSettingType {
```
Razzile, line 24: Same as above, enum class is preferred (and PascalCase beginning with k for the enumerations "kUserLanguage")

src/xenia/cpu/backend/x64/x64_seq_vector.cc
```diff
@@ -594,7 +594,7 @@ struct VECTOR_ADD
 
                     e.vpsrad(e.xmm2, e.xmm1, 31);
                     e.vpxord(e.xmm2, e.xmm2, e.GetXmmConstPtr(XMMSignMaskI32));
-                    e.vpblendmd(dest | saturate, e.xmm1, e.xmm2);
```
Razzile, line 597: I know this change was most likely done by the formatter but it does pollute the commit so probably worth reverting (@gibbed can decide for sure)

src/xenia/app/emulator_window.cc
```diff
@@ -937,6 +1055,37 @@ void EmulatorWindow::ToggleDisplayConfigDialog() {
...
+      }
+      break;
+    }
+    case xcRegionSetting_ConsoleRegion:
+      for (int i = 0; i < xe::countof(ui_console_region_codes); i++) {
+        cur_menu->SetChecked(i, (value == ui_console_region_codes[i]));
+      }
+      break;
+  }
+  xe::kernel::xboxkrnl::xcSaveRegionSetting((xcRegionSettingType)setting,
```
Razzile, line 1085: prefer static_cast to C-style casts

Razzile (changes requested):
Looks good, I would change your enum additions to `enum class` and match our naming conventions for enums

src/xenia/cpu/backend/x64/x64_seq_vector.cc
beeanyew: Ah yeah, no problem to revert it. The CI linting check was failing because of it, so I figured I'd leave it in at the time. Working on the coding style changes, will push a commit soon.
(I've mostly done C code in the past and I'm not super familiar with C++ language features, so all feedback and code review is very much appreciated.)

beeanyew: I removed this commit for now, so the CI linting will probably fail, but I can add it back if desired.

Razzile (approved):
LGTM. I saw Rick was discussing whether this should be a config file setting only but I have no bearing on that discussion

beeanyew:
I had actually forgotten to copy files again after editing them... I really should stop doing my coding out of the current git tree, it's just that it's such a giant pain to have GitAhead sit around and eat 100% on one thread every single time a file is modified in the tree, even if it's excluded by .gitignore. :D
Really sorry about that, the last commit is now actually using the correct and up to date files. It's just a minor change and a typo in a log message, but still...

src/xenia/kernel/xam/xam_info.cc
```diff
@@ -200,8 +202,41 @@ dword_result_t XGetAVPack_entry() {
...
+//          Probably not very useful for now? Not added.
+uint32_t xeXGetGameRegion() {
+  if (cvars::console_region == "us") {
+    return 0x00FFu;
+  } else if (cvars::console_region == "jp") {
+    return 0x0101u;
+  } else if (cvars::console_region == "asia") {
+    return 0x01FFu;
+  } else if (cvars::console_region == "eu") {
+    return 0x02FFu;
```
NicknineTheEagle, line 231: These appear to correlate with xex region flags (listed in xex2_region_flags). So it likely goes like this:
0x00FF - North America
0x0101 - Japan
0x0102 - China
0x01FF - Asia
0x0201 - Australia
0x02FF - Europe

beeanyew: Hi, thanks for this comment, but how did you come to that conclusion? I remember looking at the enums in `xex2_info.h` and not finding them particularly useful as to what value this should return, as it looks like this, which seems to be a 32-bit mask of some sort that I don't know the origin of:
```
enum xex2_region_flags : uint32_t {
  XEX_REGION_NTSCU = 0x000000FF,
  XEX_REGION_NTSCJ = 0x0000FF00,
  XEX_REGION_NTSCJ_JAPAN = 0x00000100,
  XEX_REGION_NTSCJ_CHINA = 0x00000200,
  XEX_REGION_PAL = 0x00FF0000,
  XEX_REGION_PAL_AU_NZ = 0x00010000,
  XEX_REGION_OTHER = 0xFF000000,
  XEX_REGION_ALL = 0xFFFFFFFF,
};
```
Like the comment in my code says, my values are based entirely on experimentation and in one (unused) case what DashLaunch displayed on screen for one value (`0x07FF`) when I typed it in on my RGH Xbox for the "Region spoofing" setting, but many of these values appeared to be wrong and/or empty.

I can of course add Australia as a preset value if you've checked this, and if it results in different game operation compared to just selecting "Europe". (That was the only reason I added Japan as a separate entry from what appeared to work as a generic NTSC-J "Asia" setting with any other `0x02XX` value.

NicknineTheEagle: It lines up when you look at the same side-by-side - look at it this way, in xex2_region_flags, NTSC-U is byte 0, NTSC-J is byte 1 and PAL is byte 2 and the actual value within the byte is sub-region.

beeanyew: I really don't know what you're talking about, sorry. The value returned from `xeXGetGameRegion` is definitely a 16-bit value, there's no third or fourth byte.
To be clear: I'm sure this is accurate for *XEX files*, but I still don't see any connection to the region value returned by the console and the region value set in the XEX file, and I'm pretty sure that *no* value from the XEX file should be used when checking for game region using this function.

NicknineTheEagle: Yeah, you misunderstood me again. How about this, in xex2_region_flags, count octets from right to left. The first one is NTSC-U, the second one is NTSC-J, the third one is PAL. This correlates to values returned by xeXGetGameRegion where, in the bottom 8 bits, 0 - is NTSC-U, 1 is NTSC-J and 2 is PAL.

beeanyew: Ah, sorry. I have no idea how octets work. It's not that I misunderstand, it's simply that I do not understand, like I said.
But I mean, it seems like you have a proper solution for this, and if you want to rework the code to use the proper math for it instead, you are of course free to do so, I have no idea how to do it.

NicknineTheEagle: Octets is a group of 8 bits.
Take 0x12345678 number, for example. Since bits in a number are counted from right to left, 78 is the first octet, 56 is the second octet, 34 is the third octet, 12 is the fourth octet.
Do you see it now?

beeanyew: No, sorry. I honestly have no idea. It's a pair of bits?
Please though, I'm glad that you found a proper solution for this, I really am.
But you really don't have to educate me on octets, it's fine. I fully admit that I have no idea, so please give me a bit of a break here.
I have no desire to understand octets, nor do I have any practical use for it. If something cannot be interpreted as a simple value or different bits set, that's way out of my league. I'm sorry if I somehow gave you the wrong impression, but I am not a software engineer by any means, I'm probably completely unsuitable for tasks like this, and I'm fine with that. I just do what I can figure out sometimes for fun, Anyone can close this pull request if it's unsuitable, or I can close it if asked to, it's really no big deal.

beeanyew:
After a few hours of contemplation, realizing how I managed to get the entire thing backwards and being unable to figure out how to use already existing data in the code base to approach this problem, I'm going to close the pull request. If anyone that understands octects and whatnots wants to make a proper implementation, please refer them to the unresolved conversation between myself and @NicknineTheEagle in this PR.

PR #2074, opened 2022-10-09
Add screenshot feature

SirQuartz:
This pull request adds the ability to take screenshots from within the emulator.


Razzile:
nice work! I'm kinda surprised how easy this was to implement

src/xenia/app/emulator_window.cc
```diff
@@ -928,6 +936,41 @@ void EmulatorWindow::ToggleFullscreen() {
   SetFullscreen(!window_->IsFullscreen());
 }
 
+void EmulatorWindow::TakeScreenshot() {
+  xe::ui::RawImage myimage;
```
Triang3l, line 940: The `my` prefix has no semantic value here.

src/xenia/app/emulator_window.cc
```diff
@@ -928,6 +936,41 @@ void EmulatorWindow::ToggleFullscreen() {
...
+  if (emulator()->is_title_open()) {  // Make sure we have a game open first
+    GetGraphicsSystemPresenter()->CaptureGuestOutput(myimage);
+    ExportImage(myimage);
+  }
+}
+
+void EmulatorWindow::ExportImage(const xe::ui::RawImage& image) {
+  auto path = std::filesystem::path("screenshot.png");
```
Triang3l, line 948: The path can be passed as a temporary, without an intermediate variable. Though overall it's better to add some date and time (and also some resolution for multiple screenshots taken during the same second) instead of overwriting the previous scerenshot.

src/xenia/app/emulator_window.cc
```diff
@@ -928,6 +936,41 @@ void EmulatorWindow::ToggleFullscreen() {
...
+  auto file = std::ofstream(path, std::ios::binary);
+  if (!file.is_open()) {
+    XELOGE("Failed to open file for writing: {}", path.string());
+    return;
+  }
+
+  auto result = stbi_write_png_to_func(
+      [](void* context, void* data, int size) {
+        auto file = reinterpret_cast<std::ofstream*>(context);
+        file->write(reinterpret_cast<char*>(data), size);
```
Triang3l, line 964: `const char*`, not `char*`, is enough for the data.

src/xenia/app/emulator_window.h
```diff
@@ -60,6 +60,8 @@ class EmulatorWindow {
   void UpdateTitle();
   void SetFullscreen(bool fullscreen);
   void ToggleFullscreen();
+  void TakeScreenshot();
+  void ExportImage(const xe::ui::RawImage& image);
```
Triang3l, line 64: This looks like a generic function for saving images (or "exporting", though it's not clear what exactly "export" means there, "save", just like in that private function is a more descriptive word), but internally it calls all images "screenshot.png" even though it's not obvious from the name of the function that it should be used only for screenshots.

src/xenia/app/emulator_window.cc
```diff
@@ -755,6 +759,10 @@ void EmulatorWindow::OnKeyDown(ui::KeyEvent& e) {
     case ui::VirtualKey::kF11: {
       ToggleFullscreen();
     } break;
+
+    case ui::VirtualKey::kF12: {
+      TakeScreenshot();
+    } break;
```
Triang3l, line 765: Missing new line after the break.

src/xenia/app/emulator_window.cc
```diff
@@ -16,6 +16,7 @@
 #include <string>
 #include <utility>
 
+#include "stb/stb_image_write.h"
```
Triang3l, line 19: Missing `third_party` directory path.

src/xenia/app/emulator_window.cc
```diff
@@ -928,6 +936,41 @@ void EmulatorWindow::ToggleFullscreen() {
   SetFullscreen(!window_->IsFullscreen());
 }
 
+void EmulatorWindow::TakeScreenshot() {
+  xe::ui::RawImage myimage;
+  if (emulator()->is_title_open()) {  // Make sure we have a game open first
```
Triang3l, line 941: The comment it not necessary since it's obvious from the code itself. But, the check is also not strictly required, and everything works without it. For example, if game closing is implemented at some point, maybe we'd want to keep the last frame of the game displayed (and thus saveable), but that's still debatable from the UX point of view.

src/xenia/app/emulator_window.cc
```diff
@@ -928,6 +936,41 @@ void EmulatorWindow::ToggleFullscreen() {
...
+    GetGraphicsSystemPresenter()->CaptureGuestOutput(myimage);
```
Triang3l, line 942: Capturing may fail (if the game hasn't presented a single frame yet, or for other internal reasons), `CaptureGuestOutput` returns a `bool` indicating whether it has succeeded.

Triang3l, line 942: The presenter is not initialized immediately after the window is opened — see how the call to `SetupGraphicsSystemPresenterPainting` is done. Therefore, for some time (possibly seconds) after opening the emulator, `GetGraphicsSystemPresenter()` may return `nullptr`, and trying to take a screenshot may result in a crash due to dereferencing a null pointer.

Triang3l, line 942: Note that the guest output image, depending on the game, may have a different aspect ratio than the display. Halo 3 provides a 1152x640 image (1.8), for instance, while 16:9 is 1.777…

Triang3l (changes requested):

src/xenia/app/emulator_window.cc
Triang3l, line 948: It may be desirable to save the screenshot somewhere in a more user-faced directory. The best default location would likely be a subdirectory (preferably something like `Xenia/game name` with characters not supported in paths on the host file system filtered out, or `Xenia/title ID` if there's no game name provided in the XEX) in the user images directory in the OS. Other non-default options may be a directory specified in the config, and the executable directory in case the emulator was launched in the portable mode.

Triang3l (changes requested):

SirQuartz:
> nice work! I'm kinda surprised how easy this was to implement. Do we have a way of notifying users that the screenshot has been saved (maybe imgui has some kind of notification functionality)? Also I'd change the output filename to something like `{title or title ID} - {date}.png`

Thanks! I was honestly surprised too. It helped that the project already included a third-party library that would save image files and that there was already a method to capture the raw render output. Also, GitHub Copilot sleuthed about half of the solution, and I deduced the rest to get it working. Otherwise, I wouldn't have even known about the `stbi_write_to_func()` method.

I'll see about making the filename more descriptive, probably something like `"Screenshot - [Title_Name or Title_ID] - [Datetime] - [Time in nanoseconds since epoch].png"`. The time since the epoch in nanoseconds will look like a long random number, but also never be the same, it can only ever increase. It also increases at a rate faster than human reaction time, that way, there is no possible way to accidentally overwrite an existing screenshot, no matter how fast you press the button. It will always give you a unique filename.

And yeah, I can add a message box with imgui to say "Screenshot saved to [screenshot path]".

src/xenia/app/emulator_window.cc
SirQuartz: I think for purpose of debugging, this is a perfectly fine limitation. The intent is to capture only what's actually being rendered by the emulator, not the entire display. Anything beyond the bounds of the guest output image would be a black letterbox anyway.

SirQuartz: That's fine, we can just check to make sure that `GetGraphicsSystemPresenter()` isn't a nullptr. If it is, we can log the error.

src/xenia/app/emulator_window.h
SirQuartz: I think my reasoning at the time was that if there was ever a feature implemented in the future to allow you to save screenshots in a different way (i.e through a dialog, maybe even choose different image formats) then you'd want to pass that logic through `ExportImage();`.

Exporting in this context simply means doing something with the image, might be saving it, might be converting it, etc. Either way, I think I'll rename it to `ExportScreenshot()` instead for clarity.

Wunkolo:
> I'll see about making the filename more descriptive, probably something like `"Screenshot - [Title_Name or Title_ID] - [Datetime] - [Time in milliseconds since epoch].png"`

Maybe something like [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)? So the date format is `%Y%m%dT%H%M%S` which maps to time-strings such as `20221011T102845` which is sortable alphabetically in a file-browser and pretty human-readable.

Razzile:
> > nice work! I'm kinda surprised how easy this was to implement. Do we have a way of notifying users that the screenshot has been saved (maybe imgui has some kind of notification functionality)? Also I'd change the output filename to something like `{title or title ID} - {date}.png`
> 
> Thanks! I was honestly surprised too. It helped that the project already included a third-party library that would save image files and that there was already a method to capture the raw render output. Also, GitHub Copilot sleuthed about half of the solution, and I deduced the rest to get it working. Otherwise, I wouldn't have even known about the `stbi_write_to_func()` method.
> 
> I'll see about making the filename more descriptive, probably something like `"Screenshot - [Title_Name or Title_ID] - [Datetime] - [Time in milliseconds since epoch].png"`. The time since the epoch in milliseconds will look like a long random number, but also never be the same, it can only ever increase. That way, it can give a unique file name.
> 
> And yeah, I can add a message box with imgui to say "Screenshot saved to [screenshot path]".

You used github copilot for this? How did that process work?

SirQuartz:
Yeah, GitHub Copilot is surprisingly helpful in finding the missing pieces to a solution at times. Especially in this case where the answer was just sitting right in front of me but I hadn't noticed it yet. In this case, I didn't know the project had a save-to-PNG method, but GitHub Copilot did, which saved me a lot of time. The only way I would have known about that would have been by searching through every file in the project until I randomly stumbled upon it.

https://user-images.githubusercontent.com/62965063/195423727-616e4354-608a-416a-a89a-fea95082c459.mp4

GitHub Copilot will use the context of the rest of your project to help find what you're looking for, and it does it remarkably quickly. It doesn't always work perfectly, and won't always give you what you're looking for. Most of the time you still need to fill in the rest or make changes. Overall, it's still very useful and I'd recommend it.

I'll see about changing the title instead of showing a message box when a screenshot is saved.

Razzile:
> Yeah, GitHub Copilot is surprisingly helpful in finding the missing pieces to a solution at times. Especially in this case where the answer was just sitting right in front of me but I hadn't noticed it yet. In this case, I didn't know the project had a save-to-PNG method, but GitHub Copilot did, which saved me a lot of time. The only way I would have known about that would have been by searching through every file in the project until I randomly stumbled upon it.
> Github.Copilot.mp4
> 
> GitHub Copilot will use the context of the rest of your project to help find what you're looking for, and it does it remarkably quickly. It doesn't always work perfectly, and won't always give you what you're looking for. Most of the time you still need to fill in the rest or make changes. Overall, it's still very useful and I'd recommend it.
> 
> I'll see about changing the title instead of showing a message box when a screenshot is saved.

Wow that's actually pretty insane. This is the first time I've seen github copilot be used for anything other than hello world or basic algorithms

SirQuartz:
> Wow that's actually pretty insane. This is the first time I've seen github copilot be used for anything other than hello world or basic algorithms

I've seen it construct some really advanced functions that would've taken me hours or days to figure out. This is actually a pretty decent example of a practical use for it in the real world.

Masamune3210:
Does it still have the issue of randomly spitting out chunks of other people's code? Or has it been trained enough for it to be genericized

SirQuartz:
> Does it still have the issue of randomly spitting out chunks of other people's code? Or has it been trained enough for it to be genericized

I've personally never seen that happen, but I have the "suggestions matching public code" option disabled so maybe that's why? Either way, it's pretty good at coming up with original solutions, as is the case in this example. Like I said, it favors using the context of your project and what it already has for methods to come up with a solution over just randomly yoinking one from the internet.

Copilot

src/xenia/app/emulator_window.cc
```diff
@@ -928,6 +937,68 @@ void EmulatorWindow::ToggleFullscreen() {
   SetFullscreen(!window_->IsFullscreen());
 }
 
+void EmulatorWindow::TakeScreenshot() {
+  xe::ui::RawImage image;
+  if (!GetGraphicsSystemPresenter()->CaptureGuestOutput(image) ||
```
Triang3l, line 942: Note that not all games output the image with the same aspect ratio as the display. Halo 3, for example, presents a 1152x640 image, which has an aspect ratio of 16.2:9 rather than 16:9, so it may appear slightly stretched horizontally. Also, we apply FXAA before writing to the guest output image (for simplicity, as it needs to work at the game's resolution), but CAS or FSR after that — so this will include just the FXAA-processed image with the size provided by the game.

Triang3l (changes requested):

PR #2075, opened 2022-10-10
Add message box when emulator fails to load invalid or corrupted file.

SirQuartz:
This pull request implements a message box when the user attempts to open an invalid or corrupt file format, either by dragging and dropping or by opening it through the file system.

Before, the emulator would just immediately terminate if this happened, which is something that should be avoided whenever possible because it's not particularly constructive or informative for the user. In this case it **_is_** completely avoidable, so it might as well be avoided.

With this pull request the emulator will no longer immediately crash, it will instead show a message box that actually tells the user what went wrong.

https://user-images.githubusercontent.com/62965063/194938414-00e7e7f6-d651-4974-b842-ccfc37cdab57.mp4

Xenia Message Box


src/xenia/app/emulator_window.cc
```diff
@@ -806,7 +806,11 @@ void EmulatorWindow::FileDrop(const std::filesystem::path& filename) {
   }
   auto result = emulator_->LaunchPath(filename);
   if (XFAILED(result)) {
-    // TODO: Display a message box.
+    xe::ui::ImGuiDialog::ShowMessageBox(
+        imgui_drawer_.get(), "Error: Failed to Launch",
+        fmt::format("Failed to launch target {}",
+                    filename.filename().string() +
```
CookiePLMonster, line 812:
```suggestion
                    xe::path_to_utf8(filename.filename()) +
```

src/xenia/app/emulator_window.cc
```diff
@@ -838,7 +842,11 @@ void EmulatorWindow::FileOpen() {
     auto abs_path = std::filesystem::absolute(path);
     auto result = emulator_->LaunchPath(abs_path);
     if (XFAILED(result)) {
-      // TODO: Display a message box.
+      xe::ui::ImGuiDialog::ShowMessageBox(
+          imgui_drawer_.get(), "Error: Failed to Launch",
+          fmt::format("Failed to launch target {}",
+                      abs_path.filename().string() +
```
CookiePLMonster, line 848:
```suggestion
                      xe::path_to_utf8(abs_path.filename()) +
```

CookiePLMonster (changes requested):
The current approach will break with Unicode characters present in the filename. `xe::path_to_utf8` will help.

Issue #1817, opened 2021-05-24
FinalizationPass::Run only sets Label::name properly for labels 0-9999


emoose:
Moving this issue over from https://github.com/xenia-project/game-compatibility/issues/245#issuecomment-843539147, `FinalizationPass::Run` only allocates a buffer for `Label::name` with 4 chars for the ID, limiting it to storing 0-9999: https://github.com/xenia-project/xenia/blob/800db1aa49814e12044809b52ed69d0dd941a740/src/xenia/cpu/compiler/passes/finalization_pass.cc#L47-L49

Labels that have IDs larger than 9999 will have it truncated in the name string, usually causing Xbyak to crash out, I guess due to labels sharing the same name.

This can be bypassed by enabling `disassemble_functions=true`, as that seems to setup `Label::name` using a different format for the name before FinalizationPass::Run is executed. Obviously not a great fix though, but still pretty neat that it works.

There's a few games in the compatibility tracker that mention only working with `disassemble_functions=true` set, probably mostly all related to this issue.
Extending the buffer that gets allocated seems to let some of them work (simplest way is probably allocating 8 bytes and storing the ID as hex?), at least Avatar: TLA worked without needing `disassemble_functions` set.

Not sure if this is actually the root cause though, could something else be going wrong for it to be creating Labels with IDs greater than 9999?

Margen67:
I've created a game-compatibility label for this: [requires_disassemble_functions_true](https://github.com/xenia-project/game-compatibility/labels/requires_disassemble_functions_true)

Triang3l:
What exactly is limiting this string to a fixed size at all?

PR #2081, opened 2022-10-27
[CPU] Increase amount of possible labels used in FinalizationPass

Gliniak:
Related to: https://github.com/xenia-project/xenia/issues/1817

I've got title mentioned in this issue (https://github.com/xenia-project/game-compatibility/issues/245) and found out after hitting assertion and checking which function causes issues that indeed it is caused by lack of space for labels.

They have this huge routine in their code that causes this crash: 
image

I've decided that instead of increasing label_len it will be better to just use hexadecimal notation.

### DIFF
--- a/src/xenia/cpu/compiler/passes/finalization_pass.cc
+++ b/src/xenia/cpu/compiler/passes/finalization_pass.cc
@@ -46,8 +46,8 @@ bool FinalizationPass::Run(HIRBuilder* builder) {
       if (!label->name) {
         const size_t label_len = 6 + 4;
         char* name = reinterpret_cast<char*>(arena->Alloc(label_len + 1, 1));
-        assert_true(label->id <= 9999);
-        auto end = fmt::format_to_n(name, label_len, "_label{}", label->id);
+        assert_true(label->id <= 65535);
+        auto end = fmt::format_to_n(name, label_len, "_label{:04X}", label->id);
         name[end.size] = '\0';
         label->name = name;
       }


emoose:
Nice, good to see the function label stuff is being looked into.

About your message on discord:
>For whatever reason in certain games if you run release build with debug_info_flags set to 0 it will crash, HOWEVER if you set that flag even to invalid one it will go on correctly or just run debug build. I removed line 127 and just reset debug_info always and it resolved crashes

IIRC having any debug_info_flag set at all lets debug_info get filled in, with that set `PPCHIRBuilder` then calls into `AnnotateLabel`: https://github.com/xenia-project/xenia/blob/a37b57ca8dd0b86b0476156d30da18d001aea90f/src/xenia/cpu/ppc/ppc_hir_builder.cc#L283-L287 (or it might be the other AnnotateLabel call inside there, both seem to check for debug_info first at least)

PPCHIRBuilder::AnnotateLabel then gives the function a label which doesn't have the 4 character limit:
https://github.com/xenia-project/xenia/blob/a37b57ca8dd0b86b0476156d30da18d001aea90f/src/xenia/cpu/ppc/ppc_hir_builder.cc#L244-L249

So AFAIK the `disassemble_functions=true` method was mostly just a way for users to set debug_info_flags so that debug_info would be filled in, letting AnnotateLabel take over the labelling.

Really it's a little strange that there's two different ways that functions can be labelled, maybe it'd be good to have a way for both `PPCHIRBuilder` & `FinalizationPass` classes to share a single naming function.

Gliniak:
> IIRC having any debug_info_flag set at all lets debug_info get filled in, with that set `PPCHIRBuilder` then calls into `AnnotateLabel`

Just to be clear that message isn't related to this particular case.

Yes, it fills. However how about debug build? it doesn't fill it and in that specific title I tested it on doesn't crash. (https://github.com/xenia-project/game-compatibility/issues/1536) That is super sketchy in my opinion

PR #2082, opened 2022-10-28
[UI] Resolved issue with next window disappearing after providing input

Gliniak:
Steps to reproduce:
1. Open input window in game that support it.
2. Enter something and confirm with Enter.
3. Open window again.

Cause:
Closing window wasn't clearing input that was provided and it was still in queue in state pressed down (especially enter button)

PS. Not sure if this solution works correctly in every case (multiple input windows), but I've never seen any game that would open many windows at the same time, so it is good enough.

### DIFF
--- a/src/xenia/ui/imgui_drawer.cc
+++ b/src/xenia/ui/imgui_drawer.cc
@@ -94,12 +94,12 @@ void ImGuiDrawer::RemoveDialog(ImGuiDialog* dialog) {
       presenter_->RemoveUIDrawerFromUIThread(this);
     }
     window_->RemoveInputListener(this);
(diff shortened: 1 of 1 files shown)


Triang3l:
This would clear all the input if _any_ dialog is closed, but not just the last dialog and thus the whole ImGui drawer (as after detaching the drawer, it stops receiving input events).

The original issue is fixed in 778333b1b50f9c50b5aaa0010346aa0855840292 — `ClearInput` was done only when removing the last dialog from outside, but not when removing it from the draw function of some dialog (I forgot to add the call to the deferred detaching code in the end of `ImGuiDrawer::Draw`).

PR #2087, opened 2022-11-09
Added Missing LICENSE for dlmalloc

p01arst0rm:
The License for the dlmalloc dependency currently does not have a license included alongside the file; add the license file for the dlmalloc project (Creative Commons Zero 1.0 Universal)


Razzile:
Good spot. I don't have merge powers unfortunately.

Triang3l:
Where are we using dlmalloc at this point? :o

Gliniak:
Seems like it is unused and can be removed. 
I've just build Xenia (at least on windows) with success without it

p01arst0rm:
quite possible, ive noticed theres a couple orphaned dependencies through reviewing the buildsystem recently

PR #2092, opened 2022-11-29
[UI] Added recently opened titles list

Gliniak:
This is rewritten and upgraded version of PR: https://github.com/xenia-project/xenia/pull/1396 (Thanks orginal creator for this idea)

What it changes in regards to original PR:
- Uses cpptoml to handle loading/storing entries
- Allows user to define how many entries it can store and show (Probably there should be some hard limit on viewing like 30)
- Instead of showing path it shows title name from titlebar. In special cases when there is no title name it uses path
- Stores only titles that were correctly initialized

Another benefit is that it is easily expandable to track for example last run time or amount of times certain game was booted


Gliniak:
Small update:
- Added storing last run time
- Safer data loader
- A bit of cleanup (removed unnecessary auto)

PR #2096, opened 2022-12-06
added fxaa LICENSE file

p01arst0rm:
The License for the fxaa header dependency currently does not have a license included alongside the file; add the license file for the FXAA3_11.H file (BSD 3-Clause License)


Triang3l:
The license is in the file itself (it was copied directly from a sample application).

PR #2102, opened 2022-12-29
moved clang-format to tools directory

p01arst0rm:
Moved the vswhere tool into the /tools directory.


Triang3l:
It's a third-party component, why (and also why #2098)? For the ease of managing licenses and updating, we don't put, for instance, the Vulkan headers in `src/xenia/ui/vulkan`.

gibbed:
I accepted #2098 because vswhere is a binary, I don't think we need to move `clang-format`.

Devman219:
Thank you

p01arst0rm:
> It's a third-party component, why (and also why #2098)? For the ease of managing licenses and updating, we don't put, for instance, the Vulkan headers in `src/xenia/ui/vulkan`.

using a single directory to contain all third party code is a legacy buildsystem design technique. placing tools in directories that align with their actual project use is easier;- it also allows you to ignore entire directories to exclude them from packages. newer project layouts place documentation in `/doc` tools in `/tools` data in `/data` and so on

There is also some code within `third_party` which is actually xenia project code, the `third_party` label appears to be entirely arbitrary .

PR #2106, opened 2023-01-05
[NET] NetDll___WSAFDIsSet: Fixed incorrect endianness of fd_count

Gliniak:
I did a little bit of an oopsie back when I was implementing this function.

I also added check to prevent going over 64 entries (hardcoded in structure amount)
Thanks to Bo98 for pointing that out


Triang3l:
Is the kernel supposed to return any error code in case of an overflow of the 64 limit?

Gliniak:
> Is the kernel supposed to return any error code in case of an overflow of the 64 limit?

No. It's ItSet call and only can return 0 or 1, however I see different thing now. I somehow forgot to add nullptr check for 2nd argument

PR #2105, opened 2022-12-29
[UI] Add include to fix compiling

Shoegzer:

PR #2118, opened 2023-01-26
[UI] Add missing #include for optional

roy-stewart:


src/xenia/ui/imgui_drawer.h
```diff
@@ -14,6 +14,7 @@
 #include <cstdint>
 #include <memory>
 #include <vector>
+#include <optional>
```
Triang3l, line 17: Hi, thank you for your fix — and welcome to Xenia!

`optional` should be between `memory` and `vector` alphabetically — overall, please apply `clang-format` (via `xb format` in our build script, after `git add`, and before another `git add` and `git commit`), which sorts includes among many other things, to all the changes submitted to Xenia.

Also, please add `[UI] ` to the beginning of the commit message, and prefer the present tense (what the commit itself does — "Add") as opposed to the past ("Added").

Triang3l (changes requested):

src/xenia/ui/imgui_drawer.h
roy-stewart: Thanks for your feedback and sorry about those issues. I had totally missed them. I committed the file again after I ran `xb format` and I fixed the commit message.

Shoegzer:
I've already submitted PR #2105 for this.

Triang3l:
I've merged the original pull request. By the way, while we, as far as I know, can do this in the GitHub UI while merging (although I'm not sure if it allows squash + rebase, or only squash + merge, as we do the former to avoid cluttering the history with merge commits) by ourselves, it's preferable that pull requests include just one commit (or multiple architecturally separated commits), especially without a "version history" within a pull request.

To modify the latest change — the contents and/or the message — you can `git add` your amendments, and then perform `git commit --amend` instead of just `git commit`. This will combine your new changes and the existing commit into one — and it also allows you to specify a different commit message for the new combined commit (if you want to only edit the message, you can run `git commit --amend` without making any new changes). Note that if you have already pushed your branch with the old commit, you'll have to use `git push --force` instead of `git push` after amending.

In addition, you can "squash" multiple commits into one using one of the many methods provided at https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together — though again, the requirement for `git push --force` applies here too.

PR #2120, opened 2023-01-28
[Kernel/Thread] Set TLS slot to 0 while freeing

Gliniak:
This fixes https://github.com/xenia-project/game-compatibility/issues/721 requirement for ``protect_zero = false`` and further random crashes

### DIFF
--- a/src/xenia/kernel/kernel_state.cc
+++ b/src/xenia/kernel/kernel_state.cc
@@ -138,7 +138,17 @@ void KernelState::set_process_type(uint32_t value) {
 
 uint32_t KernelState::AllocateTLS() { return uint32_t(tls_bitmap_.Acquire()); }
 
-void KernelState::FreeTLS(uint32_t slot) { tls_bitmap_.Release(slot); }
+void KernelState::FreeTLS(uint32_t slot) {
+  const std::vector<object_ref<XThread>> threads =
+      object_table()->GetObjectsByType<XThread>();
+
+  for (const object_ref<XThread>& thread : threads) {
+    if (thread->is_guest_thread()) {
+      thread->SetTLSValue(slot, 0);
+    }
+  }
+  tls_bitmap_.Release(slot);
+}
 
 void KernelState::RegisterTitleTerminateNotification(uint32_t routine,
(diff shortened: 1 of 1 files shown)


src/xenia/kernel/kernel_state.cc
```diff
@@ -138,7 +138,17 @@ void KernelState::set_process_type(uint32_t value) {
 
 uint32_t KernelState::AllocateTLS() { return uint32_t(tls_bitmap_.Acquire()); }
 
-void KernelState::FreeTLS(uint32_t slot) { tls_bitmap_.Release(slot); }
+void KernelState::FreeTLS(uint32_t slot) {
+  const std::vector<object_ref<XThread>> threads =
+      kernel_state()->object_table()->GetObjectsByType<kernel::XThread>();
```
gibbed, line 143: You shouldn't need to have to do `kernel_state()` here, you're already in `KernelState`.

gibbed (changes requested):

src/xenia/kernel/kernel_state.cc
Gliniak: Fair point. That's habit for me at this point 😅 Fixed

src/xenia/kernel/kernel_state.cc
```diff
@@ -138,7 +138,17 @@ void KernelState::set_process_type(uint32_t value) {
 
 uint32_t KernelState::AllocateTLS() { return uint32_t(tls_bitmap_.Acquire()); }
 
-void KernelState::FreeTLS(uint32_t slot) { tls_bitmap_.Release(slot); }
+void KernelState::FreeTLS(uint32_t slot) {
+  const std::vector<object_ref<XThread>> threads =
+      object_table()->GetObjectsByType<kernel::XThread>();
```
gibbed, line 143: No need for `kernel::` namespace here.

gibbed (changes requested):

src/xenia/kernel/kernel_state.cc
Gliniak: Fixed

Issue #2119, opened 2023-01-26
Random error message displayed when opening game on linux


roy-stewart:
### Validation

- [X] I've read the [FAQ](https://github.com/xenia-project/xenia/wiki/FAQ).
- [X] The Xenia build used is from the master branch (not MLBS/AlexVS/Canary/pull requests, etc.)
- [X] This issue isn't for tech support (help with Xenia).
- [X] If this issue occurs in a specific game, I've done analysis to locate the faulty subsystem of the emulator and a potential reason in it.
- [X] I've checked if this issue hasn't already been reported.
- [X] My device meets the minimum requirements: https://github.com/xenia-project/xenia/wiki/Quickstart#system-requirements
- [X] (If building) I have read the building doc: https://github.com/xenia-project/xenia/blob/master/docs/building.md

### Describe what's going wrong

When attempting to open any game in Xenia on Manjaro Linux, the following error message is displayed once the file is selected.
![xenia-error-screenshot](https://user-images.githubusercontent.com/16023378/214949768-5841fdd8-9c72-43a2-a056-0c75402b46e5.png)
When I try to exit out of the error message Xenia freezes.
My system is set to use English as the language but the error message is using a different set of characters. After passing the screen shot through google translate it returns a message of random letters and numbers.

### Describe what should happen

When a game is selected from the file picker it should launch the game.

### If applicable, upload a logfile and link it here

The log file after the error was encountered.
[xenia.log](https://github.com/xenia-project/xenia/files/10513379/xenia.log)

PR #2121, opened 2023-01-29
[Base] Filter out relative directories on linux

hd-genius:
Partial fix for #2119. Filters out the . and .. directory entries on linux.


Triang3l (changes requested):
Please `--amend` the `[Base] ` prefix to the commit message.

hd-genius:
Sorry about that, I thought the prefix was only for the merge message. Thanks for letting me know!

PR #2122, opened 2023-01-29
[Base] Set the path for posix file info

hd-genius:
Partial fix for #2119. The path to the file was not being set on posix devices and was preventing fragment files from being opened.


Triang3l (changes requested):
Please `--amend` the `[Base] ` prefix to the commit message.

PR #2123, opened 2023-02-05
[STFS] Add critical section to file reading

Gliniak:
Affected titles:
https://github.com/xenia-project/game-compatibility/issues/827
https://github.com/xenia-project/game-compatibility/issues/125 

Description:
These titles use multiple threads to load same file, which in case of extracted file doesn't make any issue, however for STFS package it is huge problem as it can randomly receive incorrect data.

I checked it multiple times and calculated hashes of loaded data between extracted and packed game.
Conclusion for me is simple, at certain point thread starts getting invalid data, which is simply resolved by including critical section

### DIFF
--- a/src/xenia/vfs/devices/stfs_container_file.cc
+++ b/src/xenia/vfs/devices/stfs_container_file.cc
@@ -52,12 +52,15 @@ X_STATUS StfsContainerFile::ReadSync(void* buffer, size_t buffer_length,
     size_t read_length =
         std::min(record.length - read_offset, remaining_length);
 
-    auto& file = entry_->files()->at(record.file);
-    xe::filesystem::Seek(file, record.offset + read_offset, SEEK_SET);
-    auto num_read = fread(p, 1, read_length, file);
+    auto global_lock = global_critical_region_.Acquire();
+    {
+      auto& file = entry_->files()->at(record.file);
+      xe::filesystem::Seek(file, record.offset + read_offset, SEEK_SET);
+      auto num_read = fread(p, 1, read_length, file);
 
-    *out_bytes_read += num_read;
-    p += num_read;
+      *out_bytes_read += num_read;
+      p += num_read;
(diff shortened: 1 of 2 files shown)


src/xenia/vfs/devices/stfs_container_file.cc
```diff
@@ -52,12 +52,15 @@ X_STATUS StfsContainerFile::ReadSync(void* buffer, size_t buffer_length,
     size_t read_length =
         std::min(record.length - read_offset, remaining_length);
 
-    auto& file = entry_->files()->at(record.file);
-    xe::filesystem::Seek(file, record.offset + read_offset, SEEK_SET);
-    auto num_read = fread(p, 1, read_length, file);
+    auto global_lock = global_critical_region_.Acquire();
```
Triang3l, line 55: Is that a race between guess threads accessing the _same_ guest file handle, or _different_ guest handles implemented through one host file handle? If that's the latter, we should be opening the host container file multiple times, for each guest handle, or if possible, use a memory mapping. Using the global critical region here will create a lot of contention that may block other threads for milliseconds — the comment in `xenia/base/mutex.h` explicitly says that no I/O must be done there.

Triang3l (changes requested):

src/xenia/vfs/devices/stfs_container_file.cc
Gliniak: Ohh my bad for not reading comments in mutex.h. 
It uses multiple handles, so rewrites then to allow memory mapping probably

gibbed: That's strange, I thought it _was_ already using memory mapping for STFS mapping? Maybe I'm forgetting when it got changed. I know I wanted the capability to do both, since we need the ability to load STFS packages nested within another STFS package.

PR #2131, opened 2023-02-19
[vfs dump] Use source as out if out isn't provided

Margen67:


gibbed (commented):
Would this not try to immediately create a new directory with the same name as the original file, and always fail?

PR #2132, opened 2023-02-20
[XAM] xeXamContentCreate - Disposition cleanup

Gliniak:


src/xenia/kernel/xam/xam_content.cc
```diff
@@ -116,6 +116,8 @@ dword_result_t XamContentCreateEnumerator_entry(
 }
 DECLARE_XAM_EXPORT1(XamContentCreateEnumerator, kContent, kImplemented);
 
+enum class kDispositionState : uint32_t { Unknown = 0, Create = 1, Open = 2 };
```
Triang3l, line 119: I guess it's a bit too late oof… But generally in Xenia, the `k` prefix is used for the values rather than the type itself.

Triang3l (changes requested):

src/xenia/kernel/xam/xam_content.cc
Gliniak: Ohhh... My bad, not sure why I did it that way 🤔
Well, it's still easy to change and doesn't affect anything, so no big deal

PR #2135, opened 2023-02-25
[UI] Allow loading custom font

Gliniak:
- Unified font size to 12. This causes default UI to look a bit bigger
- Set oversample to 2 to make font more readable (especially custom fonts)


src/xenia/ui/imgui_drawer.cc
```diff
@@ -218,6 +192,58 @@ std::optional<ImGuiKey> ImGuiDrawer::VirtualKeyToImGuiKey(VirtualKey vkey) {
...
+
+  if (io.Fonts->Fonts.empty()) {
+    io.Fonts->AddFontFromMemoryCompressedBase85TTF(
+        kProggyTinyCompressedDataBase85, default_font_size, &font_config,
+        io.Fonts->GetGlyphRangesDefault());
+  }
+
+  // TODO(benvanik): jp font on other platforms?
+  // https://github.com/Koruri/kibitaki looks really good, but is 1.5MiB.
+  const char* jp_font_path = "C:\\Windows\\Fonts\\msgothic.ttc";
```
Margen67, line 232: This should use the `windir` environment variable.

Gliniak: Done, but in a way less prone to errors

gandalfthewhite19890404:
Hello!
Can you add more glyph ranges - not only Latin?

Gliniak:
Sure (at least on windows), do you know font that looks good and support these characters?

gandalfthewhite19890404:
> 

SegoeUI, Arial, Times New Roman.
There many of them, if some font does not have needed symbols - there will be a question marks. So adding glyphs to sources is safe change. If I talk wrong things - sorry, make edits.

Gliniak:
Not sure what's a better idea there.
To assume that custom font might include them or to always load another font that will contain them?

gandalfthewhite19890404:
What if bundle with emulator some free nice looking font as default, for example Koruri, it could be big - size is an issue?
And let user to choose his own font if he wants?

Gliniak:
I will suggest similar solution that is used in canary.

Priority of font loading:
1. Custom font
2. (on windows only) Tahoma from fonts directory

If no font is loaded then load default font.

+ at the end always load Japanese font

How Tahoma looks like (with font set to 16)
obraz

gandalfthewhite19890404:
Brilliant! Thank you very much!!!

src/xenia/ui/imgui_drawer.cc
```diff
@@ -218,6 +196,130 @@ std::optional<ImGuiKey> ImGuiDrawer::VirtualKeyToImGuiKey(VirtualKey vkey) {
...
+};
+
+const std::filesystem::path ImGuiDrawer::GetWindowsFont(std::string font_name) {
+  std::filesystem::path font_path = "";
+
+#if XE_PLATFORM_WIN32
+  PWSTR fonts_dir;
+  HRESULT result = SHGetKnownFolderPath(FOLDERID_Fonts, 0, NULL, &fonts_dir);
+  if (FAILED(result)) {
+    CoTaskMemFree(static_cast<void*>(fonts_dir));
```
Triang3l, line 214: `fonts_dir` would likely (though I'm not entirely sure about this specific case, of course) only be initialized if the function has succeeded, this freeing may cause a crash.

src/xenia/ui/imgui_drawer.cc
```diff
@@ -218,6 +196,130 @@ std::optional<ImGuiKey> ImGuiDrawer::VirtualKeyToImGuiKey(VirtualKey vkey) {
...
+    return "";
+  }
+
+  font_path = std::wstring(fonts_dir);
+  CoTaskMemFree(static_cast<void*>(fonts_dir));
+#endif
+  font_path.append(font_name);
```
Triang3l, line 221: A blank line above this line would make the code visually more structured.

src/xenia/ui/imgui_drawer.cc
```diff
@@ -218,6 +196,130 @@ std::optional<ImGuiKey> ImGuiDrawer::VirtualKeyToImGuiKey(VirtualKey vkey) {
...
 
+static const ImWchar font_glyph_ranges[] = {
+    0x0020, 0x00FF,  // Basic Latin + Latin Supplement
+    0x0370, 0x03FF,  // Greek
+    0x0400, 0x044F,  // Cyrillic
+    0x2000, 0x206F,  // General Punctuation
+    0,
+};
+
+const std::filesystem::path ImGuiDrawer::GetWindowsFont(std::string font_name) {
```
Triang3l, line 207: I think this might have been more platform-independent as other operating systems also have their own font storages, and possibly accepting the name (family, weight, italic, etc. possibly) of the font with the necessary lookups in the OS rather than directly the file name. The fallback behavior is also pretty non-obvious in this function. On Windows, it returns an empty path if it has failed to obtain the system fonts directory. However, on other platforms, it just silently tries to get the path from the working directory instead.

src/xenia/ui/imgui_drawer.cc
```diff
@@ -218,6 +196,130 @@ std::optional<ImGuiKey> ImGuiDrawer::VirtualKeyToImGuiKey(VirtualKey vkey) {
...
+  }
+  return true;
+}
+
+bool ImGuiDrawer::LoadWindowsFont(ImGuiIO& io, ImFontConfig& font_config,
+                                  const float font_size) {
+  const std::filesystem::path font_path = GetWindowsFont("tahoma.ttf");
+  if (!std::filesystem::exists(font_path)) {
+    XELOGW(
+        "Unable to find tahoma font in Windows fonts directory. Switching to "
```
Triang3l, line 258: tahoma → Tahoma

Triang3l (changes requested):

src/xenia/ui/imgui_drawer.cc
Gliniak: https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath

``The calling process is responsible for freeing this resource once it is no longer needed by calling [CoTaskMemFree](https://learn.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-cotaskmemfree), whether SHGetKnownFolderPath succeeds or not``

PR #2140, opened 2023-03-12
[Base] Added Listing directories and filtering filenames by regex

Gliniak:


src/xenia/base/filesystem.h
```diff
@@ -123,6 +124,9 @@ struct FileInfo {
 };
 bool GetInfo(const std::filesystem::path& path, FileInfo* out_info);
 std::vector<FileInfo> ListFiles(const std::filesystem::path& path);
+std::vector<FileInfo> ListDirectories(const std::filesystem::path& path);
+std::vector<FileInfo> FilterByName(const std::vector<FileInfo> files,
```
Gliniak, line 128: Whoopsie! To fix later

src/xenia/base/filesystem.h
```diff
@@ -123,6 +124,9 @@ struct FileInfo {
 };
 bool GetInfo(const std::filesystem::path& path, FileInfo* out_info);
 std::vector<FileInfo> ListFiles(const std::filesystem::path& path);
+std::vector<FileInfo> ListDirectories(const std::filesystem::path& path);
```
Triang3l, line 127: Did the files where these functions are called accidentally slip under the radar of `git add`?

Triang3l (changes requested):

src/xenia/base/filesystem.h
Gliniak: No, these will be used in the future and in different PRs.
It will simplify loading patches and simplify loading profile directories in the future

Triang3l: This may end up being removed as dead code at some point between the merging of this pull request and when the patches/profile changes arrive, or it may just deteriorate as nothing tests this code in practice, I think it'd be safer to submit these functions alongside the first use of them.

Gliniak: I agree with that, but it's quite annoying if that change would appear in many PRs.
Maybe instead it would be better to add note about future usage?

Especially when we consider how long resolving all quirks in bigger PRs might take

PR #2152, opened 2023-04-28
LLVM: Fix -0 handling in Accurate xfloat mode

Scofield1008:
Resulted incorrectly in -2 instead. Makes the accuracy test results closer to real hardware.


Gliniak:
Direct merge from another branch from the same repository (aka. current one). No changes included

PR #2156, opened 2023-05-13
Sonic unleashed d3d12 crash fix/ graphic device lost in d3d12

Megasonic999:
I'm just new, here so here's my request:
Could you please find a solution about the graphic device lost issue with d3d12, because it makes it imopssible to pleay some games (Especially Sonic Unleashed)
PS: Sorry if I did it wrong, because like I said, I'm new here.


Razzile:
Please stop spamming bogus PRs on multiple different accounts

PR #2199, opened 2023-06-02
[XMA] New XMA decoder

Gliniak:
[DRAFT PR]
I will write something more interesting here when I'm done with it.

For now this might introduce some new bugs and fix many older bugs.

Feel free to point to any issues, convoluted parts and things that requires improvement, but aren't marked as TODO


TorutheRedFox:
NFS games now actually work because of this PR

gg

JeBobs:
This has some sound issues within Burnout Paradise, it's like bits of other sounds play over top any new sound.

boxter0360:
How use this code on Xbox Xenia?

requeijaum:
Any news on this? Seems to fix Burnout Revenge also.

PR #2201, opened 2023-06-09
D3D12 patch request for sonic unleashed

Megasonic999:
Hi, so you all know that Sonic unleashed has that crashing issue on xenia on d3d12, so I thought if someone could create a patch that fixes the game's compatibility with d3d12. Would be much appreciated. Thank you.
PS: it's just an Idea that I thought could fix that issue. I don't mean to rush anyone or something.


Gliniak:
image

We cannot just merge whole canary branch into master

PR #2205, opened 2023-06-22
Add cvars for video modes

Aero002:
I've added configuration variables for Interlaced, Hi-Def, and Widescreen video modes. 
This update also eliminates the need for custom builds when testing games in standard definition fullscreen video modes.

PR #2206, opened 2023-06-23
Add cvars for video modes

Aero002:
I've added configuration variables for Interlaced, Hi-Def, and Widescreen video modes.
This update also eliminates the need for custom builds when testing games in standard definition fullscreen video modes.

Razzile, if you're seeing this, please review my PR.
(Reposted because margen is an asshole)


Margen67:
Duplicate of #2205

PR #2209, opened 2023-06-28
Log ISO path

Margen67:


src/xenia/emulator.cc
```diff
@@ -332,7 +332,7 @@ X_STATUS Emulator::LaunchDiscImage(const std::filesystem::path& path) {
   // Register the disc image in the virtual filesystem.
   auto device = std::make_unique<vfs::DiscImageDevice>(mount_path, path);
   if (!device->Initialize()) {
-    xe::FatalError("Unable to mount disc image; file not found or corrupt.");
+    xe::FatalError(fmt::format("Unable to mount disc image {}; file corrupt or not found.", xe::path_to_utf8(path)));
```
Triang3l, line 335: Please run clang-format to limit this line to 80 characters.

Triang3l (changes requested):

Triang3l:
Can you please add `[VFS] ` in the beginning of the commit message, and possibly split the two changes into two commits (a `[Core]` one for the error message, and a `[VFS]` one for the ISO path)?

Triang3l:
On a semi-unrelated note, I think for privacy reasons we should avoid printing the content directory path in the log as it contains the Windows user name, considering that we already have a button that opens it, thus it's not as important to users.

Triang3l:
That's not what I meant kinda 😜

PR #2220, opened 2023-08-01
[Testing] Added test project for vfs

Gliniak:
- Added test case for: decode_fat_timestamp
- Changed location of: decode_fat_timestamp
- Fixed invalid month decoding in decode_fat_timestamp

### DIFF
--- a/src/xenia/vfs/devices/stfs_xbox.h
+++ b/src/xenia/vfs/devices/stfs_xbox.h
@@ -10,11 +10,39 @@
 #ifndef XENIA_VFS_DEVICES_STFS_XBOX_H_
 #define XENIA_VFS_DEVICES_STFS_XBOX_H_
 
+#include <time.h>
+
+#include "xenia/xbox.h"
 #include "xenia/base/string_util.h"
 #include "xenia/kernel/util/xex2_info.h"
 
 namespace xe {
 namespace vfs {
+
+// Convert FAT timestamp to 100-nanosecond intervals since January 1, 1601 (UTC)
+inline uint64_t decode_fat_timestamp(const uint32_t date, const uint32_t time) {
+  struct tm tm = {0};
+  // 80 is the difference between 1980 (FAT) and 1900 (tm);
+  tm.tm_year = ((0xFE00 & date) >> 9) + 80;
+  tm.tm_mon = ((0x01E0 & date) >> 5);
+  tm.tm_mday = (0x001F & date) >> 0;
+  tm.tm_hour = (0xF800 & time) >> 11;
+  tm.tm_min = (0x07E0 & time) >> 5;
+  tm.tm_sec = (0x001F & time) << 1;  // the value stored in 2-seconds intervals
+  tm.tm_isdst = 0;
+
+#if XE_PLATFORM_WIN32
+  time_t timet = _mkgmtime(&tm);
+#else
+  time_t timet = timegm(&tm);
+#endif
+
+  if (timet == -1) {
+    return 0;
+  }
+  // 11644473600LL is a difference between 1970 and 1601
+  return (timet + 11644473600LL) * 10000000;
+}
 
 // Structs used for interchange between Xenia and actual Xbox360 kernel/XAM
 
--- a/src/xenia/vfs/devices/stfs_xbox.h
+++ b/src/xenia/vfs/devices/stfs_xbox.h
@@ -24,7 +24,7 @@ inline uint64_t decode_fat_timestamp(const uint32_t date, const uint32_t time) {
   struct tm tm = {0};
   // 80 is the difference between 1980 (FAT) and 1900 (tm);
   tm.tm_year = ((0xFE00 & date) >> 9) + 80;
-  tm.tm_mon = ((0x01E0 & date) >> 5);
+  tm.tm_mon = ((0x01E0 & date) >> 5) - 1;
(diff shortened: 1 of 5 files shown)


src/xenia/vfs/devices/stfs_xbox.h
```diff
@@ -10,12 +10,37 @@
 #ifndef XENIA_VFS_DEVICES_STFS_XBOX_H_
 #define XENIA_VFS_DEVICES_STFS_XBOX_H_
 
+#include "xenia/xbox.h"
 #include "xenia/base/string_util.h"
 #include "xenia/kernel/util/xex2_info.h"
 
+#if XE_PLATFORM_WIN32
+#include "xenia/base/platform_win.h"
+#define timegm _mkgmtime
```
Triang3l, line 19: It'd be less intrusive to use:
```cpp
#if XE_PLATFORM_WIN32
time_t timet = _mkgmtime(&tm);
#else
time_t timet = timegm(&tm);
#endif
```
at the call location instead of redefining a platform function name to something else (or even a `src/xenia/base` wrapper function or at least a `XE_` definition for better reusability, but if it's needed in only one place that's not necessary). The previously used approach was relatively fine in `.cc` files (though still pretty dirty), but in headers it's dangerous as it may affect other headers included after this one.

src/xenia/vfs/devices/stfs_xbox.h
```diff
@@ -10,12 +10,37 @@
...
+#endif
+
 namespace xe {
 namespace vfs {
 
+// Convert FAT timestamp to 100-nanosecond intervals since January 1, 1601 (UTC)
+static uint64_t decode_fat_timestamp(const uint32_t date, const uint32_t time) {
```
Triang3l, line 26: I'm not entirely sure why (maybe `static` results in a new copy of the function being created in every translation unit it's used in where it actually happens to be compiled to a subroutine, possibly increasing the size of the executable), but Xenia's convention is to use `inline` instead of `static` for functions defined in headers.

Triang3l (changes requested):

src/xenia/vfs/devices/stfs_xbox.h
```diff
@@ -10,12 +10,37 @@
...
+  struct tm tm = {0};
+  // 80 is the difference between 1980 (FAT) and 1900 (tm);
+  tm.tm_year = ((0xFE00 & date) >> 9) + 80;
+  tm.tm_mon = ((0x01E0 & date) >> 5) - 1;
```
Triang3l, line 30: This is a functional change, so it needs to be reflected in the first line of the commit message for easier tracing of changes. But ideally it should even be in a separate commit (preferably the month fix in the first, everything testing-related in the second, so the latter doesn't reintroduce broken code in a different location).

Triang3l (changes requested):

src/xenia/vfs/devices/stfs_xbox.h
Gliniak: Sure. I will split it into a separate commit within this PR

src/xenia/vfs/devices/stfs_xbox.h
Gliniak: Ok. I will change it to inline

src/xenia/vfs/devices/stfs_xbox.h
```diff
@@ -10,12 +10,38 @@
...
+  tm.tm_mday = (0x001F & date) >> 0;
+  tm.tm_hour = (0xF800 & time) >> 11;
+  tm.tm_min = (0x07E0 & time) >> 5;
+  tm.tm_sec = (0x001F & time) << 1;  // the value stored in 2-seconds intervals
+  tm.tm_isdst = 0;
+
+  #if XE_PLATFORM_WIN32
```
Triang3l, line 32: The indentation of the conditions is inconsistent for some reason — did this slip past `clang-format`? The convention in Xenia is not to indent preprocessor conditionals (like the `#else` and `#endif` here).

Triang3l (changes requested):

src/xenia/vfs/devices/stfs_xbox.h
```diff
@@ -10,12 +10,38 @@
...
+#include "xenia/xbox.h"
 #include "xenia/base/string_util.h"
 #include "xenia/kernel/util/xex2_info.h"
 
 namespace xe {
 namespace vfs {
 
+// Convert FAT timestamp to 100-nanosecond intervals since January 1, 1601 (UTC)
+inline uint64_t decode_fat_timestamp(const uint32_t date, const uint32_t time) {
+  struct tm tm = {0};
```
Triang3l, line 22: This file requires `#include <time.h>` now, I think it should be sufficient for `struct tm` and both `_mkgmtime` and `timegm`.

Triang3l (changes requested):

src/xenia/vfs/devices/stfs_xbox.h
Gliniak: I double checked it. Seems like format doesn't care about this case.
I will change it manually

src/xenia/vfs/devices/stfs_xbox.h
Gliniak: Done

PR #2226, opened 2023-09-09
[LINUX/CPU/BASE] Added System-V ABI support in x64 backend & fixed mmap usage

guccigang420:
Current thunks in x64_backend.cc assumed that the calling convention is Microsoft ABI. Added an implementation for the System-V ABI which is used in Unix(-like) systems (Linux/MacOS/FreeBSD/etc).

When memory mapping a file, we should ensure it is large enough. Additionally, if "AllocFixed" is called with a nullptr base_address, the MAP_FIXED flag should not be used.

See:
https://wiki.osdev.org/System_V_ABI
https://man7.org/linux/man-pages/man2/mmap.2.html

### DIFF
--- a/src/xenia/base/memory_posix.cc
+++ b/src/xenia/base/memory_posix.cc
@@ -85,8 +85,11 @@ void* AllocFixed(void* base_address, size_t length,
                  AllocationType allocation_type, PageAccess access) {
   // mmap does not support reserve / commit, so ignore allocation_type.
   uint32_t prot = ToPosixProtectFlags(access);
-  void* result = mmap(base_address, length, prot,
-                      MAP_PRIVATE | MAP_FIXED | MAP_ANONYMOUS, -1, 0);
+  int flags = MAP_PRIVATE | MAP_ANONYMOUS;
+  if (base_address != nullptr) {
+    flags |= MAP_FIXED;
+  }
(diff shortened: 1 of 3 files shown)


src/xenia/base/memory_posix.cc
```diff
@@ -85,8 +85,13 @@ void* AllocFixed(void* base_address, size_t length,
...
   uint32_t prot = ToPosixProtectFlags(access);
-  void* result = mmap(base_address, length, prot,
-                      MAP_PRIVATE | MAP_FIXED | MAP_ANONYMOUS, -1, 0);
+  int flags = 0;
+  if (base_address != nullptr) {
+    flags = MAP_PRIVATE | MAP_FIXED | MAP_ANONYMOUS;
+  } else {
+    flags = MAP_PRIVATE | MAP_ANONYMOUS;
+  }
+  void* result = mmap(base_address, length, prot, flags, -1, 0);
```
Gliniak, line 94: You can simplify this whole fragment to: 

```
int flags = MAP_PRIVATE | MAP_ANONYMOUS;
if (base_address != nullptr) {
  flags |= MAP_FIXED;
}

guccigang420: Done.

PR #2227, opened 2023-09-18
[Premake] Added missing links to vfs-testing premake

Gliniak:


hd-genius:
I would like to add my support to merge this PR in. This is one of the changes needed to fix the current build issue for linux. A more in depth discussion on this change had taken place on an accidental duplicate PR #2277.

has207:
> Is it by design that the links are not propagated from `xenia-vfs`, or is that more of an issue with some of the Premake `action`s? Though as far as I understand, it's a pretty common pattern in build systems that static libraries don't have any dependencies on static libraries, only shared libraries and executables do, but seeing such inconsistency in Premake is somewhat weird. But I remember from making `premake-androidndk` that I had to write A LOT of code specifically for gathering all the needed libraries from the dependency tree, including not only `links`, but also `-l` linker flags. Do you use the `cmake` action or `gmake2`?

I think this is working as intended, by which I mean that we do need explicit deps on xenia-base and fmt in xenia-vfs-tests. xenia-vfs is a static lib so doesn't propagate its dependencies by design, and testing code in test_suite.lua attempts to compile it with src/xenia/base/console_app_main_$platform_suffix.cc which makes it necessary to add the link targets here. FWIW it's the same pattern in other tests such as src/xenia/cpu/testing/premake5.lua which also add these the sam way.

PR #2230, opened 2023-10-01
[LINUX/BASE] Fixed mmap usage and (re)-implemented most parts of memory_posix.cc

guccigang420:
As the title says. Previously the MapFileView function did not take into account the handle argument, as it used the MAP_ANONYMOUS and the MAP_PRIVATE flags.

Additionally, AllocFixed used mmap and DeallocFixed used munmap always, with the MAP_FIXED flag. This was not ideal for several reasons:

1. Invoking mmap the way it was invoked silently unmaps the range from the file it was previously mapped to.
2. MAP_FIXED was used even with a nullptr base_address.
3. Mapped file memory should not be unmapped in DeallocFixed.

For more information regarding mmap, see:
https://man7.org/linux/man-pages/man2/mmap.2.html


src/xenia/base/memory_posix.cc
```diff
@@ -178,12 +218,37 @@ void CloseFileMappingHandle(FileMappingHandle handle,
 void* MapFileView(FileMappingHandle handle, void* base_address, size_t length,
                   PageAccess access, size_t file_offset) {
   uint32_t prot = ToPosixProtectFlags(access);
-  return mmap64(base_address, length, prot, MAP_PRIVATE | MAP_ANONYMOUS, handle,
+
+  int flags = MAP_SHARED;
+  if (base_address != nullptr) {
+    flags = flags | MAP_FIXED_NOREPLACE;
```
Gliniak, line 224: flags |= MAP_FIXED_NOREPLACE;

src/xenia/base/memory_posix.cc
```diff
@@ -178,12 +218,37 @@ void CloseFileMappingHandle(FileMappingHandle handle,
...
+  }
+
+  void* result = mmap(base_address, length, prot, flags, handle,
                 file_offset);
+
+  if (result == MAP_FAILED) {
+    return nullptr;
+  }else {
```
Gliniak, line 232: else is unnecessary here

src/xenia/base/memory_posix.cc
```diff
@@ -96,6 +127,15 @@ void* AllocFixed(void* base_address, size_t length,
 
 bool DeallocFixed(void* base_address, size_t length,
                   DeallocationType deallocation_type) {
+  size_t region_begin = (size_t)base_address;
```
Gliniak, line 130: constify

guccigang420:
So I did implement some of these recommendations. I also added a mutex to avoid race conditions in multi-threaded scenarios.

src/xenia/base/memory_posix.cc
```diff
@@ -96,20 +138,97 @@ void* AllocFixed(void* base_address, size_t length,
...
+  }
+
+  switch(deallocation_type) {
+    case DeallocationType::kDecommit:
+      return Protect(base_address, length, PageAccess::kNoAccess);
+    case DeallocationType::kRelease:
+      return munmap(base_address, length) == 0;
+    default:
+      assert_unhandled_case(deallocation_type);
+  }
```
RodoMa92, line 168: In a release build the compiler here complained that there is no return value for all control paths. Just add a return false at the end of the function just to make it happy for now.

PR #2247, opened 2024-02-08
[x64] Add GFNI and AVX512VBMI optimization for `VECTOR_SHL`(int8)

Wunkolo:
Uses `{v}gf2p8mulb` as a general int8-multiplication instruction to simulate bit-shifts.

Verified the implementation with `instr__gen_vslb` from #1348. Can be grabbed with:
```
git checkout origin/gen_tests -- src\xenia\cpu\ppc\testing\*vslb.s
```


antoniodesousa:
You probably have better chance of having this merged in the Canary repo:
https://github.com/xenia-canary/xenia-canary

Wunkolo:
Moved this over to canary

PR #2250, opened 2024-02-26
[3PP] Replace cpptoml with tomlplusplus

Gliniak:
Reasoning:
 - Cpptoml is obsolete.

Benefits:
 - toml++ is still developed and maintained.
 - Good documentation.
 - Supports UTF8 with BOM, so tricks skip BOM can be removed.
 - <probably more>

PS. Not sure if it fully works right now and I'm not sure if functions from toml_internal namespace are still required


Gliniak:
I don't know, but I agree. 
I was surprised how easy it was to make it work, but I'm still not 100% sure if it works as intended. I saw that in specific cases it works differently. For example it replaces 1.0 with 1

Gliniak:
I double checked it and it seems like it works flawlessly. Please review it just in case

PR #2292, opened 2024-12-19
Corrected the premake binary search functionality

hd-genius:
Currently the build premake step is failing on linux due to a dependency issue in premake. This issue propmpted me to add a prebuilt premake to my local project but I noticed that it was not being acknowledge during the xb  build process. These changes fix the executable search logic in the xb and premake python scripts so that drop in prebuilt binaries can be used.

### DIFF
--- a/tools/build/premake
+++ b/tools/build/premake
@@ -12,12 +12,14 @@ import json
 import os
 import shutil
 import subprocess
+import pathlib
 import sys
 import re
+from helpers import is_executable
 
 
 self_path = os.path.dirname(os.path.abspath(__file__))
-root_path = os.path.join(self_path, '..', '..')
+root_path = pathlib.Path(self_path).parent.parent.absolute()
 premake_submodule_path = os.path.join(root_path, 'third_party', 'premake-core')
 premake_path = premake_submodule_path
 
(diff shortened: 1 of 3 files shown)


nullequal:
due to commit [fe71eb7](https://github.com/xenia-project/premake-core/commit/fe71eb790c7d085cd3c6a7b71a50167b4da06e69) you would have to fork the xenia-project fork and rebase to premake master then build
so it would be better if the devs updated the fork and let `xb setup` build the executable

tools/build/premake
```diff
@@ -54,15 +56,15 @@ setup_premake_path_override()
 def main():
   # First try the freshly-built premake.
   premake5_bin = os.path.join(premake_path, 'bin', 'release', 'premake5')
-  if not has_bin(premake5_bin):
```
hd-genius, line 57: The has_bin function searches for the binary globally in the PATH. The code here builds a path to the expected premake location `xenia/third_party/premake-core/bin/release/premake5` which is not in the user's PATH. This change just checks to see if the binary is available at that location instead.

hd-genius:
I just pushed an update that will also check for a globally installed version in the PATH.

PR #2325, opened 2025-11-20
Fix per-game config values leaking across titles

vivekstills:
# Per-Game Config Bug Fix

## The Problem

Settings from one game were incorrectly persisting when loading a different game. When Game A defined configuration values in its TOML file and Game B did not override those same settings, Game B would still use Game A's values. This occurred because the cvar system retained the previous game's game_config_value_ state without clearing it before loading the next title's configuration.

## The Solution

This patch adds a cleanup step that resets all per-game overrides before loading a new title's configuration. Each game now starts from global and default values unless it explicitly defines its own overrides.

## Changes Made

1. Added reset method to config variables
   - Implemented ResetGameConfigValue in the IConfigVar interface
   - Clears the cached game_config_value_ pointer
   - Triggers UpdateValue() to recalculate the setting using the standard precedence: default → global → CLI → game-specific

2. Added system-wide reset function
   - Created ResetGameConfigValues() in src/xenia/config.cc
   - Iterates through all registered config variables and resets each one

3. Modified game config loading
   - Updated LoadGameConfig() to call ResetGameConfigValues() at the start
   - Ensures each title loads from a clean state before applying its TOML overrides

## Impact

- Eliminates unintended inheritance of config values between games
- Ensures per-title configurations operate independently
- Resolves issues where settings from a previous session would affect subsequent titles

### DIFF
--- a/src/xenia/base/cvar.h
+++ b/src/xenia/base/cvar.h
@@ -50,6 +50,7 @@ class IConfigVar : virtual public ICommandVar {
   virtual std::string config_value() const = 0;
   virtual void LoadConfigValue(std::shared_ptr<cpptoml::base> result) = 0;
   virtual void LoadGameConfigValue(std::shared_ptr<cpptoml::base> result) = 0;
+  virtual void ResetGameConfigValue() = 0;
   virtual void ResetConfigValueToDefault() = 0;
 };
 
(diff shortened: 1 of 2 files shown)


vivekstills:
@benvanik @hlide @gibbed @shuffle2

Could you please review this PR when you get time?

Issue #2297, opened 2025-02-02
Segmentation when running a game on Linux


ovanes-ohanesian:
### Validation

- [x] I've read the [FAQ](https://github.com/xenia-project/xenia/wiki/FAQ).
- [x] The Xenia build used is from the master branch (not MLBS/AlexVS/Canary/pull requests, etc.)
- [x] This issue isn't for tech support (help with Xenia).
- [x] If this issue occurs in a specific game, I've done analysis to locate the faulty subsystem of the emulator and a potential reason in it.
- [x] I've checked if this issue hasn't already been reported.
- [x] My device meets the minimum requirements: https://github.com/xenia-project/xenia/wiki/Quickstart#system-requirements
- [x] (If building) I have read the building doc: https://github.com/xenia-project/xenia/blob/master/docs/building.md

### Describe what's going wrong

Emulator crashes on game load.

I tried to locate the place in code, it appears to be https://github.com/xenia-project/xenia/blob/3d30b2eec3ab1f83140b09745bee881fb5d5dde2/src/xenia/cpu/backend/x64/x64_function.cc#L37

### Describe what should happen

Emulator shouldn't crash on game load

### If applicable, provide a callstack here, especially for crashes

`i> 000084DC ----------------- END OF ACHIEVEMENTS ----------------
i> 000084DC Added handle:F8000024 for N2xe6kernel7XObjectE
i> 000084DC XThreadF8000024 (5) Stack: 70110000-70130000
[New Thread 0x7fffa2000640 (LWP 34065)]
[New Thread 0x7fff97e00640 (LWP 34066)]
i> 000084DC KernelState: Launching module...
i> 000084DC Added handle:F8000028 for N2xe6kernel7XObjectE
K> F8000024 XThread::Execute thid 5 (handle=F8000024, 'Kernel Dispatch (F8000024)', native=A2000640, <host>)
i> 000084DC XThreadF8000028 (6) Stack: 70150000-70160000
K> F8000028 XThread::Execute thid 6 (handle=F8000028, 'Main XThread (F8000028)', native=97E00640)

Thread 38 "Main XThread (F" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff97e00640 (LWP 34066)]
0x000055555811a540 in ?? ()
(gdb) bt
#0  0x000055555811a540 in  ()
#1  0x00000000a000003f in  ()
#2  0x00007fff97def650 in  ()
[...]

PR #2326, opened 2025-11-20
[Linux] [CPU] Fix crash on game load caused by corrupted global_mutex pointer

TheusHen:
## Description

Fixes emulator crashes during game load caused by memory corruption of the `PPCContext::global_mutex` pointer.

## Problem

The emulator would crash with `SIGSEGV` in `std::recursive_mutex::unlock()` with an invalid `this` pointer (e.g., `0x1`). Analysis of the callstack showed:
- Crash occurred during `GuestFunction::Call` → `std::recursive_mutex::unlock()`
- The `PPCContext::global_mutex` pointer was corrupted
- Likely caused by guest code overwriting the context structure beyond the allocated stack red zone

## Solution

Added defensive validation and restoration of the `global_mutex` pointer at three critical points:

1. **`Processor::Execute`**: Validates before executing any function
2. **`GuestFunction::Call`**: Validates before executing guest code
3. **`BuiltinFunction::Call`**: Validates builtin arg0 pointer (global mutex reference)

When corruption is detected:
- The pointer is restored to the valid global critical region mutex
- An error is logged with the corrupted pointer value
- Execution continues safely, preventing the crash

## Testing

- Tested with games that previously crashed on load
- Corruption detection logging helps identify problematic guest functions for further investigation

>  #2297

### DIFF
--- a/src/xenia/cpu/function.cc
+++ b/src/xenia/cpu/function.cc
@@ -44,6 +44,12 @@ bool BuiltinFunction::Call(ThreadState* thread_state, uint32_t return_address) {
   }
 
   assert_not_null(handler_);
+  // Detect corruption of builtin argument pointers (e.g., global mutex
+  // accidentally overwritten by guest code). A very low non-null address is
+  // almost certainly invalid here and has led to crashes in unlock().
+  if (arg0_ && reinterpret_cast<uintptr_t>(arg0_) < 0x1000) {
+    XELOGE("BuiltinFunction '{}' arg0 pointer appears corrupt: {:p}", name(), arg0_);
+  }
   handler_(thread_state->context(), arg0_, arg1_);
 
   if (original_thread_state != thread_state) {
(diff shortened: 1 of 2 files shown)


TheusHen:
I apologize for the approach taken in my previous commit. You're absolutely right, I was attempting to fix symptoms after corruption had already occurred rather than preventing the corruption in the first place.

The defensive checks I added merely detect that global_mutex has been corrupted and attempt to restore it, which doesn't solve the underlying problem and could mask more serious issues. The root cause is that guest code is somehow writing to the structure, corrupting the global_mutex pointer field.

I'll revert these changes and investigate the actual source of the memory corruption - likely related to stack overflow, red zone violations, or incorrect memory protection around the context structure. Thank you for the guidance.

Triang3l:
What places in which games is that crash reproducible in, by the way? It may be useful to be able to look at what actually happens there in the future.

PR #2329, opened 2025-12-12
[3PP] Uplifted gslang 16, SPIR-V Tools, Vulkan-Headers and VMA

Gliniak:


xenia-build
```diff
@@ -1084,22 +1080,14 @@ class BuildShadersCommand(Command):
                 if subprocess.call([
                        spirv_opt,
                        '-O',
+                       '--canonicalize-ids',
+                       '--strip-debug',
```
Triang3l: Will binding names and uniform buffer structure fields be preserved with `--strip-debug`? They're pretty useful for development, and have no effect on runtime performance and probably negligibly affect compilation time.

Triang3l (changes requested):

Gliniak:
Uplifted.
- Removed shader changes as new version of libs doesn't impact it.
- Issue with VULKAN_HPP_NO_TO_STRING still remains

PR #2334, opened 2026-01-06
./xb setup improvements: Successful Targeting of Mac OS

ryanwisemanmusic:
To properly port Xenia to Mac OS X, this PR contains 5 commits that enable ./xb setup to work on Mac OS X. This does not yet enable ./xb build: that will be my next focus, along with efficiently addressing any bugs as they arise.

I’m building Xenia via VSCode on macOS 10.13. If the included Info.plist file is too OS-specific, maintainers can cherry-pick the other working commits. I’ll keep the Info.plist private going forward if this happens

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -43,6 +43,12 @@ flags({
   "FatalWarnings",        -- Treat warnings as errors.
 })
 
+filter("platforms:Mac")
+  removeflags({
+    "FatalWarnings"
+  })
+filter({})
+
 filter("kind:StaticLib")
   defines({
     "_LIB",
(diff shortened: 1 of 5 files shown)


premake5.lua
```diff
@@ -43,6 +43,12 @@ flags({
   "FatalWarnings",        -- Treat warnings as errors.
 })
 
+filter("platforms:Mac")
+  removeflags({
+    "FatalWarnings"
```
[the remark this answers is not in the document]
ryanwisemanmusic: Yes, this was added specifically due to -Wdeprecated-declarations causing ./xb setup to fail on macOS.

I have begun work to fix ./xb build, so the code you see is only temporary. Once I'm able to get successful compilation with that command, I'll see what else needs to be changed so I can safely remove it.

premake5.lua
```diff
@@ -98,7 +104,20 @@ filter("platforms:Linux")
...
+    "-Wdeprecated-declarations",
+  })
+  links({
+    "stdc++",
+    "dl",
+    "pthread",
+  })
+  
+  pkg_config.all("gtk+-3.0")
   links({
```
Triang3l: These `links` now won't apply to the `Linux` platform, because the `filter` is set to `platforms:Mac` at this location rather than `platforms:Linux`.

Triang3l (changes requested):

tools/build/scripts/pkg_config.lua
```diff
@@ -21,7 +21,7 @@ function pkg_config.cflags(lib)
 end
 
 function pkg_config.lflags(lib)
-  if not os.istarget("linux") then
+    if not (os.istarget("linux") or os.istarget("macosx")) then
```
Margen67, line 24:
```suggestion
  if not (os.istarget("linux") or os.istarget("macosx")) then
```

PR #2335, opened 2026-01-06
[Base] Move Xenia to use C++20 standard

Gliniak:


Gliniak:
It won't compile at all and u8 is not needed or in this case we can explicitly use char8_t type, but I'm not sure if that's a good idea.

According to: https://stackoverflow.com/a/65468019
> On most platforms normal char strings are already UTF-8 and on Windows with MSVC you can compile with /utf-8 which will give you portable Unicode support on major operating systems.

Also these tests seems to fail to print correctly even in current version in exactly the same way so I see no regression

Gliniak:
So we need to leave u8 as it was? Previously there was some implicit conversion to char(?), but now but all utf8 related methods must be redone to use char8_t?

PR #2346, opened 2026-03-26
Add CCI (compressed xbox disc) support

EqUiNoX-Labs:
Supports compressed Xbox 360 images created by

https://github.com/Team-Resurgent/Simple360CCI

This format is also used in the upcoming Aurora release + NetISO loading


has207:
You should make this PR against xenia-canary, also why is lz4 being vendored this way here?

EqUiNoX-Labs:
> You should make this PR against xenia-canary, also why is lz4 being vendored this way here?

Ok will also do one against canary.

As to LZ4 was not sure best way to do it, and thought do like a library incase needed again for other future formats

PR #2372, opened 2026-09-13
corrections to vmx128 docs and disassembly typos

iximeow:
i've been working through a disassembler for the various Power implementations and `vmx128.txt` was a great help in understanding the layout of some of the Xenon-specific bits. i found a few bugs along the way and wanted to see them fixed here too. since i was already looking at vmx128 from Xenia i'd used the disassembly from `ppc_opcode_disasm_gen.cc` to compare against for Xenon-specific details and also found a few quirks there too that look to me like bugs here as well.

the two changes are distinct commits in case you'd want only one or the other, but for all of them i did check that the corresponding bits in decoding (for vmx128) or emitting (for the disasm changes) match the actual emulation behavior. i'm not sure where or if anyone ever looks at the ppc disassembly from Xenia so maybe it's not important to change! but it was super useful to me :)


iximeow:
oh, https://github.com/xenia-project/xenia/pull/2331 is much better on the vmx128 docs front. it includes everything i'd noticed.

